PR #5, opened 2019-04-04
Support for multiple architectures, new build system

danielinux:
* Makefile refactored to support RISC-V architecture
* Risc-V bootvector and startup files
* Per-target test application
* hifive1 HAL stub + test application


danielinux:
+ Automated tests passing (make test-all in CI)

test-app/hifive1.c
```diff
@@ -0,0 +1,163 @@
+/* 
+ * # sifive-welcome
+ * (A simple welcome example which prints SiFive banner and uses board LEDs
+ *
+ * Copyright 2018 SiFive, Inc
```
dgarske: I don't think we can/should have copyrights that are not ours in our wolfBoot library. Is there a way to write this hifive example so its original wold code?

danielinux: I've removed the test application and replaced with a while(1) loop for the moment, instead. It will be more appropriate indeed to have a full example in [wolfboot-examples](https://github.com/wolfssl/wolfboot-examples) later on.
Thanks!

danielinux:
Thank you!

PR #6, opened 2019-04-17
Python keytools

danielinux:
Host key tools rewritten in python to improve portability.
Build system updated accordingly.

All tests are sucessful.


Makefile
```diff
@@ -150,23 +150,17 @@ test-app/image.bin:
 
 include tools/test.mk
 
-tools/ed25519/ed25519_sign:
-	@make -C tools/ed25519
+ed25519.der:
+	@tools/keytools/keygen.py $(KEYGEN_OPTIONS) src/ed25519_pub_key.c
```
dgarske: On Mac OSX I had to change this to `@python ./tools/keytools/keygen.py ...`.

Also since I am on Python 2.17 I got the following error:

```
Creating file src/ed25519_pub_key.c
Traceback (most recent call last):
  File "./tools/keytools/keygen.py", line 86, in <module>
    f.write("0x%02X, " % c)
TypeError: %X format: a number is required, not str
make: *** [ed25519.der] Error 1
```

Makefile
```diff
@@ -184,8 +178,6 @@ clean:
 	@make -C test-app clean
 
 distclean: clean
-	@make -C tools/ed25519 clean
```
dgarske: On Max OSX had to change `xargs -x rm -f` to `xargs rm -f`

Makefile
danielinux: This won't work with python2 unless major changes are done to the script. At this point, we should just abandon legacy python an making python3 a requirement. I don't think it makes sense to support python 2 as it will go end-of-life in a few months

danielinux: Will change this to `@python3 ./tools/keytools/keygen.py` if it makes sense

Makefile
danielinux: ack

Makefile
dgarske: If that is compatible then great. Otherwise, just document the Python3 requirement in the README.

danielinux: This is done (see  94af1b7). Could you please check if everything compiles on MacOS?

PR #7, opened 2019-04-19
Ramcode

danielinux:
### Introducing RAMCODE tag to transfer functions to RAM

- Moved functions in the flash write path to RAM, so their execution
does not depend on flash access

- RAMCODE can be enabled via "make RAM_CODE=1"


danielinux:
wolfBoot can now update itself.
- Added CI tests for self-update (ed25519 + ecc256)

danielinux:
`make test-all` passing all test cases on the testbench :+1:

PR #11, opened 2019-05-22
Added support for STM32F7. Added HW-assisted dual-bank swap.

danielinux:
Added support for STM32F7 + test application.

Compile with `DUALBANK_SWAP=1` to activate HW-assisted Dual-bank swap. 

Requires an extra hal function. Implemented on STM32F76x-77x using the dual bank support as documented in [ST AN4826](https://www.st.com/content/ccc/resource/technical/document/application_note/group0/d2/bd/77/e8/0d/3a/43/cf/DM00266999/files/DM00266999.pdf/jcr:content/translations/en.DM00266999.pdf)


danielinux:
- No regressions in CI tests
- Preliminary tests with dual-bank on F769 successful

hal/stm32f7.c
```diff
@@ -0,0 +1,510 @@
...
+ *  - Dual bank with swapping, 2 x 512KB banks, 8 sectors each
+ *  - Single bank, 1 x 1MB, 8 sectors in total
+ *
+ *  The chosen configuration depends on the FLASH_OPTCR_DBANK value.
+ *
+ *
+ * */
+
+#define FLASH_SECTOR_UNUSED    0xFFFFFFFF
+/*  This memory mapping is for 2MB flash (STM32F769). */
```
miyazakh: When we use a device which has only 1MB flash, how should we configure? 
Can we use the same code?

test-app/stm32f7.c
```diff
@@ -0,0 +1,327 @@
...
+    uint32_t r_total = 0;
+    uint32_t tot_len = 0;
+    uint32_t next_seq = 0;
+    uint32_t version = 0;
+    uint8_t *v_array = (uint8_t *)&version;
+    int i;
+    memset(page, 0xFF, PAGESIZE);
+    hal_flash_unlock();
+//    version = wolfBoot_current_firmware_version();
+//    if ((version & 0x01) == 0)
```
miyazakh: Better to delete unused code?

PR #17, opened 2019-07-16
Added support for STM32-G0

danielinux:
Support for STM32-G0 @64MHz

- hal fully tested (including update)
- flash requires NVM_FLASH_WRITEONCE

to review/merge after STM32L0 PR


danielinux:
Updated wolfSSL to latest master, issue is fixed. `__gnu_thumb1_case_uqi` is only undefined if the switch is used on a local variable. `wc_HashInit` in hash.c has been fixed to use fn argument instead.

Cortex-M0 targets now build correctly with `-Os`

PR #43, opened 2020-04-03
External, emulated non-volatile memory over UART

danielinux:
Using the ext_flash interface, this uart driver allows to access an emulated partition hosted on a different machine than the target.

A binary file on the host machine is used to emulate the `UPDATE` and the `SWAP` partition on the target.

Updates can be made available on the host machine via the [uart-flash-server tool](../tree/uart-flash/tools/uart-flash-server).


src/loader.c
```diff
@@ -172,6 +173,10 @@ int main(void)
 {
     hal_init();
     spi_flash_probe();
+#ifdef UART_FLASH
+    uart_init((4 * 115200), 8, 'N', 1);
```
dgarske, line 177: What is `4 *` here? This loader.c seems like it should be generic and just pass 115200 and the porting implementation should handle the `4 *`?

config/examples/stm32f4-small-blocks-uart-update.config
```diff
@@ -0,0 +1,35 @@
+ARCH?=ARM
```
dgarske, line 1:
```
$ cp config/examples/stm32f4-small-blocks-uart-update.config .config
$ make
	[CC-ARM] hal/stm32f4.o
	[CC-ARM] src/loader.o
src/loader.c:26:24: fatal error: uart_flash.h: No such file or directory
```

danielinux:
I managed to reproduce that. It looks like you may have an empty `.bootloader-partition-size`, which is an intermediate, temporary file used to set the partition size in the linker script based on the configuration.

I committed a change to remove the file in `make clean`, which should get rid of the error

include/uart_flash.h
```diff
@@ -0,0 +1,44 @@
...
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1335, USA
+ */
+
+#ifndef UART_FLASH_DRI_H
+#define UART_FLASH_DRI_H
+#include <stdint.h>
+
+#ifdef UART_FLASH
+    #define UART_FLASH_BITRATE 460800
```
dgarske, line 36: Perhaps this should be wrapped in `#ifndef UART_FLASH_BITRATE`, so it can be overridden?

tools/uart-flash-server/ufserver.c
```diff
@@ -0,0 +1,418 @@
...
+		if (serinfo.custom_divisor * rate != serinfo.baud_base) {
+			warnx("actual baudrate is %d / %d = %f",
+			      serinfo.baud_base, serinfo.custom_divisor,
+			      (float)serinfo.baud_base / serinfo.custom_divisor);
+		}
+	}
+
+	fcntl(fd, F_SETFL, 0);
+	tcgetattr(fd, &options);
+	cfsetispeed(&options, speed ?: B38400);
```
dgarske, line 155: Does it make sense to use B460800, since that's what was used in the wolfBoot UART code?

tools/uart-flash-server/ufserver.c
```diff
@@ -0,0 +1,418 @@
...
+    return -1;
+}
+void hal_flash_unlock(void)
+{
+}
+void hal_flash_lock(void)
+{
+}
+
+#include <termio.h>
```
dgarske, line 100: I sent a patch via email to fix building this on the Mac. 
Don't include termio.h and linux/serial.h and change the B() macro to just return the value. And exclude the "custom divisor" chunk for Mac, since it doesn't have `struct serial_struct`.

include/uart_flash.h
danielinux: good idea

tools/uart-flash-server/ufserver.c
danielinux: patch applied thanks

danielinux: I re-added the includes under `#ifndef __MACH___` so that it builds on GNU again

tools/uart-flash-server/ufserver.c
danielinux: OK, default changed

danielinux:
PR updated after reviews

PR #48, opened 2020-05-22
Fixes/improvements: psoc6 test application

danielinux:
- fixed linker script
- linking with cy_gpio/cy_device for testing boot leds/pins


test-app/Makefile
```diff
@@ -110,6 +106,22 @@ ifeq ($(TARGET),hifive1)
   APP_OBJS+=hifive1_write_page.o
 endif
 
+# $(CYPRESS_PDL)/devices/templates/COMPONENT_MTB/COMPONENT_CM0P/system_psoc6_cm0plus.o
+
+ifeq ($(TARGET),psoc6)
+  LSCRIPT_TEMPLATE:=ARM-psoc6.ld
```
dgarske, line 112: Check indentation here. Some Makefiles require tabs.

danielinux: Fixed, thanks David! In this case it's within ifeq-s and not in a recipe, tabs cannot be used and indentation is just for readability and not enforced by makefile's syntax.

danielinux:
> Getting error:
> 
> Using ./config/examples/cypsoc6.config and `make clean && make`:
> 
> ```
> app_psoc6.c:29:10: fatal error: cy_gpio.h: No such file or directory
>    29 | #include "cy_gpio.h"
>       |          ^~~~~~~~~~~
> ```

do you have psoc6pdl in lib/psoc6? The file should be in the include path due to `CFLAGS+=-I$(CYPRESS_PDL)/drivers/include/` in `test-app/Makefile`

dgarske:
Yes:

```
find . -name cy_gpio.h
./lib/psoc6pdl/drivers/include/cy_gpio.h
```

Guessing its because the example .config is using a path that assumes its in test-app

PR #52, opened 2020-05-25
Added support for memory protection on ARM using MPU

danielinux:
This patch enables MPU by default on 32-bit ARM platforms, if supported.
- When in use, MPU limit the execution to the wolfBoot partition.
- When `RAM_CODE` is enabled, execution is allowed from static memory as well
- MPU is turned on at bring-up (before calling main) and it's disabled before boot


Makefile
```diff
@@ -157,6 +157,10 @@ ifeq ($(V),0)
   Q=@
 endif
 
+ifeq ($(NO_MPU),1)
+  CFLAGS+=-DWOLFBOOT_NO_MPU
```
dgarske, line 161: What about people who are building wolfBoot not using our Makefile? If they are on ARM it will default to having MPU code enabled. Perhaps it should be reverse logic, so it's disabled by default?

src/boot_arm.c
```diff
@@ -35,6 +36,130 @@ extern uint32_t *END_STACK;
 
 extern void main(void);
 
+#ifndef WOLFBOOT_NO_MPU
+#define MPU_BASE (0xE000ED90)
+#define MPU_TYPE			*((volatile uint32_t *)(MPU_BASE + 0x00))
+#define MPU_CTRL			*((volatile uint32_t *)(MPU_BASE + 0x04))
+#define MPU_RNR				*((volatile uint32_t *)(MPU_BASE + 0x08))
```
dgarske, line 43: Indent off here.

Makefile
danielinux: This is done on purpose so MPU is on by default, unless explicitly disabled (when compiling without the makefile, same applies as other options with the same logic such as NO_VTOR).

It is safe to keep MPU on on platforms that don't support it, because the code will do nothing if MPU->TYPE is '0'. (MPU->TYPE must be implemented as '0' by ARM Cortex-M cores when the feature is not available).

PR #68, opened 2020-07-06
Fix last-resort update

danielinux:
When the update image is valid, yet not marked for update, and the boot image is invalid/absent/corrupted, wolfBoot would try to run a forced last-resort update.

This mechanism has a logic flaw, which is fixed by this PR.

Existing tests on Jenkins are successful :heavy_check_mark:


danielinux:
see ZD10572

PR #75, opened 2020-08-11
Relocate encryption key slot to the end of the flash partition

danielinux:
- Encryption key slot moved to the end of the partition
- When encryption is enabled, partition flag and sector flags bitmap are moved back by `ENCRYPT_KEY_SIZE + ENCRYPT_NONCE_SIZE`
- Updated `uart_flash_server` to detect and support encrypted partitions
ZD 10733

### DIFF
--- a/src/libwolfboot.c
+++ b/src/libwolfboot.c
@@ -36,8 +36,10 @@
         #define XMEMCPY memcpy
         #define XMEMCMP memcmp
     #endif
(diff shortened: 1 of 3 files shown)


danielinux:
@dgarske I tested both encrypted and non-encrypted case on STM32WB55.

- For the encrypted case, you may need to reboot twice, because the test application will set the temporary key.

- It could be that the update is not starting because the image is not marked. Please use `tools/scripts/prepare_encrypted_update.sh` to create `update.bin` and then launch `ufserver` with `update.bin` which is padded and contains the right flag to trigger the installation.

Thanks,

-- 
Daniele

PR #77, opened 2020-09-03
Xilinx Zynq Improvements

dgarske:
* Increase stack/heap for Zynq.
* Pass return code from `wolfBoot_verify_signature` when calling `wolfBoot_verify_authenticity`.
* Add additional Zynq debugging logs for update_ram.c.
* Zynqmp updates to use xil_printf.
* Adds UART support for debugging.
* eFuse reading.
* Fix the exclude items in wolfBoot Xilinx SDK project.
* Adjust the QSPI clock.


src/update_ram.c
```diff
@@ -64,9 +75,13 @@ void RAMFUNCTION wolfBoot_start(void)
     }
 
     for (;;) {
-        if ((wolfBoot_open_image(&os_image, active) < 0) ||
-            (wolfBoot_verify_integrity(&os_image) < 0) ||
-            (wolfBoot_verify_authenticity(&os_image) < 0)) {
+        if (((ret[0] = wolfBoot_open_image(&os_image, active)) < 0) ||
```
danielinux, line 78: Not sure this is needed. The status of each of the three operation is kept in `os_image.hdr_ok`, `os_image.sha_ok` and `os_image.signature_ok`, respectively, which can be used in the xil_printf below. Assignments in conditions are more prone to errors.

src/spi_flash.c
```diff
@@ -200,7 +202,7 @@ uint16_t spi_flash_probe(void)
 
 void spi_flash_sector_erase(uint32_t address)
 {
-    uint8_t status;
+    //uint8_t status;
```
danielinux, line 205: remove if unused

src/spi_flash.c
```diff
@@ -94,7 +96,7 @@ static void flash_write_enable(void)
 
 static void flash_write_disable(void)
 {
-    uint8_t status;
+    //uint8_t status;
```
danielinux, line 99: remove if unused

src/update_ram.c
dgarske: It is the return code from `wolfBoot_verify_signature` being captured here. I've cleaned this up.

danielinux (approved):

PR #80, opened 2020-09-21
increase spi speed

LOorts-Aloxy:
Hi,

By removing this flag, we're setting the baudrate prescaler from fPCLK/64 to fPCLK/2. This results in faster transfer speeds to external flash. We've already tested this change on all of our instances for a longer time.

### DIFF
--- a/hal/spi/spi_drv_stm32.c
+++ b/hal/spi/spi_drv_stm32.c
@@ -155,7 +155,11 @@ void spi_init(int polarity, int phase)
         spi_tpm2_pin_setup();
         APB2_CLOCK_ER |= SPI1_APB2_CLOCK_ER_VAL;
         spi1_reset();
+#ifdef PLATFORM_stm32l0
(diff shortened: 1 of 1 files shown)


danielinux (changes requested):
This change breaks the SPI communication on other platforms. On STM32F4 (used for non-regression tests), this will prevent any SPI communication. 

Would it be possible to make this setting target-specific?

L-I-Am:
you mean to surround this line with #ifdef PLATFORM_stm32l0? or how would do you envision it to be target-specific? perhaps set a define of the contents of the SPI1_CR1 register depending on the platform?

LOorts-Aloxy:
and how do you envision this target-specific? surround that line with #ifdef PLATFORM_stm32l0? or create a define with the contents of SPI1_CR1 depending on the platform? or... ?

danielinux:
`#ifdef PLATFORM_stm32l0` would be OK! Thanks!

danielinux:
Tests are now passing. Merged. Thanks @LOorts-Aloxy !

PR #82, opened 2020-10-19
Set tag type and size to 2 bytes in doc

DurandA:
In the firmware image documentation, expand the type and size of the header
tags from 1 to 2 bytes to match the implementation.


DurandA:
Is the header `HDR_IMG_TYPE` mandatory? I could add it to the documentation.

DurandA:
> A '0' in the Type field indicate the end of the Tag. The rest of the header carries no more Tags. The 'end of tags' type has no **size** field.

Both the python (_sign.py_) and C (_sign.c_) signing tools pad the header with 0xFF until the end. Is the end tag deprecated?

PR #85, opened 2020-11-04
Sector flags relocate

danielinux:
Relocate sector flags to a well-known address in boot partition, when the option `FLAGS_HOME` is used.

See ZD11054 for reference.


dgarske (approved):
Thanks Daniele. Code changes look good and build correctly. I don't have L0 hardware to test on.

PR #90, opened 2020-11-23
Encryption bug - integrity check of boot partition fails after updating

LOorts-Aloxy:
This bug got introduced in https://github.com/wolfSSL/wolfBoot/commit/72eb2c205ba49df5ad50365712d9b2e67452bd38.


danielinux:
Thanks for spotting this.
Since each sector uses 4 bits (one nibble) to keep its state, I believe that the formula to calculate the overhead should be:
`(4 + 1 + (WOLFBOOT_PARTITION_SIZE  / (2 * WOLFBOOT_SECTOR_SIZE))) /* MAGIC + PART_FLAG (1B) + (N_SECTORS / 2)`

LOorts-Aloxy:
Tested on our board with the proposed OVERHEAD and seems to work!

danielinux:
Thanks! Could you please fix the comment as well `(N_SECTORS / 2)`? I will merge right away and re-run all the tests.

LOorts-Aloxy:
should be fixed btw

danielinux:
Merged, thanks!

PR #95, opened 2020-12-09
Add measured boot

tomoveu:
This PR adds measured boot by extending the boot image digest calculated by wolfBoot into a TPM PCR. The PCR index is selected by the user.

@danielinux please review and let me know if the way I added the user option is acceptable

@dgarske We should take into account that Microsoft Windows (with the most heavy PCR usage out there) uses PCR0-11 and the only remaining **safe PCRs** for use by us, as a bootloader, are PCR12-15.


src/image.c
```diff
@@ -657,6 +675,10 @@ int wolfBoot_verify_integrity(struct wolfBoot_image *img)
         return -1;
     if (image_hash(img, digest) != 0)
         return -1;
+#if defined(WOLFBOOT_TPM) && defined(WOLFBOOT_MEASURED_BOOT)
```
dgarske, line 678: Consider putting this after the memcmp, so its only updated on success.

tomoveu: I had this comment in my initial code, but removed it in the last moment. Perhaps I should re-add it?
+    /*
+     * TPM Measurement must be performed regardless of the
+     * verification outcome afterwards, becuse the purpose
+     * of a Measured Boot is to record the current boot state
+     */

options.mk
```diff
@@ -187,6 +187,11 @@ ifeq ($(WOLFTPM),1)
   endif
 endif
 
+ifeq ($(MEASURED_BOOT),1)
```
dgarske, line 190: Please add this new option `MEASURED_BOOT` to the `tools/config.mk`. Add documentation for the new option. Note it requires `.config` with `WOLFTPM=1` and `MEASURED_BOOT=[PCRNUM]`.

tomoveu: I was wondering if I could code this requirement the way we do in wolfTPM, but it seems with option.mk such dependencies are not possible. I will make sure to add detailed info in documentation. Thanks

dgarske: Yes, code it to force WOLFTPM=1 in the Makefile

src/image.c
dgarske: Yes add this comment.

tomoveu:
Thanks for the review @dgarske

Both suggestions are added and I want to finalize the testing before making the new push.

options.mk
danielinux: or even better, in `options.mk`

tomoveu:
Feedback taken in. Some testing done too. I will add updated _README.md_ and new file in _docs/measured_boot.md_ to finalize this PR. 

And I want to test bit more the code, because I had some issues with my HW setup .Purely HW stuff, SPI wiring issue between my STM32F4 and LetsTrust TPM.

Also, @danielinux  made a great suggestion for an extra effort - To add an example of measured_boot app in the _wolfboot-examples_ repo. This would be a separate effort and if time allows it. So, right now I have to push the docs and do a bit more testing.

tomoveu:
New feature tested with Infineon TPM9670_Raspberry module and seems to be working as expected.

I verified that PRC extend operations, using the same firmware image, are reproducible in wolfBoot :)

@danielinux please review the code, I will wrap this PR with updating the documentation as discussed.

tomoveu:
Documentation about measured boot added. README.md is also updated.

I would like at least one more person to test on the real hardware before I merge this PR.

I am making a new wolfBoot-example for measured boot, but that would be a PR for the other repo.

dgarske (approved):

dgarske:
@tomoveu : I will hold off to merge until I can setup on hardware here with your wolfboot-examples repo (once pushed).

tomoveu:
@dgarske Just to make sure there is no misunderstanding: This is the measured boot implementation and it is ready for merging.

It can be tested by using gdb, and checking the PCR value is reproducible when using the same test-app (default one works).

The new wolfboot-example is only to make it easier for people to use the measured boot. It is not required for having measured boot.

PR #96, opened 2020-12-15
Remove unnecessary KEYGEN_TOOL from test-enc.mk

nkapu:
KEYGEN_TOOL is not used in this file and doesn't have proper detection logic for right variant of the tool. Main Makefile includes this file after test.mk and stray KEYGEN_TOOL define here overrides tool variant already detected by test.mk with python version.


danielinux:
Outdated by ecfc4b5008353634277000b46709deb919c10d43.

PR #115, opened 2021-02-05
Fix IAR warnings

kaizoku-oh:
- Edited .gitignore to ignore IAR generated files
- Removed unused variables resulting in warnings
- Casted variables to avoid integer conversion warnings
- Remove unreached return statement resulting in warning


danielinux:
These have been addressed in #178 . Closing.

PR #117, opened 2021-03-18
Changed RSA4096 to use SP math. Added G0 details. Update submodules.

dgarske:


dgarske:
Replaced with PR #118.

PR #119, opened 2021-04-22
make: PPC/Big Endian support

elms:
Add `bin-assemble` tool instead of `dd` for more flexible assembly of factory image.
Add Big Endian support for header and trailer

makefile clean up
 * remove unused targets
 * distclean - cleans all tools
 * remove use of some temp files

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,19 +47,15 @@ all: $(MAIN_TARGET)
 
 wolfboot.bin: wolfboot.elf
 	@echo "\t[BIN] $@"
-	$(Q)$(OBJCOPY) -O binary $^ $@
+	$(Q)$(OBJCOPY) -O binary --gap-fill=255 $^ $@
 
 align: wolfboot-align.bin
 
-.bootloader-partition-size:
-	$(Q)printf "%d" $(WOLFBOOT_PARTITION_BOOT_ADDRESS) > .wolfboot-offset
-	$(Q)printf "%d" $(ARCH_FLASH_OFFSET) > .wolfboot-arch-offset
-	$(Q)expr `cat .wolfboot-offset` - `cat .wolfboot-arch-offset` > .bootloader-partition-size
-	$(Q)rm -f .wolfboot-offset .wolfboot-arch-offset
+BOOTLOADER_PARTITION_SIZE=$$(( $(WOLFBOOT_PARTITION_BOOT_ADDRESS) - $(ARCH_FLASH_OFFSET)))
 
-wolfboot-align.bin: .bootloader-partition-size wolfboot.bin
-	$(Q)dd if=/dev/zero bs=`cat .bootloader-partition-size` count=1 2>/dev/null | tr "\000" "\377" > $(@)
-	$(Q)dd if=wolfboot.bin of=$(@) conv=notrunc 2>/dev/null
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -100,15 +96,20 @@ test-app/image_v1_signed.bin: test-app/image.bin
 
 factory.bin: $(BOOT_IMG) wolfboot-align.bin $(PRIVATE_KEY) test-app/image_v1_signed.bin
 	@echo "\t[MERGE] $@"
-	$(Q)cat wolfboot-align.bin test-app/image_v1_signed.bin > $@
+	$(Q)$(OBJCOPY) --set-start=$(ARCH_OFFSET) \
+		--add-section .signed_image=test-app/image_v1_signed.bin \
+		--change-section-vma .signed_image=$(WOLFBOOT_PARTITION_BOOT_ADDRESS) \
+		--set-section-flags .signed_image=alloc,readonly \
+		wolfboot.elf factory.elf
```
danielinux, line 103: Why do you call the output binary file "wolfboot.elf"? Is there a reason why you are using objcopy to create the factory image? This breaks the cortex-M build

danielinux (changes requested):
Using objcopy to merge the two images (that should be contiguous in the same address space) breaks other builds. Is there a possibility to still use cat to merge the two files? Why do you need vma address? Could you use ARCH_FLASH_OFFSET instead?

tools/test-expect-version/test-expect-version.c
```diff
@@ -44,11 +44,14 @@
 #define MSGLEN      (4 + 4 + 8)
 
 #ifndef UART_DEV
-#define UART_DEV "/dev/ttyS0"
+#define UART_DEV "/dev/ttyACM1"
```
danielinux, line 47: This breaks the jenkins tests, please revert to "/dev/ttyS0"

tools/test-expect-version/test-expect-version.c
```diff
@@ -87,7 +90,7 @@ int main(int argc, char** argv)
     tty.c_cc[VTIME] = 5;
     tcsetattr(serialfd, TCSANOW, &tty);
 
-    alarm(60);
+    alarm(TIMEOUT);
```
danielinux, line 93: 30 seconds may be too short in some cases to wait until the update is completed. 60 seconds was a safer choice to avoid false negatives on jenkins tests

Makefile
elms: I abandoned this approach and have added bin-assemble to allow creating combined flash objects with some more flexibility. For T2080 the app image needs to be before the bootloader.

tools/test-expect-version/test-expect-version.c
elms: reverted

tools/test-expect-version/test-expect-version.c
elms: Set back to 60. accidental commit while testing.

elms:
@danielinux I think this is ready for review. The test node is not reachable and when I was trying to run tests last the TPM appeared to be a state that even master was not passing the TPM tests.

danielinux:
@elms Thanks! All tests are passing now. Merging!

PR #124, opened 2021-05-24
Add support for TMS570LC43xx

elms:
- [X] README
- [X] Link with TI compiler
- [X] Simple test image
- [X] Verify image
- [X] Authenticate image
- [X] Link F021 flash library and load in RAM
- [X] IDE project files
- [x] Jump to app


tools/keytools/keygen.c
```diff
@@ -34,6 +34,7 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <errno.h>
+#include <libgen.h>
```
dgarske, line 37: Note... we also have a python version of the keygen tool, which may also need to be updated.

elms:
Still working on the jump to app code. Things I don't love are some of the changes to support the different compiler and linker flags.

I want to test with stm32l5 as the `mpu_init` logic changed some and also against some other targets to make sure the building changes didn't miss any compiler/linker flags.

.gitignore
```diff
@@ -51,6 +51,7 @@
 *.sig
 
 # automatically generated source files
+**_pub_key.c
```
danielinux, line 54: The demo public key in the IAR example should still be part of the repo (it should be OK as long as the file does not get removed)

tools/keytools/keygen.c
```diff
@@ -294,12 +297,22 @@ int main(int argc, char** argv)
         }
     }
     output_pubkey_file = strdup(argv[argc - 1]);
+    dir = strdup(argv[argc - 1]);
```
danielinux, line 300: Looks like a command line argument has been added (`dir`) as positional, last argument? This should be at least documented in usage()

danielinux: Also ensure that keygen.c and keygen.py have the same interface and do the same thing (as @dgarske pointed out in his comment)

danielinux (changes requested):

tools/keytools/keygen.c
elms: I'll split that out as a separate PR. It's not necessary anymore for this PR.

elms:
I realized the changes to `keygen` aren't required for the PR. The key can now reside in the normal location and I have a link in the project resources.

I'd appreciate any feedback on the changes to the core tooling. Particularly I want to make sure changes don't impact stm32l5 targets around `mpu_init`

src/boot_arm.c
```diff
@@ -187,12 +196,12 @@ void isr_reset(void) {
         dst++;
     }
 
-#if !defined(PLATFORM_stm32l5) || (__ARM_FEATURE_CMSE == 0) || (__ARM_FEATURE_CMSE != 3U)
```
elms, line 190: Removed this guard as I believe it is handled by `WOLFBOOT_NO_MPU` making `mpu_init` a nop.

hal/ti_hercules.c
```diff
@@ -0,0 +1,159 @@
...
+#include "hal.h"
+
+#ifndef CORTEX_R5
+#   error "wolfBoot TI Hercules HAL: wrong architecture selected. Please compile with TARGET=ti_hercules."
+#endif
+
+#include <F021.h>
+#include <stdio.h>
+
+//#define FLASH_DEMO 1
```
dgarske, line 35: Prefer C style `/* */` code comments

hal/ti_hercules.c
```diff
@@ -0,0 +1,159 @@
...
+
+/* public HAL functions */
+void hal_init(void)
+{
+    int freq_MHz = 16;
+    int st = Fapi_initializeFlashBanks(freq_MHz);
+    if (st != 0) {
+        printf("Failed Fapi_initializeFlashBanks(%d) => (%d)\n", freq_MHz, st);
```
dgarske, line 43: Avoid printf in wolfBoot. Instead use the `wolfBoot_printf` macro/function enabled with `PRINTF_ENABLED`.

hal/ti_hercules.c
```diff
@@ -0,0 +1,159 @@
...
+        return;
+    }
+
+#if FLASH_DEMO
```
dgarske, line 47: This could not be defined. Prefer `#if defined(FLASH_DEMO) && FLASH_DEMO`

hal/ti_hercules.ld
```diff
@@ -0,0 +1,53 @@
+
```
dgarske, line 1: This file appears duplicated in `IDE/CCS/TMS570LC43xx/target.ld`? Would be neat if TI CCS could use this one, otherwise two is fine, but might be worth documenting the differences (one with our make cross and one with TI).

src/boot_arm.c
```diff
@@ -187,12 +196,23 @@ void isr_reset(void) {
...
 
-#if !defined(PLATFORM_stm32l5) || (__ARM_FEATURE_CMSE == 0) || (__ARM_FEATURE_CMSE != 3U)  
     mpu_init();
-#endif 
+
     /* Run the program! */
     main();
 }
+
+// forward to app handler
```
dgarske, line 205: Perhaps wrap this new ISR forwarder with `#ifdef CORTEX_R5`? Also note the `//` should use `/* */`

src/boot_arm.c
```diff
@@ -290,6 +310,22 @@ typedef void(*NMIHANDLER)(void);
 #   define isr_NMI isr_empty
 #endif
 
+#ifdef CORTEX_R5
+//__attribute__ ((section(".isr_vector")))
```
dgarske, line 314: Can we remove this code comment or point to the ISR forwarder macro/code above?

test-app/startup_arm.c
```diff
@@ -29,15 +29,18 @@ extern unsigned int _end_bss;
 extern unsigned int _end_stack;
 extern unsigned int _start_heap;
 
-static int zeroed_variable_in_bss;
-static int initialized_variable_in_data = 42;
-
+#ifdef STM32F
```
dgarske, line 32: I am not seeing `STM32F`? Perhaps you meant the new `STM32` one in test-app/MakeFile:53? Perhaps consider renaming this, since it is for all ARM archs?

test-app/startup_arm.c
```diff
@@ -283,7 +286,7 @@ void (* const IV[])(void) =
     isr_empty, //	UCPD1_IRQHandler
     isr_empty, //	ICACHE_IRQHandler
     isr_empty, //	OTFDEC1_IRQHandler
-#else /* For STM32F4 */
+#elif STM32 /* For STM32F4 */
```
dgarske, line 289: Also if someone is building outside our make system you should also check `defined(STM32)`.

dgarske (changes requested):

hal/ti_hercules.ld
elms: This version allows for different config settings with substitution of `WOLFBOOT_PATITION_BOOT_ADDRESS`. Not sure how to support the same with the IDE.

hal/ti_hercules.c
elms: Done

hal/ti_hercules.c
elms: Done

hal/ti_hercules.c
elms: Done

options.mk
```diff
@@ -15,12 +15,13 @@ ifeq ($(SIGN),ECC256)
     ./lib/wolfssl/wolfcrypt/src/ecc.o \
     ./lib/wolfssl/wolfcrypt/src/memory.o \
     ./lib/wolfssl/wolfcrypt/src/wc_port.o \
+    ./lib/wolfssl/wolfcrypt/src/wolfmath.o \
     ./lib/wolfssl/wolfcrypt/src/hash.o
-  CFLAGS+=-DWOLFBOOT_SIGN_ECC256 -DXMALLOC_USER
+  CFLAGS+=-D"WOLFBOOT_SIGN_ECC256"
```
dgarske, line 20: Did you mean to remove the XMALLOC_USER?

options.mk
```diff
@@ -15,12 +15,13 @@ ifeq ($(SIGN),ECC256)
...
   ifeq ($(WOLFTPM),0)
     CFLAGS+=-Wstack-usage=3888
   else
-    CFLAGS+=-Wstack-usage=6680
+    #CFLAGS+=-Wstack-usage=6680
```
dgarske, line 24: Maybe keep this on for DEBUG=1 build?

dgarske (changes requested):

options.mk
elms: No, I've generalized this so it can be overridden and done that for the R5.

options.mk
elms: Nope, that was mistake from initial work. The TI compiler doesn't recognize that flag, but I refactored and have it overridden for the R5.

dgarske (approved):
Thanks @elms . Great work on this. I am assigning over to @danielinux for a final pass.

src/boot_arm.c
danielinux: OK, this was here because our MPU code is not compatible with Cortex-M33 which has a different MPU controller that we don't support.

Makefile
```diff
@@ -156,11 +159,11 @@ check_config:
 
 %.o:%.c
 	@echo "\t[CC-$(ARCH)] $@"
-	$(Q)$(CC) $(CFLAGS) -c -o $@ $^
+	$(Q)$(CC) $(CFLAGS) -c --output_file $@ $^
```
danielinux, line 162: This breaks the compiler?
```
gcc: error: hal/stm32f4.o: No such file or directory
gcc: error: unrecognized command line option ‘--output_file’; did you mean ‘--output’?
```

arch.mk
```diff
@@ -233,15 +270,21 @@ endif
...
 CFLAGS+=-DARCH_FLASH_OFFSET=$(ARCH_FLASH_OFFSET)
 
-## Toolchain setup
-CC=$(CROSS_COMPILE)gcc
-LD=$(CROSS_COMPILE)gcc
-AS=$(CROSS_COMPILE)gcc
-OBJCOPY:=$(CROSS_COMPILE)objcopy
-SIZE:=$(CROSS_COMPILE)size
-BOOT_IMG?=test-app/image.bin
+# Setup default optimizations (for GCC)
```
danielinux, line 273: without defaulting on `$(CROSS_COMPILE)gcc`, I get `CC=cc` as default on my system, so it won't invoke the cross compiler properly.

```
make V=1
	[CC-ARM] hal/stm32f4.o
cc -D"__WOLFBOOT" -D"WOLFBOOT_VERSION=0UL" [...]
```

Please restore $(CROSS_COMPILE)gcc as default choice, or the CC variable will have "cc" in it

danielinux (changes requested):

danielinux (approved):

danielinux:
Thanks @elms - looks good now. No new regressions in the tests either. Merging.

PR #128, opened 2021-06-28
Configuration switch to disable firmware authentication

danielinux:
By compiling with `SIGN=NONE`, wolfBoot disables signature verification.

Changes:
- Added `HDR_IMG_TYPE_AUTH_NONE` in the manifest header image type
- `SIGN=NONE` option enables `WOLFBOOT_NO_SIGN`, which disables authentication
- sign.py new option `--no-sign` creates an image without the two fields (public key hash + signature)
- sign.c new option `--no-sign` like the above
- docs/compile.md section updated
- Added two non-regression tests with NO_SIGN (171, 173)


dgarske (approved):
Reviewed and tested. All looks great. Thanks @danielinux

PR #131, opened 2021-07-09
Add WOLFBOOT_SMALL_STACK

danielinux:
Added a `WOLFBOOT_SMALL_STACK` option, that keeps stack usage under control by using a pool of fixed objects.

This is needed on platforms with limited stack space assigned to the bootloader stage (i.MX-RT1060).


src/xmalloc.c
```diff
@@ -0,0 +1,194 @@
...
+#include "target.h"
+
+
+struct xmalloc_slot {
+    uint8_t *addr;
+    uint32_t size;
+    uint32_t in_use;
+};
+
+#define SP_DIGIT_SIZE (4)
```
dgarske, line 39: Is this true for 64 bit targets?

src/xmalloc.c
```diff
@@ -0,0 +1,194 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1335, USA
+ */
+
+#include <stdlib.h>
+#include <stdint.h>
+#include <stddef.h>
+#include <wolfssl/wolfcrypt/sp.h>
```
dgarske, line 29: Perhaps include settings.h above for good coding practice?

src/xmalloc.c
```diff
@@ -0,0 +1,194 @@
...
+
+static struct xmalloc_slot xmalloc_pool[] = {
+#ifdef WOLFBOOT_HASH_SHA256
+    { (uint8_t *)sha_block, HASH_BLOCK_SIZE * sizeof(uint32_t), 0 },
+#endif
+    { (uint8_t *)sha512_block, sizeof(word64) * 16, 0 },
+    { NULL, 0, 0}
+};
+
+#elif defined WOLFBOOT_SIGN_RSA2048
```
dgarske, line 113: Looks like the 2048 and 4096 sections are similar enough you could combine and compute based on key size.

dgarske (changes requested):
Great work on this. Also could the indent in the xmalloc.c be made more consistent?

src/xmalloc.c
danielinux: Good point. changed to `sizeof(sp_digit)`

dgarske (approved):
Great job on the improvements.

Note: ECC256,RSA2048 and RSA4096 with SPMATH=0 fail, but that also happens in master. Consider a PR for that prior to the release.

PR #146, opened 2021-09-06
Fixed keytools error handling. Removed inclusion of target.h.

danielinux:


danielinux:
tested via:
```
make distclean
make keytools
cp config/examples/stm32wb-delta.config .config 
make test-delta-update
```

dgarske (approved):

PR #147, opened 2021-09-16
WIP: TI Cortex R5 hercules updates

elms:


elms:
I need to revisit some of the changes and make sure they are compatible with other platforms and configurations.

danielinux (approved):
Tested OK

elms:
Need to test one more minor change.

elms:
> `TRAILER_SKIP` option was introduced for another TI platform (cc26x2) to skip the radio settings at the end of the flash partition when writing the flags.
> 
> Is `PART_BOOT_ENDFLAGS` exported for a similar goal? If this is the case you may want to consider exporting `TRAILER_SKIP`, which should also work if encryption is enabled.

The request was to allow storing the flags in an unused sector that is in Bank 1. This is so application code to update the flags can do so without needing to run from RAM. To erase a sector from a flash bank, code can not be executed from that bank for this platform. I see that the `TRAILER_SKIP` could be used, but it would be a indirect (and possibly error prone) in selecting the location for storing the flags at a particular location.

PR #150, opened 2021-10-06
Fixed delta updates from external devices

danielinux:
This fix allows incremental updates to work from an external NVM.

### DIFF
--- a/src/delta.c
+++ b/src/delta.c
@@ -21,8 +21,6 @@
 #include <stdint.h>
 #include <string.h>
 #include "delta.h"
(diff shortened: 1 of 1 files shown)


danielinux:
Tested with STM32WB.

Incremental update test on internal flash:
```
cp config/examples/stm32wb-delta.config .config
make test-delta-update
```

Incremental update test on external flash simulator (via uart flash server):
```
cp config/examples/stm32wb-delta-ext.config .config
make test-delta-update-ext
```

PR #153, opened 2021-11-10
Fixes and improvements to STM32G0

dgarske:
* Adds secure boot support
* Use flash base address (not RAM)


danielinux (approved):

hal/stm32g0.c
```diff
@@ -284,11 +291,48 @@ void hal_init(void)
     clock_pll_on(0);
 }
 
-void hal_prepare_boot(void)
+static void RAMFUNCTION hal_secure_boot(void)
```
danielinux, line 294: Just a preference here, feel free to ignore. I'd avoid the `hal_` prefix on static functions within the specific hal, because that may wrongly suggest that the function is part of the `hal.h` API

danielinux:
All CI tests passing, ready to merge.

danielinux:
Merged. Thanks @dgarske !

PR #161, opened 2021-12-14
Fixes for building the C tools with Visual Studio

dgarske:
Still fixing some issues with Windows stdlib includes, but putting this work up now.


dgarske:
Replaced by #162 . Thanks @elms

PR #163, opened 2021-12-15
Delta update fixes

danielinux:
- Fixed patch offset in delta update (ZD12933)
- uart-flash-server: build own libwolfboot.o (to avoid object mismatch when building after wolfBoot)
- Fixed test cases for delta updates


dgarske (approved):

elms (approved):
Thanks @danielinux for confirming that the test cover this. I think I understand the change now too.

PR #166, opened 2021-12-30
Fixed corner case delta bug when matching across old/new sector

danielinux:
ZD12933

### DIFF
--- a/src/delta.c
+++ b/src/delta.c
@@ -50,10 +50,6 @@ struct BLOCK_HDR_PACKED block_hdr {
 #define ext_flash_check_read ext_flash_read
 #endif
 
-#ifndef WOLFBOOT_SECTOR_SIZE
-#   define WOLFBOOT_SECTOR_SIZE 0x1000
-#endif
-
 int wb_patch_init(WB_PATCH_CTX *bm, uint8_t *src, uint32_t ssz, uint8_t *patch,
         uint32_t psz)
 {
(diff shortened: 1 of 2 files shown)


elms (commented):
I spent a while trying to understand the change, but it's not clear to me how to reproduce and test. Can you add details to the description of the issue and how this fixes it?

I also had a issue running `make -C tools/delta clean delta-test` and realize I had a target.h with sector size larger than `0x1000` which was resulting in a failure. Not sure if this is expected. If so, I suggest a build time error. Something like:
```
#if WOLFBOOT_SECTOR_SIZE > 0x1000
#   error "Sector size too large to support delta"
#endif
```

danielinux:
Thanks for the review! I've added some comments in the `wb_diff` code to explain the steps there. The issue was caused by wrong indexes being incremented when the 'matching area' was being extended across the sector limit. This produced a delta patch with a matching area that was bigger than the actual one by one byte, causing in turn a difference between image 'B' and the resulting image after diff/patch, by one byte when the match occurred at the end of the sector.

>I also had a issue running make -C tools/delta clean delta-test and realize I had a target.h with sector size larger than 0x1000
> which was resulting in a failure. Not sure if this is expected. If so, I suggest a build time error. Something like:


Thanks for spotting the failing test. This is not expected, and I'm investigating further.

danielinux:
The issue with the bmdiff was due to a hardcoded delta buffer size that was creating an invalid patch when WOLFBOOT_SECTOR_SIZE was != 4096. It should be fixed in 1886ad3 now.

elms (approved):

Issue #114, opened 2021-02-05
Cannot compile IAR example project


kaizoku-oh:
The IAR example project for stm32f4 found under IDE/IAR is not able to compile.
After generating the keypair using the generate_key.bat script I wasn't able to compile the project.
I'm using IAR 8.50.9 on Windows 10 x64.
I feel like I'm missing something?

![SharedScreenshot1](https://user-images.githubusercontent.com/22129291/107048083-6485df00-677d-11eb-906d-bc891c9bcdaa.jpg)
![SharedScreenshot2](https://user-images.githubusercontent.com/22129291/107048095-6780cf80-677d-11eb-8c21-54c07ffd9065.jpg)

danielinux:
Hello @kaizoku-oh,

what is the target platform? You may need to define your partitioning in `include/target.h`. When `make` is used, the file is created automatically. In your case, you may want to create it manually starting from `include/target.h.in`, and specifying the flash geometry.

The `include/target.h` automatically generated when using make, configured for the STM32F407 with the parameters in the documentation is the following:

```
#ifndef H_TARGETS_TARGET_
#define H_TARGETS_TARGET_

#define WOLFBOOT_SECTOR_SIZE                 0x20000
#define WOLFBOOT_PARTITION_BOOT_ADDRESS      0x20000
#define WOLFBOOT_PARTITION_SIZE              0x20000 
#define WOLFBOOT_PARTITION_UPDATE_ADDRESS    0x40000
#define WOLFBOOT_PARTITION_SWAP_ADDRESS      0x60000

#endif /* !H_TARGETS_TARGET_ */
```

kaizoku-oh:
Yes, that is what I was missing, I created an **include/target.h** file and it compiled without errors (although there were some warnings)
Thanks @danielinux

PR #173, opened 2022-01-20
Stm32f4: change FLASH origin

danielinux:


danielinux:
re-run (build #114) test-base target succeeding.

elms (commented):
Looks good. I was able to run in renode and see what I think is expected on the UART.

I tried to step through with gdb, but wasn't able to. Excited to see more functionality. Will you add to this PR? If not perhaps best to break out the renode commit and create a new PR with that and some README on how to run it.

danielinux:
Hey @elms thanks for the comments and the fix to the script. The renode script was only added to double check that wolfBoot can work on the emulator. Next step in this direction (in its own PR) will be to move most of the test cases to renode to be able to run without physical boards. TPM can be probably simulated via swtpm too.

I would keep the 'base' renode script for stm32f4 here for now and add a proper README in a PR later where we can migrate some tests too.

elms (approved):

PR #175, opened 2022-02-02
Fixed STM32L4 HAL, added IMAGE_HEADER_SIZE option

danielinux:
- Fixes in STM32L4 HAL driver:
    - `hal_flash_erase` was using the wrong page index while looping. Changed to page-range to call the HAL_ function once.
    - `hal_flash_erase` was calling `hal_flash_unlock()/hal_flash_lock()`, which is not allowed (flash is unlocked at a higher level and could include multiple subsequent operations)
    - `hal_flash_write` refactored using the STM32WB as base - which is well tested and guaranteed to work with unaligned data
   
- New feature in `tools/keytools/sign.py` <- it now tries to read `IMAGE_HEADER_SIZE` from `.config`, if available and set, to override the default (256B). 

Should fix issues in ZD 13566

### DIFF
--- a/hal/stm32l4.c
+++ b/hal/stm32l4.c
@@ -67,7 +67,7 @@ static uint32_t RAMFUNCTION GetBank(uint32_t Addr)
     return bank;
 }
 
-static void RAMFUNCTION hal_flash_clear_errors(void)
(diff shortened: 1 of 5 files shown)


dgarske (approved):
Wonderful! Thanks Daniele.

PR #176, opened 2022-02-02
Added basic test cases with renode

danielinux:
- Added renode test script
- Added `renode-factory-*` test targets and `renode-factory-all` to test secure boot on emulated STM32F40x
- Update tests not yet possible (looks like flash write is not yet suppored on this platform)


danielinux:
`.config` that can be used with `make renode-factory-all` so far:
 - No `config` : defaults to stm32F407-discovery
 - `config/examples/stm32f7.config`
 - `config/examples/hifive1.config`
 - `config/examples/nrf52840.config`

tools/scripts/renode-test-all.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+set -x
```
elms, line 2: Maybe `set -e` in to stop on first error. May require setting a trap to cleanup. Or you could break and set an exit code to call at the end.

tools/test-renode.mk
```diff
@@ -0,0 +1,139 @@
...
+endif
+ifeq ($(HASH),SHA3)
+  SIGN_ARGS+= --sha3
+endif
+
+# Testbed actions
+#
+#
+renode-on: FORCE
+	@rm -f /tmp/wolfboot.uart
```
elms, line 82: `${Q}` over `@` to allow showing commands.

tools/test-renode.mk
```diff
@@ -0,0 +1,139 @@
+TEST_UPDATE_VERSION?=2
+WOLFBOOT_VERSION?=0
+RENODE_UART=/tmp/wolfboot.uart
```
elms, line 3: Would it make sense to use `$TMP` isntead of `/tmp` or maybe even `mktemp` (may be less compatible)

tools/test-renode.mk
```diff
@@ -0,0 +1,139 @@
...
+RENODE_PORT=55155
+RENODE_OPTIONS=--pid-file=/tmp/renode.pid
```
elms, line 5: use variable for `/tmp/renode.pid` so other uses can refer to the variable

tools/test-renode.mk
```diff
@@ -0,0 +1,139 @@
...
+  SIGN_ARGS+= --sha3
+endif
+
+# Testbed actions
+#
+#
+renode-on: FORCE
+	@rm -f /tmp/wolfboot.uart
+	@renode $(RENODE_OPTIONS) $(RENODE_CONFIG) 2>&1 >/tmp/renode.log &
+	@while ! (test -e /tmp/wolfboot.uart); do sleep .1; done
```
elms, line 84: replace `/tmp/wolboot.uart` with `RENODE_UART` variable in all the targets

elms (changes requested):
This is really awesome. Very exciting to see this work. Just a few little things...and I want to test a little more.

I was able to test by running(Except riscv, I'm having issues with compiler setup)

```
make keytools
make tools/test-expect-version/test-expect-version
./tools/scripts/renode-test-all.sh
```

A few suggestions for the makefile:
 * targets depend on `test-expected-version`
 * Maybe out of scope.... Default to use build and use C keytools or detect if python package needs to be installed (`python -c "import wolfcrypt" &> /dev/null`) and suggest the command `pip install wolfcrypt`

danielinux:
Thanks @elms
fixed most points.

I don't think that defaulting to C tools should be part of this PR. 
Other tests use the same logic (if a compiled tool exist use it, otherwise fallback to the default .py), if we change this here we should align `tools/test*` too.

tools/scripts/renode-test-all.sh
danielinux: Removed `-x`. Disagree on `-e`, but I've added exit points on `make` invocations.

elms (approved):
LGTM.

I was still unable to test with riscv, but looks like it's an issue with my toolchain. Would be good to document the requirements and maybe even a docker image we can leverage for it.

PR #179, opened 2022-02-25
EFI: fixes and expand documenation

elms:
* Add EFI x86_64 include
* fix keytools build with EFI config
* Enable debug by default
* Expand `Targets.md` EFI section


elms:
Encountered some minor issues when testing this when preparing #180

rizlik (approved):

PR #183, opened 2022-03-14
Mitigate fault injections

danielinux:
Countermeasures against fault injections attacks (e.g. via voltage glitch, side channel, EMFI, etc.)


include/image.h
```diff
@@ -42,6 +42,11 @@
 #endif
 #endif
 
+#define likely(x)       __builtin_expect((x),1)
```
dgarske, line 45: You might need to wrap these in `__GNUC__`?

include/image.h
```diff
@@ -55,6 +60,282 @@
...
+    uint32_t hdr_ok;
+    uint32_t canary_FEED4567;
+    uint32_t signature_ok;
+    uint32_t canary_FEED6789;
+    uint32_t not_signature_ok;
+    uint32_t canary_FEED89AB;
+    uint32_t sha_ok;
+};
+
+static void __attribute__((noinline)) wolfBoot_image_confirm_signature_ok(struct wolfBoot_image *img)
```
dgarske, line 83: Suggest adding macro for noinline like we do in wolfSSL:

```
#ifdef noinline
    #define WB_NOINLINE noinline
#elif defined(_MSC_VER)
    #define WB_NOINLINE __declspec(noinline)
#elif defined(__ICCARM__) || defined(__IAR_SYSTEMS_ICC__)
    #define WB_NOINLINE _Pragma("inline = never")
#elif defined(__GNUC__) || defined(__KEIL__) || defined(__DCC__)
    #define WB_NOINLINE __attribute__((noinline))
#else
    #define WB_NOINLINE
#endif
```

include/image.h
```diff
@@ -55,6 +60,282 @@
...
+{
+    img->canary_FEED4567 = 0xFEED4567;
+    img->signature_ok = 1;
+    img->canary_FEED6789 = 0xFEED6789;
+    img->not_signature_ok = ~(1);
```
dgarske, line 88: Suggest using `1UL`

include/image.h
```diff
@@ -55,6 +60,282 @@
...
+    img->canary_FEED89AB = 0xFEED89AB;
+}
+
+#define PART_SANITY_CHECK(p) \
+    asm volatile("mov r2, #0"); \
```
dgarske, line 93: I recommend using `__asm__ __volatile__`. Also you'll need macros for:

```
#ifdef __IAR_SYSTEMS_ICC__
    #define __asm__        asm
    #define __volatile__   volatile
#endif
#ifdef __KEIL__
    #define __asm__        __asm
    #define __volatile__   volatile
#endif
```

src/image.c
```diff
@@ -38,51 +38,45 @@ static WOLFTPM2_DEV wolftpm_dev;
...
+            img->sha_hash, WOLFBOOT_SHA_DIGEST_SIZE, &res, &ed);
 }
 
 #endif /* WOLFBOOT_SIGN_ED25519 */
 
 #ifdef WOLFBOOT_SIGN_ED448
 #include <wolfssl/wolfcrypt/ed448.h>
 
-static int wolfBoot_verify_signature(uint8_t *hash, uint8_t *sig)
+static void wolfBoot_verify_signature(struct wolfBoot_image *img, uint8_t *sig)
```
dgarske, line 64: Please add code comment indicating the verify success is returning in `wolfBoot_image`.

dgarske (changes requested):
Great start, just a few tips for better compiler portability.

include/image.h
danielinux: Code has been made gcc-arm only

include/image.h
danielinux: ARMORED=1 code is now arm-gcc only

include/image.h
danielinux: This is no longer necessary as the armored code is arm-gcc only

src/image.c
danielinux: :+1:  (added in the caller since this function has many prototypes)

dgarske (changes requested):

dgarske (approved):

PR #195, opened 2022-04-08
Fixed SIGN=NONE. Improved tests

danielinux:
- Fixed SIGN=NONE option broken since #183 
- Improved renode tests: 
   - added SIGN=NONE cases
   - removed assembly cases with edwards curves
   - added boot time measurement to renode factory test


danielinux:
TODO: fix `make SIGN=NONE HASH=SHA3`

PR #199, opened 2022-04-21
Added support for NXP i.MX-RT105x

danielinux:
+ updated RT1060 support to latest SDK (2.11.0).


dgarske (approved):
Tested build for RT1060 and RT1050. Updated Targets.md with details. Thanks @danielinux

PR #223, opened 2022-07-21
Fixes for arm64/MMU targets

danielinux:
- Raspberry pi demo on qemu should now be working as expected.
- Fixed some regressions in UEFI after reworking rpi-3 target
- Fixed minor compile issues with the 'sim' target


dgarske (approved):
Confirmed works with Pi and Xilinx UltraScale+.

PR #235, opened 2022-08-22
RX72N support

kojo1:
MCU:          Renesas RX72N
Board:        RX72N/Envision Kit
IDE:          e2Studio
Compiler:     CCRX
FIT Module:   r_flash_rx
Flash access: hal/renesasRx.c

e2Studio Project:
wolfBoot         IDE/Renesas/e2studio_CCRX/wolfBoot
Sample app    IDE/Renesas/e2studio_CCRX/app_RenesasRX01


hal/renesas-rx.c
```diff
@@ -0,0 +1,176 @@
...
+static void longJump(unsigned long app_offset)
+{
+    jmp   r1;
+}
+
+void do_boot(const uint32_t *app_offset)
+{
+    longJump(app_offset);
+}
+#endif
```
danielinux, line 63: please do not redefine `do_boot`. You should link `boot_arm.o` instead. The function does more than just jumping to the app.
In particular, your version of `do_boot` does not:
 - set the stack pointer for the app/OS
 - change the address of the int vector for the app via VTOR (interrupt won't work in the OS you are booting)
 - set up the MPU

src/loader.c
```diff
@@ -32,6 +32,60 @@ static volatile const uint32_t __attribute__((used)) wolfboot_version = WOLFBOOT
...
+    state = *(part + WOLFBOOT_PARTITION_SIZE - sizeof(uint32_t) - 1);
+    wolfBoot_printf("Status:   %02x\n", state);
+    magic = part + WOLFBOOT_PARTITION_SIZE - sizeof(uint32_t);
+    wolfBoot_printf("Tail Mgc: %c%c%c%c\n", magic[0], magic[1], magic[2], magic[3]);
+
+#ifdef WOLFBOOT_PARTIION_VERBOS
+    v = (uint32_t *)part;
+    for(i = 0; i < 0x100/4; i++) {
+        if(i % 4 == 0)
+            wolfBoot_printf("\n%08x: ", (uint32_t)v+i*4);
```
danielinux, line 62: location is not guaranteed to be fixed. You should use TLV parser to retrieve versions (e.g. `wolfBoot_current_firmware_version` etc.)

danielinux (changes requested):
Great work adding RX72N target!

A few notes:
- `do_boot` should not be redefined. The new version introduced for this platform does not have many of the required features. If there are compatibility issues with your toolchain I'm happy to help.
- Usually debug printfs are not on in embedded targets. I don't mind as long as they can be compiled out, and if they are compiled out by default. A better choice would have been printing all the info from the application (the API from libwolfboot.c can retrieve image info and versions)
Please make PRINTF_ENABLED and WOLFBOOT_PARTITION_INFO off by default in your project if possible, or remove those entirely if they are not strictly needed. Activating UART/printf makes the code bigger and most of all increases the attack surface

- Nice documentation. I agree it should be under the `IDE` tree mostly because instructions are for compiling under windows, unlike all other targets except the IAR demo. Please add a mention in `docs/Target.md` that redirects to your `Readme.md`

hal/renesas-rx.c
kojo1: > please do not redefine `do_boot`. You should link `boot_arm.o` instead. The function does more than just jumping to the app. In particular, your version of `do_boot` does not:
> 
> * set the stack pointer for the app/OS
> * change the address of the int vector for the app via VTOR (interrupt won't work in the OS you are booting)
> * set up the MPU

The demo application includes BSP startup which takes care changing SP, interruption vector and other settings.
I am assuming you can simply jump to the entry point.

RX is not ARM architecture, but Renesas proprietary. Would it be good to have it in something like boot_RenesasRx.c?

kojo1:
>> Usually debug printfs are not on in embedded targets. 

I see it is usually not included, but Renesas e2Studio has by default. I was assuming nice to enable in the sample project.

hal/renesas-rx.c
danielinux: Sorry, my bad for missing the architecture of this port.

`do_boot` is part of the architecture code, not target. Yes in general it would make sense to have a boot_$(ARCH).c like all other architectures, but only worth it if we plan to port to other targets in the same $ARCH.

Usually the bring-up code is self-contained in wolfBoot. We define our own interrupt vector, which is empty apart for the critical faults, because we want all interrupt to be off during the bootloader stage. We usually write our own reset ISR to have the flow under control.

Relocation of the interrupt vector is also necessary before staging. Your simple application works even without setting a new offset, but if you loaded an OS or a demo that uses interrupts in your application code it won't probably trigger any interrupts because the offset has not been set. I suggest to do the same we do with other architectures to avoid this problem. The register equivalent to ARM `VTOR` seems to be `EXTB` in Renesas-RX.

danielinux:
> I see it is usually not included, but Renesas e2Studio has by default. I was assuming nice to enable in the sample project.

I agree, and I am in favor of having debug in the example application. I don't like debug prints in the bootloader code, for safety and code size reasons. All the data you print are available in the application space as well and can be printed after the application has been staged, avoiding to bring in kilobytes of libC code + device drivers in your bootloader code.

I see your current bootloader partition is 64KB, I wonder if it can be made smaller.

hal/renesas-rx.c
kojo1: Sorry, I have pushed boot_renesas_rx.c for do_boot before noticing your comment. I would arrange it either way you feel appropriate.

Regarding wolfBoot and the application configuration, I assume I am doing the way as your comment. 
wolfBoot and the sample application are totally independent, self-contained. Each has own interrupt vector appropriately located. e2Studio for Rx takes care the location and includes own startup routine for each to setup ISP, USP or IV, as long as you have a project for each. This looks like working, so far, and will see if it works for more complicated configurations.

Based on this configuration, I assume, wolfBoot can run totally interrupt disabled or its own very limited IV.

cconlon:
@kojo1 Please fix merge conflicts on this PR, then check with @danielinux to see if there are additional feedback items that need to be addressed here.  Thanks!

miyazakh:
Re-submit the PR as [PR313](https://github.com/wolfSSL/wolfBoot/pull/313)

PR #238, opened 2022-09-20
Encrypted self-update

danielinux:
Fixed encrypted self-update (bootloader update).
The combination between `ENCRYPT` and `RAM_CODE` requires some architecture specific changes. This PR introduces a custom linker script for l0 `hal/stm32l0_chacha_ram.ld` that moves the symbols from chacha into RAM. 

A description of the procedure along with instruction to port on other targets is provided in `docs/encrypted_partitions.md`

The self-update procedure has been tested on STM32L0 with external SPI flash, Chacha encryption, and confirmed also to work (via ZD14514).

### DIFF
--- a/hal/spi/spi_drv_stm32.h
+++ b/hal/spi/spi_drv_stm32.h
@@ -72,15 +72,26 @@
 #define APB2_CLOCK_ER     (*(volatile uint32_t *)(0x40021034))
 #define APB2_CLOCK_RST    (*(volatile uint32_t *)(0x40021024))
 #define RCC_GPIO_CLOCK_ER (*(volatile uint32_t *)(0x4002102C))
+#define GPIOA_BASE (0x50000000)
 #define GPIOB_BASE (0x50000400)
-#define SPI_GPIO    GPIOB_BASE
+
 #define SPI_CS_GPIO GPIOB_BASE
-#define SPI_CS_FLASH 8 /* Flash CS connected to GPIOB8 */
 #define SPI1_PIN_AF    0 /* Alternate function for SPI pins */
-#define SPI1_CLOCK_PIN 3 /* SPI_SCK: PB3  */
-#define SPI1_MISO_PIN  4 /* SPI_MISO PB4  */
-#define SPI1_MOSI_PIN  5 /* SPI_MOSI PB5  */
-#endif
+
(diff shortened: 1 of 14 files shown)


test-app/app_stm32l0.c
```diff
@@ -26,12 +26,49 @@
...
+		 "0123456789abcdef";
+
 void main(void) {
+    uint32_t version, updv;
+    volatile uint32_t i, j;
+#ifdef SPI_FLASH
+    spi_flash_probe();
+#endif
+    version = wolfBoot_current_firmware_version();
+    //updv = wolfBoot_update_firmware_version();
```
dgarske, line 51: What is purpose of this commented line? Should it be removed?

test-app/app_stm32l0.c
```diff
@@ -26,12 +26,49 @@
...
+
+    for (i = 0; i < version; i++) {
+        boot_led_on();
+        for (j = 0; j < 200000; j++)
+                ;
+        boot_led_off();
+        for (j = 0; j < 200000; j++)
+                ;
+    }
     boot_led_on();
```
dgarske, line 71: Not calling wolfBoot success anymore?

hal/spi/spi_drv_stm32.h
```diff
@@ -72,14 +72,15 @@
 #define APB2_CLOCK_ER     (*(volatile uint32_t *)(0x40021034))
 #define APB2_CLOCK_RST    (*(volatile uint32_t *)(0x40021024))
 #define RCC_GPIO_CLOCK_ER (*(volatile uint32_t *)(0x4002102C))
+#define GPIOA_BASE (0x50000000)
```
dgarske, line 75: Any concern about this change breaking any existing users?

dgarske (changes requested):

test-app/app_stm32l0.c
danielinux: only with even versions, to facilitate update/revert tests like on other platforms connected to jenkins

hal/spi/spi_drv_stm32.h
danielinux: You are right. I've reverted the original values and left an instance with the values I've used to test locally under ifdef.

dgarske (approved):

PR #245, opened 2022-10-21
Fixed bug in encrypted-delta update

danielinux:
Variable for return value was reused in delta+encrypt, leading to a double write at the wrong offset, causing to store corrupted patch images in the swap partion. 

This is the root cause for the bit flipping in SPI flash memory reported in ZD 15030.

This PR should fix the issue reported and encrypted delta updates should be working again.


danielinux:
Fix has been confirmed.

rizlik (approved):

PR #256, opened 2022-12-13
Fixes for encrypt/decrypt with unaligned address

### LINKED ISSUE #253: Prevent ext_flash_decrypt_read from writing more data than requested
@Tweakie
 Found a potential buffer overflow in wolfBoot.
If row_offset in ext_flash_decrypt_read is different than 0, ext_flash_decrypt_read will write len + ENCRYPT_BLOCK_SIZE - row_offset bytes to the buffer given to ext_flash_decrypt_read. If this buffer has the same size than len, ext_flash_decrypt_read will overwrite the data after this buffer.
I encountered this issue during an encrypted delta upgrade.

Note that ext_flash_decrypt_read will always write at least ENCRYPT_BLOCK_SIZE bytes to the given buffer. If the buffer is smaller than ENCRYPT_BLOCK_SIZE, a buffer overflow will occur.

Created this issue based on #252

dgarske:
* Fixes for encrypt/decrypt with unaligned address.
* Fix issue with reading byte count on a file using Mac OS.
* Cleanups for uart-flash-server.

Fixes #253

### DIFF
--- a/include/image.h
+++ b/include/image.h
@@ -98,7 +98,8 @@ struct wolfBoot_image {
  * With ARMORED setup, the flag is redundant, and the information is wrapped in
  * between canary variables, to mitigate attacks based on memory corruptions.
  */
-static void __attribute__((noinline)) wolfBoot_image_confirm_signature_ok(struct wolfBoot_image *img)
+static void __attribute__((noinline)) wolfBoot_image_confirm_signature_ok(
+    struct wolfBoot_image *img)
 {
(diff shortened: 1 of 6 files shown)


danielinux:
Confirmed no regressions on `test-enc-update` after applying on top of #262.

danielinux (approved):
Tested via `test-enc-update` - no regressions. To be merged with #262

PR #263, opened 2022-12-16
Improvements to gap fill

dgarske:
The default gap filling byte is `0xFF`. If using `FLAGS_INVERT=1` uses `0x00`. Can be overridden at build-time using `FILL_BYTE`.
Fixes ZD 15356


dgarske:
@jpbland1 can you review this PR also?

PR #269, opened 2023-01-03
fix assumed keysize error

jpbland1:
the key_import and keystore_add functions seem to be written when only 2048 size rsa keys were supported since key_import only reads blocks of 2048 size, which will return 0 and write no key for non RSA keys, and keystore_add takes the size of the buffer at face value instead of checking for a valid keysize. this change makes key_import read the keyfile regardless of size and keystore_add check the size of the key buffer and use the real public key size since ecc keys also include the private part of the key after the 2 public parts.

Reproducer:

```
make keytools
tools/keytools/keygen --rsa2048 -g rsa.der
openssl rsa -pubout -inform der -in rsa.der -outform der -out rsaPub.der
tools/keytools/keygen --rsa2048 -i rsaPub.der
```

The result is a `src/keystore.c` that contains a zero sized public key.

ZD 15412

### DIFF
--- a/tools/keytools/keygen.c
+++ b/tools/keytools/keygen.c
@@ -217,6 +217,7 @@ void keystore_add(uint32_t ktype, uint8_t *key, uint32_t sz, const char *keyfile
 {
     static int id_slot = 0;
     struct keystore_slot sl;
+
     if (ktype == KEYGEN_RSA2048 || ktype == KEYGEN_RSA3072 || ktype == KEYGEN_RSA4096)
         fprintf(fpub, Slot_hdr_int_size,  keyfile, id_slot, KType[ktype], sz);
     else
@@ -491,13 +492,54 @@ static void key_import(uint32_t ktype, const char *fname)
     uint8_t buf[KEYSLOT_MAX_PUBKEY_SIZE];
     FILE *f;
     int r;
+    int i;
+    int keySize = 0;
     f = fopen(fname, "rb");
     if (f == NULL) {
         fprintf(stderr, "Fatal error: could not open file %s to import public key\n", fname);
         exit(6);
     }
-    r = fread(buf, sizeof(buf), 1, f);
-    keystore_add(ktype, buf, r, fname);
+
+    switch (ktype) {
+        case KEYGEN_ED25519:
+            keySize = KEYSTORE_PUBKEY_SIZE_ED25519;
+            break;
+        case KEYGEN_ED448:
+            keySize = KEYSTORE_PUBKEY_SIZE_ED448;
+            break;
+        case KEYGEN_ECC256:
+            keySize = KEYSTORE_PUBKEY_SIZE_ECC256;
+            break;
+        case KEYGEN_ECC384:
+            keySize = KEYSTORE_PUBKEY_SIZE_ECC384;
+            break;
+        case KEYGEN_RSA2048:
+            keySize = KEYSTORE_PUBKEY_SIZE_RSA2048;
+            break;
+        case KEYGEN_RSA3072:
+            keySize = KEYSTORE_PUBKEY_SIZE_RSA3072;
+            break;
+        case KEYGEN_RSA4096:
+            keySize = KEYSTORE_PUBKEY_SIZE_RSA4096;
+            break;
+        default:
+            keySize = 0;
+    }
+
+    i = 0;
+
+    do
+    {
+        r = fread(buf + i, 1, 1, f);
+        i += r;
+    } while ( i < keySize && r > 0);
+
+    if (i <= 0 || i > keySize) {
+        printf("Fatal error: could not find valid key in file %s\n", fname);
+        exit(6);
+    }
+
+    keystore_add(ktype, buf, i, fname);
 }
 
 int main(int argc, char** argv)
--- a/tools/keytools/keygen.c
+++ b/tools/keytools/keygen.c
@@ -217,7 +217,7 @@ static uint32_t get_pubkey_size(uint32_t keyType)
 {
     uint32_t size = 0;
 
-    switch (ktype) {
+    switch (keyType) {
         case KEYGEN_ED25519:
             size = KEYSTORE_PUBKEY_SIZE_ED25519;
             break;
--- a/tools/keytools/keygen.c
+++ b/tools/keytools/keygen.c
@@ -339,6 +339,8 @@ static void keygen_ecc(const char *priv_fname, uint16_t ecc_key_size)
         exit(3);
     }
 
+    wc_ecc_free(&k);
+
     fpriv = fopen(priv_fname, "wb");
     if (fpriv == NULL) {
         fprintf(stderr, "Unable to open file '%s' for writing: %s", priv_fname,  strerror(errno));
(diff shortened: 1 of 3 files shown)


dgarske (changes requested):
```
Keytype: RSA2048
[75](https://github.com/wolfSSL/wolfBoot/actions/runs/3832568694/jobs/6523030688#step:9:76)
Gen wolfboot_signing_private_key.der
[76](https://github.com/wolfSSL/wolfBoot/actions/runs/3832568694/jobs/6523030688#step:9:77)
Generating key (type: RSA2048)
[77](https://github.com/wolfSSL/wolfBoot/actions/runs/3832568694/jobs/6523030688#step:9:78)
RSA public key len: 294 bytes
[78](https://github.com/wolfSSL/wolfBoot/actions/runs/3832568694/jobs/6523030688#step:9:79)
Invalid key size, expected at least 320, received 294
```

tools/keytools/keygen.c
```diff
@@ -491,13 +492,54 @@ static void key_import(uint32_t ktype, const char *fname)
...
+    int keySize = 0;
     f = fopen(fname, "rb");
     if (f == NULL) {
         fprintf(stderr, "Fatal error: could not open file %s to import public key\n", fname);
         exit(6);
     }
-    r = fread(buf, sizeof(buf), 1, f);
-    keystore_add(ktype, buf, r, fname);
+
+    switch (ktype) {
```
dgarske, line 503: This logic is duplicated above in `keystore_add`. Can this get consolidated to a single new function?

tools/keytools/keygen.c
```diff
@@ -491,13 +492,54 @@ static void key_import(uint32_t ktype, const char *fname)
...
+            keySize = 0;
+    }
+
+    i = 0;
+
+    do
+    {
+        r = fread(buf + i, 1, 1, f);
+        i += r;
+    } while ( i < keySize && r > 0);
```
dgarske, line 535: Loose the leading space. `(i < keySize && r > 0);`

tools/keytools/keygen.c
```diff
@@ -491,13 +492,54 @@ static void key_import(uint32_t ktype, const char *fname)
...
+        case KEYGEN_RSA4096:
+            keySize = KEYSTORE_PUBKEY_SIZE_RSA4096;
+            break;
+        default:
+            keySize = 0;
+    }
+
+    i = 0;
+
+    do
```
dgarske, line 531: `do {`

tools/keytools/keygen.c
```diff
@@ -491,13 +492,54 @@ static void key_import(uint32_t ktype, const char *fname)
...
+    {
+        r = fread(buf + i, 1, 1, f);
+        i += r;
```
dgarske, line 534: Don't add to `i` until r has been checked. It can be negative.

tools/keytools/keygen.c
```diff
@@ -491,13 +492,54 @@ static void key_import(uint32_t ktype, const char *fname)
...
+    } while ( i < keySize && r > 0);
+
+    if (i <= 0 || i > keySize) {
```
dgarske, line 537: Should `r` also be checked for `< 0`?

tools/keytools/keygen.c
```diff
@@ -491,13 +492,54 @@ static void key_import(uint32_t ktype, const char *fname)
...
     FILE *f;
     int r;
+    int i;
+    int keySize = 0;
     f = fopen(fname, "rb");
     if (f == NULL) {
         fprintf(stderr, "Fatal error: could not open file %s to import public key\n", fname);
         exit(6);
     }
-    r = fread(buf, sizeof(buf), 1, f);
```
dgarske, line 499: This works fine using `r = fread(buf, 1, sizeof(buf), f);`

dgarske (changes requested):
Please revert and just fix the original code to use the right `fread` arguments. I suspect the build slaves honor using it either way. My Mac was picky.

tools/keytools/keygen.c
jpbland1: yes, I was thinking i would wind up negative but not if it fails after reading a few bytes

tools/keytools/keygen.c
dgarske: Just this feedback item left.
Also can you investigate adding a test case for the import feature?

jpbland1: yes the switch in keystore_add assumes raw keys and needs to be called from the key generate function but since import is now expected to handle keys with headers we need to check the length here as well, if it's just ecc keys that have this difference I can remove the switch and just do if statements for the 3 ecc key types

dgarske: I just mean a new static function for the duplicate code. There are two matching switch statements.

jpbland1: ah you mean make a get key size function, thought you meant try to merge the two functions into a single since they had a similar part

rjbarnet:
Hello,

I've been testing these changes related to enabling the use of an ECC256 Key that is generated from OpenSSL. The changes here have addressed an issue that I experienced when updating the keystore.c file using `keygen.py` but I am now running into an issue when trying to generate the SHA256 digest binary file to sign. Here are the steps that I've done:

Generate Private Signing ECC256 Key with OpenSSL:

`openssl ecparam -out ecc256_signing_key.der -outform der -name prime256v1 -genkey`

Generate Public Key in DER Format with OpenSSL:

`openssl ec -in ecc256_signing_key.der -inform der -pubout -outform der -out ecc256_signing_pubkey.der`

Generate the keystore.c file:

```
$ python3 tools/keytools/keygen.py --ecc256 -i ecc256_signing_pubkey.der
keys to import:
['ecc256_signing_pubkey.der']
keys to generate:
[]
Output C file:        src/keystore.c
Public key slot:      0
Selected cipher:      ecc256
Input public key:     ecc256_signing_pubkey.der
```

Generate the SHA256 Digest:

```
$ python3 ./tools/keytools/sign.py --ecc256 --sha-only --sha256 app_image.bin ecc256_signing_pubkey.der 1
wolfBoot KeyTools (Python version)
wolfcrypt-py version: 5.5.3-0
Update type:          Firmware
Input image:          app_image.bin
Selected cipher:      ecc256
Private key:          ecc256_signing_pubkey.der
Output digest:        app_image_v1_digest.bin
Not Encrypted
Target partition id:  1 
Error: key size does not match any cipher
```

Note the error of `Error: key size does not match any cipher`

I've attached the application image, private signing key and the corresponding public key in DER format in the zip below.


[ecc256keys_image.zip](https://github.com/wolfSSL/wolfBoot/files/10354108/ecc256keys_image.zip)

rjbarnet:
With the latest changes - I am still getting an error (but new):

```
$ python3 ./tools/keytools/sign.py --ecc256 --sha-only --sha256 app_image.bin ecc256_signing_pubkey.der 1
wolfBoot KeyTools (Python version)
wolfcrypt-py version: 5.5.3-0
Update type:          Firmware
Input image:          app_image.bin
Selected cipher:      ecc256
Private key:          ecc256_signing_pubkey.der
Output digest:        app_image_v1_digest.bin
Not Encrypted
Target partition id:  1 
Traceback (most recent call last):
  File "./tools/keytools/sign.py", line 556, in <module>
    tmpEcc.decode_key(wolfboot_key_buffer)
  File "/home/rbarnett/projects/nsc/venv/lib/python3.8/site-packages/wolfcrypt/ciphers.py", line 1026, in decode_key
    raise WolfCryptError("Key decode error (%d)" % ret)
wolfcrypt.exceptions.WolfCryptError: Key decode error (-140)
```

dgarske (approved):

PR #275, opened 2023-01-20
Support for wolfcrypt in secure-mode from TrustZone-M secure domain

danielinux:
Support for non-secure callable crypto functions.

- Access crypto functions from non-secure application, through specific NSC methods
- PKCS11 interface in non-secure world (mapped through nsc)

Secure-mode wolfBoot integrates wolfPKCS11 and exposes a PKCS11 NSC API to perform crypto operations.

Private keys and secrets can be stored in secure-domain. Test NS application uses wolfCrypt PKCS11 support.


arch.mk
```diff
@@ -133,7 +134,15 @@ ifeq ($(ARCH),ARM)
...
     ifeq ($(TZEN),1)
-      CFLAGS += -mcmse
+      OBJS+=hal/stm32_tz.o
+      CFLAGS+=-mcmse
+      ifeq ($(WCSM),1)
+        SECURE_OBJS+=./src/wc_callable.o
+        SECURE_OBJS+=./lib/wolfssl/wolfcrypt/src/random.o
+        SECURE_OBJS+=./lib/wolfssl/wolfcrypt/src/asn.o
+        CFLAGS+=-DWOLFCRYPT_SECURE_MODE
+        SECURE_LDFLAGS+=-Wl,--cmse-implib -Wl,--out-implib=./src/wc_secure_calls.o
```
dgarske, line 144: Daniele, I don't see a file `./src/wc_secure_calls.c`? Should it be there or is this generated?

danielinux: the .o is generated using the `--out-implib` option, combined with `--cmse-implib` . That file is generated by the secure domain linking step, and included by non-secure domain application.

There is no source file associated to that .o. The .o will contain non-secure-callable entry points that can be called by non-secure applications.

Cfr: https://sourceware.org/binutils/docs/ld/ARM.html

> The ‘--cmse-implib’ option requests that the import libraries specified by the ‘--out-implib’ and ‘--in-implib’ options are secure gateway import libraries, suitable for linking a non-secure executable against secure code as per ARMv8-M Security Extensions. 

> [...]  if ‘--out-implib’ is specified, the symbols are added to the output import library. A warning is also given if some symbols present in the input import library have disappeared from the executable. This option is only effective for Secure Gateway import libraries, ie. when ‘--cmse-implib’ is specified.

dgarske (approved):

dgarske (approved):

PR #276, opened 2023-01-26
NVM_FLASH_WRITEONCE refactoring

danielinux:
- Using two sectors to keep partition/sector flags
- Keep two redundant set of flags, update one at a time
- Erase is done when the sector is old
- Flags update is faster because Erase is done in advance
- Accessing trailer information (including encryption keys) is done by selecting the newest information

Tested via renode, using nrf52 with NVM_FLASH_WRITEONCE flag on.

### DIFF
--- a/src/libwolfboot.c
+++ b/src/libwolfboot.c
@@ -255,7 +255,7 @@ int RAMFUNCTION hal_set_partition_magic(uint32_t addr)
 
 static uint8_t* RAMFUNCTION get_trailer_at(uint8_t part, uint32_t at)
 {
-    uint32_t sel_sec = 0
+    uint32_t sel_sec = 0;
 #ifdef NVM_FLASH_WRITEONCE
         nvm_select_fresh_sector(part);
 #endif
--- a/src/libwolfboot.c
+++ b/src/libwolfboot.c
@@ -133,6 +133,12 @@ static const uint32_t wolfboot_magic_trail = WOLFBOOT_MAGIC_TRAIL;
  *
  */
 
+#ifndef FLAGS_INVERT
+#define FLAG_CMP(a,b) ((a < b)? 0 : 1)
+#else
+#define FLAG_CMP(a,b) ((a > b)? 0 : 1)
+#endif
+
 #include <stddef.h>
 #include <string.h>
 static uint8_t NVM_CACHE[NVM_CACHE_SIZE] __attribute__((aligned(16)));
@@ -140,10 +146,11 @@ static int nvm_cached_sector = 0;
 
 static int nvm_select_fresh_sector(int part)
 {
+    int sel;
     uint32_t off;
     uint32_t base;
-    int sel;
     uint32_t addr_align;
+
     if (part == PART_BOOT)
         base = PART_BOOT_ENDFLAGS;
     else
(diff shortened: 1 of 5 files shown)


src/libwolfboot.c
```diff
@@ -125,63 +125,151 @@ static const uint32_t wolfboot_magic_trail = WOLFBOOT_MAGIC_TRAIL;
...
 static uint8_t NVM_CACHE[NVM_CACHE_SIZE] __attribute__((aligned(16)));
+static int nvm_cached_sector = 0;
+
+static int nvm_select_fresh_sector(int part)
+{
+    uint32_t off;
+    uint32_t base;
+    int sel;
+    uint32_t addr_align;
+    if (part == PART_BOOT)
```
dgarske, line 147: Add space above. Consider putting uint32_t and int declarations grouped.

src/libwolfboot.c
```diff
@@ -125,63 +125,151 @@ static const uint32_t wolfboot_magic_trail = WOLFBOOT_MAGIC_TRAIL;
...
-int RAMFUNCTION hal_set_partition_magic(uint32_t addr)
+static int RAMFUNCTION partition_magic_write(uint8_t part, uint32_t addr)
 {
     uint32_t off = addr % NVM_CACHE_SIZE;
     size_t base = (size_t)addr - off;
     int ret;
+    size_t addr_read, addr_write;
+    nvm_cached_sector = nvm_select_fresh_sector(part);
+    addr_read = base - (nvm_cached_sector * NVM_CACHE_SIZE);
+    addr_write = base - (!!!nvm_cached_sector * NVM_CACHE_SIZE);
```
dgarske, line 242: Again why `!!!nvm_cached_sector`? Is this a shortcut to convert to 0 or 1?

src/libwolfboot.c
```diff
@@ -787,10 +882,16 @@ static int RAMFUNCTION hal_set_key(const uint8_t *k, const uint8_t *nonce)
     uint32_t addr_align = addr & (~(WOLFBOOT_SECTOR_SIZE - 1));
     uint32_t addr_off = addr & (WOLFBOOT_SECTOR_SIZE - 1);
     int ret = 0;
+    int sel_sec = 0;
+#ifdef NVM_FLASH_WRITEONCE
+    sel_sec = nvm_select_fresh_sector(PART_BOOT);
+    addr_align -= (sel_sec * WOLFBOOT_SECTOR_SIZE);
```
dgarske, line 888: Should we check that `addr_align > (sel_sec * WOLFBOOT_SECTOR_SIZE)`?

test-app/app_nrf52.c
```diff
@@ -43,6 +43,10 @@
 #define UART0_TXD_MAXCOUNT *((volatile uint32_t *)(UART0_BASE + 0x548))
 #define UART0_BAUDRATE     *((volatile uint32_t *)(UART0_BASE + 0x524))
 
+
+__attribute__((section(".rodata")))
```
dgarske, line 47: Wrap with `__GNUC__`?

dgarske (changes requested):
Tested successfully on the STM32U575. Just a couple minor cleanup suggestions.

test-app/app_nrf52.c
```diff
@@ -43,6 +43,10 @@
...
+static const char extradata[1024 * 16] = "hi!";
```
jpbland1, line 48: what's the purpose of this?

src/libwolfboot.c
```diff
@@ -258,10 +347,16 @@ static void RAMFUNCTION set_partition_magic(uint8_t part)
 #else
 static uint8_t* RAMFUNCTION get_trailer_at(uint8_t part, uint32_t at)
 {
+    uint32_t sel_sec = 0;
+#ifdef NVM_FLASH_WRITEONCE
+    sel_sec = nvm_select_fresh_sector(part);
+#endif
     if (part == PART_BOOT)
```
jpbland1, line 354: I find it confusing when there's two lines and no braces, even if it is technically one statement

jpbland1 (commented):
This could be incompatible with previous versions already deployed since it expects 2 extra sectors to be free at the end of the partition right? For example if someone's app image took up space on on the sector before or on the same sector as the flags this would erase part of that image.

src/libwolfboot.c
danielinux: not sure. This comes from the constants at the beginning of the function, marking the end of the boot partition. Also, on most architectures the address of the boot partition does not start from 0, so the check will be true anyway even if the BOOT partition is that small?

test-app/app_nrf52.c
danielinux: actually this is not needed (default section for const variables is .rodata). For some reason, the app includes this only when DEBUG=1

test-app/app_nrf52.c
danielinux: Attempt to make the symbol appear in the app. This is actually redundant since this is the default section for const variables. The app will be >20KB only when DEBUG=1.

src/libwolfboot.c
danielinux: Hard to get rid of Linux kernel coding style. Added braces here for you now.

dgarske (approved):

PR #278, opened 2023-02-06
refactor c keytools

jpbland1:
previously keygen was generating a file, keystore.der, which actually contained the raw public key of one of the generated keys. keygen has been updated to put the public keys into der format and exports each key generated into its own der file

previously sign was not capable of using a public der for no sign and sha only modes and would not work correctly if the user brought their own key in der format. now it will detect if a key is encoded and decode it before using it, I've also cleaned up the load_key function so to use one switch statement. Instead of using many different if blocks to determine what algorithms are in use the auto mode will simply try every variation of keys until the key matches

adds tests for keygen to make sure keys generated and keys imported lead to the same keystore.c file


tools/efi/compile_efi_linux.sh
```diff
@@ -17,10 +17,10 @@ fi
 BR2_EXTERNAL=$(pwd)/tools/efi/br_ext_dir make -C $WORK_DIR/$BR_DIR tiny_defconfig O=$IMAGE_DIR
 make -C $WORK_DIR/$BR_DIR O=$IMAGE_DIR
 
-SIGN_TOOL="python3 ./tools/keytools/sign.py"
-if [ -f "./tools/keytools/sign" ]; then
+#SIGN_TOOL="python3 ./tools/keytools/sign.py"
```
dgarske, line 20: Why did this need to be commented out?

tools/keytools/keygen.c
```diff
@@ -326,31 +403,29 @@ static void keygen_rsa(const char *keyfile, int kbits)
...
+    ecc_key k[1];
     uint8_t Qx[MAX_ECC_KEY_SIZE], Qy[MAX_ECC_KEY_SIZE], d[MAX_ECC_KEY_SIZE];
     uint32_t qxsize = ecc_key_size,
              qysize = ecc_key_size,
              dsize =  ecc_key_size;
     uint8_t k_buffer[2 * MAX_ECC_KEY_SIZE];
     FILE *fpriv;
 
-    if (wc_ecc_make_key(&rng, ecc_key_size, &k) != 0) {
+    if (wc_ecc_make_key(&rng, ecc_key_size, k) != 0) {
```
dgarske, line 414: Please add `wc_ecc_init` and `wc_ecc_free`

tools/keytools/keygen.c
```diff
@@ -366,31 +441,33 @@ static void keygen_ecc(const char *priv_fname, uint16_t ecc_key_size)
...
 #ifdef HAVE_ED25519
 static void keygen_ed25519(const char *privkey)
 {
-    ed25519_key k;
+    ed25519_key k[1];
     uint8_t priv[32], pub[32];
     FILE *fpriv;
     uint32_t outlen = ED25519_KEY_SIZE;
-    if (wc_ed25519_make_key(&rng, ED25519_KEY_SIZE, &k) != 0) {
+    if (wc_ed25519_make_key(&rng, ED25519_KEY_SIZE, k) != 0) {
```
dgarske, line 462: Add `wc_ed25519_init`

tools/keytools/keygen.c
```diff
@@ -402,26 +479,28 @@ static void keygen_ed25519(const char *privkey)
...
 #ifdef HAVE_ED448
 static void keygen_ed448(const char *privkey)
 {
-    ed448_key k;
+    ed448_key k[1];
     uint8_t priv[ED448_KEY_SIZE], pub[ED448_PUB_KEY_SIZE];
     FILE *fpriv;
     uint32_t outlen = ED448_KEY_SIZE;
-    if (wc_ed448_make_key(&rng, ED448_KEY_SIZE, &k) != 0) {
+    if (wc_ed448_make_key(&rng, ED448_KEY_SIZE, k) != 0) {
```
dgarske, line 495: Add `wc_ed448_init`

tools/keytools/sign.c
```diff
@@ -70,6 +70,7 @@
 #endif
 #ifdef HAVE_ECC
     #include <wolfssl/wolfcrypt/ecc.h>
+    #define MAX_ECC_KEY_SIZE 66
```
dgarske, line 73: Why not use `ECC_MAXSIZE`?

tools/keytools/sign.c
```diff
@@ -127,6 +130,7 @@
 
 #define HDR_IMG_TYPE_AUTH_MASK    0xFF00
 #define HDR_IMG_TYPE_AUTH_NONE    0xFF00
+/*
```
dgarske, line 133: Why are these commented out?

dgarske (changes requested):
Very nice refactor!

tools/efi/compile_efi_linux.sh
jpbland1: I was testing with that, will revert

tools/keytools/keygen.c
jpbland1: will add init, free is already there at the end of the function

tools/keytools/sign.c
jpbland1: I copied this from keygen.c, will replace both

tools/keytools/sign.c
jpbland1: these are re definitions from wolfboot.h, I'll just delete them since I'm not including it

dgarske (dismissed):

danielinux (commented):
Changes look good. Please ensure that the new file names/types match the documentation.

In particular, please test the multi-step signature in Signing.md in the section "Signing Firmware with External Private Key (HSM)"

Also, ensure that keygen still creates a binary file (not keystore.der, documentation says keystore.img). This is used as alternative to keystore.c when the public key is stored elsewhere and not embedded in the code (user will have to provide the 3 functions of the keystore API to import the external key).

jpbland1:
I need some clarification on the `keystore.img` contents. It says `keystore.img` in the docs but there's nothing in the git logs about such a file being written, `keystore.der` was previously written but as I said for non rsa keys it was not in der format. It was being written as
```
-    fwrite(&sl, sl.pubkey_size, 1, fpub_image);
-    sl.pubkey_size = sz;
```
which seems wrong to me since it's writing `pubkey_size` but it starts from the start of the slot struct, which is not where the pubkey is, it's the last element since it's the largest. should I just be writing each pubkey to this `keystore.img` or should I be writing each `keystore_slot` struct?

danielinux:
That's correct, the entire array of structs must be exported into a file. The file name was wrong, not the content. It's OK to export the public key in .der format, but the .img file is intended to contain the (packed) structures

PR #281, opened 2023-02-21
Support for NXP IMXRT1064

bigbrett:
### Overview
This PR adds support for the NXP i.MX RT1064 chip family. It attempts to mimic the existing support code for the i.MX RT1050 and i.MX RT1060 families.

### Configuration
Specifying the RT1064 CPU for the `imx_rt` target is done in the same manner as for the already supported 1050 and 1060 CPU variants. Specifically, the following config options should be set differently from the 1050 and 1060 (see `config/examples/imx-rt1064.config` for the full example configuration).
```
MCUXPRESSO_CPU?=MIMXRT1064DVL6A
MCUXPRESSO_DRIVERS?=$(MCUXPRESSO)/devices/MIMXRT1064
WOLFBOOT_PARTITION_BOOT_ADDRESS?=0x70010000
WOLFBOOT_PARTITION_UPDATE_ADDRESS?=0x70030000
WOLFBOOT_PARTITION_SWAP_ADDRESS?=0x70050000
```

### Testing
Testing consisted of verifying that the test application (`test-app/app_imx_rt.c`) booted successfully on the MIMXRT1064-EVK development board. After building for the RT1064 and downloading `factory.bin` to flash, the user is able to see "wolfBoot Test app, version = 1" printed on the UART console. 

### Notes/questions
- I didn't see any contributing guidelines, so please let me know if I'm missing any key steps or items in this PR! 
- This PR also fixes what I think is a bug in the original 1050/1060 support: [`arch.mk` previously added the 1050 board files to the include path for the 1060 CPU configuration](https://github.com/wolfSSL/wolfBoot/blob/master/arch.mk#L250), and vis versa. Since the two dev boards are similar, it probably didn't cause any issues, but it is fixed in this PR and extended to include the 1064. Let me know if this was for some reason intentional and I can roll it back. 
- I removed `-I$(MCUXPRESSO)/middleware/mflash/mimxrt1062` from the include path, as it is no longer exists in the NXP SDK. Let me know if we should try and maintain backward compatibility 
- The current makefile architecture treats all of the i.MX RT variants as one `TARGET` with the subtle differences in code and config switched out based on the `MCUXPRESSO_CPU` configuration setting  (and associated `CPU_MIMXRTXXX` macro). This assumes all target variants would use the same linker file, which is not the case for the RT1064 with its slightly different memory map. To address this, [I've modified the makefile to parse out and replace the flash base address with `ARCH_FLASH_OFFSET` just like it does for other values](https://github.com/bigbrett/wolfBoot/blob/nxp_imxrt1064_support/Makefile#L189). This _shouldn't_ break any other targets, since they don't mark that value for replacement, but if it does, I can figure out another way to do it.
- I was unable to get the build system to recognize the system include path of my cross-compiler (arm-none-eabi-gcc) toolchain since the user can't "append" to `CFLAGS` without overriding it. It looks like there is an attempt to allow the user to specify a path through the [`CROSS_COMPILE_PATH` variable](https://github.com/wolfSSL/wolfBoot/blob/master/arch.mk#L402), but this still hardcodes a path that might not match the users install (for example, this doesn't reflect the layout of the `arm-none-eabi-gcc` toolchains that I have typically downloaded from ARM) . It might be wise to allow the user to explicitly specify the path to the include directory. I worked around this by just manually specifying the include path in the makefile. Please me know if I missed something and there is a better way to do this!


dgarske:
Hi @bigbrett ,

Thank you for submitting this pull request! I am looking forward to reviewing it.

Can you tell us more about your project and use-case?

For code contributions we require a signed contributor agreement. Please email support@wolfssl.com with a link to this PR and we'll get you the document to sign.

Thanks,
David Garske, wolfSSL

bigbrett:
Hi @dgarske,

No specific project/use case in mind for now. Mostly just feeling guilty that the RT1064-EVK in my desk drawer has been gathering dust over the last few months, so figured I'd show it a good time.

Let me know if there is anything I need to do on my end to get the `renode` CI steps that are failing to pass...from a quick glimpse it doesn't look like you currently support any testing of the `imx_rt` targets in `renode` (just stm32, nrf52 and riscV parts, right?) so I didn't do anything with it. 

Let me know if that is an incorrect assumption, and if I should be doing anything different there.

hal/imx_rt.ld
```diff
@@ -1,7 +1,7 @@
 /* Specify the memory areas */
 MEMORY
 {
-  FLASH(rx)  : ORIGIN = 0x60000000, LENGTH = @BOOTLOADER_PARTITION_SIZE@
+  FLASH(rx)  : ORIGIN = @ARCH_FLASH_OFFSET@, LENGTH = @BOOTLOADER_PARTITION_SIZE@
```
bigbrett, line 4: @dgarske I referenced this in the description, but let me know if this (and the corresponding `sed` command in the top-level Makefile) isn't okay for some reason, or breaks anything. The issue is that the build infra assumes one linker script per target, but the RT1064 has a slightly different memory map than the RT1050 and RT1060. Specifically, the RT1064 uses `FlexSPI2` to access the in-package QSPI flash (which it boots from, and is mapped to `0x7000_0000`), whereas the 1050 and 1060 use FlexSpi1 (mapped to `0x6000_0000`). Let me know if this is too cursed and I can find a better way.

embhorn:
@bigbrett you are approved as a wolfSSL contributor

hal/imx_rt1064_nor.h
```diff
@@ -0,0 +1,252 @@
+
+/* wolfBoot interface for iMX-RT1060-EVK.
+ *
+ * Description of the NOR configuration interface required by the board.
+ *
+ * Copyright (C) 2021 wolfSSL Inc.
```
danielinux, line 6: check copyright year

hal/imx_rt.ld
danielinux: This is the cleanest way to do this

hal/imx_rt1064_nor.h
```diff
@@ -0,0 +1,252 @@
...
+#define FLEXSPI_1PAD 0
+#define FLEXSPI_2PAD 1
+#define FLEXSPI_4PAD 2
+#define FLEXSPI_8PAD 3
+
+#define FLEXSPI_LUT_SEQ(cmd0, pad0, op0, cmd1, pad1, op1)                                                              \
+    (FLEXSPI_LUT_OPERAND0(op0) | FLEXSPI_LUT_NUM_PADS0(pad0) | FLEXSPI_LUT_OPCODE0(cmd0) | FLEXSPI_LUT_OPERAND1(op1) | \
+     FLEXSPI_LUT_NUM_PADS1(pad1) | FLEXSPI_LUT_OPCODE1(cmd1))
+
+//!@brief Definitions for FlexSPI Serial Clock Frequency
```
danielinux, line 82: Please use C-style comments only (`/*`)

danielinux: throughout the file

danielinux (changes requested):
Changes look good. Just a few minor/cosmetic changes requested.

hal/imx_rt1064_nor.h
bigbrett: updated to 2023

hal/imx_rt1064_nor.h
bigbrett: Done.

This does bring up one issue I should mention: since that file is almost entirely driver code pulled from the MCUXpresso SDK (`boards/{board}/xip/{board}_flexspi_nor_config.h`), changing all of the comments (or any other similar change) will result in a very messy file diff if you wish to upgrade to a newer SDK version in the future and check what changed. 

The same situation exists right now with the 1050 and 1060 ports in their respective flexpi NOR configuration files (`imx_rt10{50,60}_nor.h`), which is why I did it the way I did from the start. **If adhering to a particular comment style is required, we might want to revisit those files and change the style too?** Happy to do that as part of this PR or introduce it in its own PR.

The real question is whether these NOR configuration files should even be part of wolfBoot source code, or instead just referenced as an external SDK file. If we want to treat it as vendor code/external, we should get rid of the `hal/imx_rtXXXX_nor.h` files in their entirety (since they are just SDK code anyway) and instead have `hal/imx_rt.c` directly `#include "{board}_flexspi_nor_config.h"` from the SDK, reducing copy-pasted code. 

We can always introduce the refactor mentioned above in a later PR, since it would increase the scope and require retesting on all supported platforms. Just food for thought for future improvement.

dgarske: Hi @bigbrett , there is also the issue or intellectual property. If you can reference the NXP SDK for these headers directly please do. If not then these should be re-written. Ping me directly if you'd like to discuss.

danielinux: The files in question do not generate any code, and are only interfaces to the NXP flash drivers API.

I agree however that these files should be referenced externally when linking to the NXP SDK rather than be part of the code distributed in the wolfBoot package.

@bigbrett feel free to remove these for 1050 and 1060 as well, and ensure that they can be found when building with the given configurations according to the include path.

bigbrett: @danielinux @dgarske sounds like we are all in agreement then. I just pushed a commit removing the file entirely for all three targets (1050, 1060, and 1064). All three targets build without issues but am only able to test the 1064 port on hardware.

arch.mk
```diff
@@ -266,7 +266,7 @@ ifeq ($(TARGET),imx_rt)
 
   ifeq ($(MCUXPRESSO_CPU),MIMXRT1052DVJ6B)
     ARCH_FLASH_OFFSET=0x60000000
-    CFLAGS+=-I$(MCUXPRESSO)/boards/evkmimxrt1050/xip/
+    CFLAGS+=-I$(MCUXPRESSO)/boards/evkbimxrt1050/xip/
```
bigbrett, line 269: One last thing to note: the `EVK-MIMXRT1050` eval board was deprecated in early 2018, as it had preliminary silicon, and replaced by the `EVKB-MIMXRT1050`. The last SDK version supporting this target was v2.3.

I assume the EVK**B** is the board rev you wanted to target anyway, since `config/examples/imx-rt1050.config` points to an SDK package for the EVKB, but `arch.mk` had a path expecting the EVK board, causing some compiler errors. I fixed it by correcting this path. 

If you need to maintain support for the original EVK because of a customer requirement, let me know.

danielinux:
Thanks @bigbrett ! I'm moving this to ready to review" and run some final tests

danielinux (approved):

danielinux:
Great addition. Thanks @bigbrett !

PR #288, opened 2023-03-13
Encryption unit tests

danielinux:
- Added unit tests for encrypt/decrypt functions with EXT_ENCRYPTED + EXT_FLASH
- Fixed two issues detected

### DIFF
--- a/tools/unit-tests/unit-extflash.c
+++ b/tools/unit-tests/unit-extflash.c
@@ -78,12 +78,12 @@ int hal_flash_erase(uint32_t address, int len)
 }
 void hal_flash_unlock(void)
 {
-    //fail_unless(locked, "Double unlock detected\n");
+    fail_unless(locked, "Double unlock detected\n");
     locked--;
 }
 void hal_flash_lock(void)
 {
-    //fail_if(locked, "Double lock detected\n");
+    fail_if(locked, "Double lock detected\n");
     locked++;
 }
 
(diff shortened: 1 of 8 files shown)


tools/unit-tests/unit-extflash.c
```diff
@@ -0,0 +1,287 @@
...
+    /* Copy the data from the input buffer to the flash memory */
+    memcpy(&flash[address], data, len);
+
+    return 0;
+}
+
+int ext_flash_erase(uintptr_t address, int len) {
+    /* Check that the erase address and size are within the bounds of the flash memory */
+    ck_assert_int_le(address + len, FLASH_SIZE);
+
```
rizlik, line 124: does it make sense to add a check that both address and len are aligned to WOLFBOOT_SECTOR_SIZE?
otherwise the memset below should be modified to also erase the last partial block

tools/unit-tests/unit-extflash.c
```diff
@@ -0,0 +1,287 @@
...
+    ck_assert_int_eq(rres, size);
+    
+    /* Check that the data read from the flash memory matches the data that was written */
+    ck_assert_mem_eq(&flash[address], test_buffer, size);
+
+}
+END_TEST
+
+
+/* End Mocks */
```
rizlik, line 255: Is this misplaced?

rizlik (changes requested):

rizlik (approved):

PR #289, opened 2023-03-14
Parallelize renode tests execution, remove unfeasible test, add compile-time check for large stack usage

danielinux:
- Renode tests running in parallel to accelerate test batch
- compile-time check for unfeasible configuration (RSA4096 + TFM requiring 70KB of stack)
- added WOLFBOOT_HUGE_STACK to allow large stack allocations


src/image.c
```diff
@@ -179,12 +179,20 @@ static void wolfBoot_verify_signature(uint8_t key_slot,
...
+
 #if defined(WOLFBOOT_SIGN_RSA2048) || \
         defined (WOLFBOOT_SIGN_RSA3072) || \
         defined (WOLFBOOT_SIGN_RSA4096)
 #include <wolfssl/wolfcrypt/asn.h>
 #include <wolfssl/wolfcrypt/rsa.h>
 
+#if defined(WOLFBOOT_SIGN_RSA4096) && \
+    (defined(USE_FAST_MATH) && \
+     !defined(WOLFSSL_SMALL_STACK) && !defined(WOLFBOOT_HUGE_STACK))
```
billphipps, line 191: Recommend to remove `!defined(WOLFSSL_SMALL_STACK) &&` .  Seems like if small_stack IS defined OR if huge_stack is NOT defined then you should error out with the note.   I see above where both small and huge cannot be both set, so this logic will work provided the above guarantees only one of them is set.  Perhaps we should consider stack_size=1,2,3, for small, normal, huge?

BTW.  This is exactly what was causing the STM32C0 to fail at runtime due to only having 12kB of RAM.  :)

dgarske (approved):

PR #295, opened 2023-03-30
Renesas RA6M4 support

miyazakh:
Board : Renesas RA6M4
Compiler : gcc
IDE : e2studio

Not SCE(Secure Cryptography Engine) use yet. Only SW use


miyazakh:
For SCE use case, its effort estimation will be done first.

cconlon:
@danielinux Can you do a review of Hide's RA6M4 work here?  He will be adding SCE hardware crypto support next, but I don't think that should block our review on this PR for now.

danielinux (approved):

PR #302, opened 2023-05-04
Support for NXP P1021 eSPI and TPM (and fixes for eLBC NAND)

dgarske:
* Fix for P1021 to disable XIP support. This enables `PART_BOOT_EXT` to correctly load the boot partition from external flash.
* Fix the eLBC NAND flash driver issues with page/col selection. 
* Allow override of the `WOLFBOOT_SHA_BLOCK_SIZE`.
* Refactor the TPM SPI interface to enable full transfers.
* Update the new `spi_xfer` to include a "continue" flag to allow leaving the CS asserted.
* Enable support for calling into target HAL for SPI. 
* Added P1021 TPM test.
* Reduce first stage init code (don't relocate CCSRBAR or invalidate all TLB's).
* Improve udelay to use timebase.
* Fix L1 cache line sizes.
* Fix L2ADDR to enable all 256KB.
* Use `do_boot` not the jump to function pointer.
* Switch PPC test app linker script to use `WOLFBOOT_LOAD_ADDRESS`.
* Improvements to the clock calculation code (dynamic based on ratio).
* Make stage1 PIC.
* Disable L1/L2 for first stage.
* Add GOT2 to flash region.
* Fix to use correct boot ROM entry (`0xFFFFF000`) in stage1 linker script.

```
wolfBoot HAL Init
Flash Init: Ret 0, ID 0x76207620
Part: Active 0, Address 0
Boot partition: 200000
Image size 3964
Firmware Valid
Loading 3964 bytes to RAM at 1E000000
Booting at 1E000000

Test App

0x00000001
0x00000002
0x00000003
0x00000004
0x00000005
0x00000006
```


hal/nxp_p1021.c
```diff
@@ -420,6 +423,18 @@ enum elbc_amask_sizes {
 #define ESPI_CSMODE_CSAFT(x) (((x) & 0xF) << 8)  /* CS assertion time in bits after frame end */
 #define ESPI_CSMODE_CSCG(x)  (((x) & 0xF) << 3)  /* Clock gaps between transmitted frames according to this size */
 
+#if defined(ENABLE_ESPI) || defined(ENABLE_DDR)
+static void udelay(uint32_t delay_us)
+{
+    uint32_t i;
+    static const uint32_t oneus = (400000000 / 1000000);
```
jpbland1, line 430: I think we should use a macro for clock speed of the chip to do this calculation

jpbland1 (changes requested):

hal/nxp_p1021.c
dgarske: Will fix. It is a macro.

jpbland1 (dismissed):

include/spi_drv.h
```diff
@@ -49,15 +49,23 @@
...
+#ifdef SPI_FLASH
 void spi_cs_on(uint32_t base, int pin);
 void spi_cs_off(uint32_t base, int pin);
 void spi_write(const char byte);
 uint8_t spi_read(void);
-#endif /* SPI_FLASH || WOLFBOOT_TPM */
+#endif
+
+#ifdef WOLFBOOT_TPM
+int spi_xfer(int cs, const uint8_t* tx, uint8_t* rx, uint32_t sz);
```
billphipps, line 67: Hrmmm.  I think we may need an interface that allows continued transactions where the CS does not get deserted, specifically when the total size of the transaction is unknown at the beginning, like with wait states.

hal/nxp_p1021.c
```diff
@@ -380,6 +390,133 @@ enum elbc_amask_sizes {
...
+    blks = (sz + 3) / 4;
+    while (blks--) {
+        /* wait till TX fifo has room */
+        while ((get32(ESPI_SPIE) & ESPI_SPIE_TNF) == 0);
+        reg = *((uint32_t*)tx);
+        set32(ESPI_SPITF, reg);
+        tx += 4;
+        set32(ESPI_SPIE, ESPI_SPIE_TNF); /* clear event */
+
+        udelay(5);
```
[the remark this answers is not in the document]
dgarske: Yeah I tried using DON also. No luck

hal/nxp_p1021.c
```diff
@@ -380,6 +390,133 @@ enum elbc_amask_sizes {
...
+    set32(ESPI_SPCOM, ESPI_SPCOM_CS(cs) | ESPI_SPCOM_TRANLEN(sz));
+
+    set32(ESPI_SPIE, 0xffffffff); /* Clear all eSPI events */
+
+    /* calculate the number of 4 byte blocks rounded up */
+    blks = (sz + 3) / 4;
+    while (blks--) {
+        /* wait till TX fifo has room */
+        while ((get32(ESPI_SPIE) & ESPI_SPIE_TNF) == 0);
+        reg = *((uint32_t*)tx);
```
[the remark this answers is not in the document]
dgarske: I agree, this needs fixed up, since I WILL add wait state support, since customer is using a Microchip ATTPM20.

include/spi_drv.h
dgarske: I agree and the eSPI has the ability to assert / dessert the CS separately.

billphipps (approved):

src/boot_ppc_start.S
```diff
@@ -123,12 +123,10 @@ Also see MPC8544ERM
 
 _reset:
 
-#ifdef DEBUG
```
dgarske, line 126: Thanks @jpbland1 for spotting this. It was on purpose for release testing. I will investigate impact of this.

PR #303, opened 2023-05-05
add script for preseal a public key to tpm

jpbland1:
add more detailed instructions on how to prepare a TPM keystore build and how to use the preseal script

tested in simulator for first time boot with unseal


tools/preseal/Makefile
```diff
@@ -0,0 +1,12 @@
...
+
+all: preseal
+
+preseal:
+	$(Q)$(CC) -o $@ $@.c -lwolftpm
+
+debug:
+	$(Q)$(CC) -o preseal preseal.c -g -lwolftpm -lwolfssl
+
+clean:
```
dgarske, line 11: Looks like you added the binary. It should not be tracked and it should be in the .gitignore.

tools/preseal/preseal.c
```diff
@@ -0,0 +1,195 @@
+#include <wolfssl/options.h>
```
dgarske, line 1: Please add our standard license header

tools/preseal/preseal.c
```diff
@@ -0,0 +1,195 @@
...
+#include <wolfssl/wolfcrypt/settings.h>
```
dgarske, line 2: Does this tool really work against a physically different TPM? Or does it need to be run using the same TPM??

dgarske (changes requested):

tools/preseal/preseal.c
jpbland1: It needs to run on the same TPM since it needs to seal the public key to NV

dgarske (approved):

PR #306, opened 2023-05-11
Initial support for LS1028A.  Booting into OCRAM app.

billphipps:
Still getting through some problems.  Here's what's working:
1. Booting without TPM into OCRAM-based app.  Used NO_XIP=1 to make wolfBoot copy into the load address
2. Booting with TPM sealing is working, except I haven't gotten the initial key into the TPM yet, so it fails to unseal and verify the signature.  TPM access is working fine.
3. I updated spi_drv_nxp to gate on platform.  There are some shenanigans that should be cleaner up as we stabilize the API.

Here's some hacks that I wasn't sure about:
1. Updated stack size usage test in options.mk because it checks for TPM first (limits to 6k) but with SP-math==0, we really need 7.5k.  This needs a systemic update to handle max() of the options.
2. Ignored DTS stuff for now.  Didn't seem to cause an issue.
4. With TPM enabled, there's a duplication of some functions with wolf crypt/misc.c.  I added a guard on WOLFBOOT_TPM.
5. Skipped getting the DDR working for reals as we are targeting a simple OCRAM app for now.
6. Updated the wolfTPM submodule to the latest and greatest.  This got sealing working.
7. I snuck the wolfBoot and app into the SPI NOR flash after the RCW and before the uBoot.  Simply updated the start address in the RCW to point to wolf boot instead of uBoot.  Can easily swap it back to get into uBoot and linux.  Likely need this to preload the TPM key...

Here's the list of things that are broke
1. Measured boot now fails to compile.  Likely a simple typo somewhere, but I didn't dig into it too far.
2. Test app is still likely still running at EL3.  This must be pushed down.
3. MMU tables are likely not correct for anything past 512MB.  Need to use the ones from CW
4. Documentation is still out of date.  Prepping slides tomorrow (today) to add pictures of what we are doing and what we should/could do instead.
5. Still need to clean up the initial startup sequence and get it closer to the fully functions CW version.

Overall, pretty darn close and likely suitable for the customer demo.  Glide path seems clear on the technical issues.


hal/nxp_ls1028a.c
```diff
@@ -0,0 +1,1370 @@
+/* ls1028a.c
+ *
+ * Copyright (C) 2021 wolfSSL Inc.
```
dgarske, line 3: 2023

options.mk
```diff
@@ -65,7 +65,8 @@ ifeq ($(SIGN),ECC256)
        STACK_USAGE=4096
   else
     ifeq ($(WOLFTPM),1)
-      STACK_USAGE=6680
+      ## Was 6680. Increasing to handle SPMATH=0
+      STACK_USAGE=7600
```
dgarske, line 69: This is okay to increase or gate on the SPMATH the two options.

dgarske (changes requested):
Looks great! Just cleanups, but otherwise okay to merge.

config/examples/nxp-ls1028a.config
```diff
@@ -0,0 +1,66 @@
...
+NVM_FLASH_WRITEONCE?=0
+WOLFBOOT_VERSION?=0
+V?=0
+NO_MPU?=0
+SPMATH?=1
+RAM_CODE?=0
+DUALBANK_SWAP?=0
+PKA?=0
+
+WOLFTPM?=0
```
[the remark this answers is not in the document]
billphipps: This was a config that I was using without TPM support.  Note compilation fails if ONLY WOLFBOOT_TPM is enabled without MEASURED_BOOT nor TPM_KEYSTORE, which was a suprise

jpbland1 (changes requested):
IDE/Renesas/e2studio/RA6M4/app_RA/src/SEGGER_RTT/myprint.c empty file

include/wolfboot/wolfboot.h
```diff
@@ -195,7 +195,11 @@ extern "C" {
 
  /* KeyStore API */
  int keystore_num_pubkeys(void);
+#if defined(WOLFBOOT_RENESAS_SCEPROTECT)
```
jpbland1, line 198: don't do this, just cast your buffer to uint32_t* in one place if you need it in that form

danielinux:
@billphipps @dgarske is this still going or should I close as stale?

billphipps:
I'll close it as stale.  I think David pushed better versions of the shared files already.

dgarske:
I believe this is still the latest LS1028A work and it is still on my list to work on. In fact I will likely be working on the Xilinx UltraScale+ and plan to pull this work in for that. Closing is fine for now

PR #308, opened 2023-05-15
Improved simulator tests

danielinux:
- internal flash now mapped at 0xC0000000 (using `ARCH_FLASH_OFFSET`)
- wolfBoot binary included in internal flash (wolfboot partition is also now writeable to test self-updates and encryption)
- sim test app supports encryption (sets the key for test)
- added new target to store the update in the external flash
- external flash mapped to 0xD0000000


jpbland1 (approved):
Tested working

PR #309, opened 2023-05-16
NVM_FLASH_WRITEONCE Fixed wrong partition erase

danielinux:
When using NVM_FLASH_WRITEONCE, before writing to the active trailer sector, we check that the sector has been previously erased. This check was erasing the wrong partition trailer, leaving the system without a trailer for a few lines of code, but potentially vulnerable to powerfailure events.

Reported by @jpbland1


jpbland1 (approved):
looks good

PR #311, opened 2023-05-16
add a simulator test for using encryption and delta updates

jpbland1:
increase argument count max since the tools duplicate the sig and hash args


danielinux:
Please add corresponding action in `.github/workflows/test-powerfail-simulator.yml` like with other cases

danielinux (commented):
please add .github action for this new scenario

danielinux (approved):
:rocket: Thanks @jpbland1

PR #312, opened 2023-05-22
add missing preseal files, build preseal tools using the lib wolfssl and lib wolfTPM

jpbland1:


billphipps:
Consider adding a README.md that either points to Preseal.md or explains that preseal should be compiled to run on a system that has direct access to the target TPM.  I think the sign and key gen tools don't need to be run on the target, so they may need to be built for the host environment.  Also recommend changing the names of the tools so they don't conflict with existing tool names.

.gitignore
```diff
@@ -70,9 +70,13 @@ tools/keytools/x64
 tools/keytools/Debug
 tools/keytools/Release
 
-# preseal binary
+# preseal binaries
 tools/preseal/preseal
 tools/preseal/preseal.exe
```
billphipps, line 75: Are these *.exe files actually created?  If so, they should be removed during make clean

tools/preseal/keygen.c
```diff
@@ -0,0 +1,229 @@
...
+
+    rc = write(fd, qy, qySz);
+    if (rc != (int)qxSz) {
+        printf("write qx failed\n");
+        goto exit;
+    }
+
+    close(fd);
+
+    printf("Policy Public Key: policy-public-key.raw\n");
```
billphipps, line 137: Consider also exporting the policy-private-key.raw in case the user needs to verify proper key generation.

tools/preseal/preseal.c
```diff
@@ -33,6 +33,7 @@
 
 #define DEFAULT_PCR_INDEX 16
 
+#ifndef NO_PRESEAL_FILESYSTEM
```
billphipps, line 36: stdio.h, fcntl.h and unistd.h may need to be guarded from inclusion as well if there is no filesystem.

tools/preseal/sign.c
```diff
@@ -0,0 +1,137 @@
...
+void usage(void) {
+    printf("Sign an image digest with an ecc private key:\n");
+    printf("./sign private-key digest");
+}
+
+int main(int argc, char** argv)
+{
+    int rc;
+    int fd;
+    uint8_t eccBuff[256];
```
billphipps, line 25: This should be sized 3*ECC_KEY_SIZE, no?  Then the check below could ensure the private key was the correct size.

billphipps (approved):
I'll grab this and try it on the LS1028A so we can add that to the testing section.

tools/preseal/Makefile
```diff
@@ -1,9 +1,21 @@
...
-CFLAGS:=
+-include ../../.config
 
-ifneq ($(NO_FILESYSTEM),)
+CC      = gcc
+WOLFBOOTDIR = ../..
+WOLFDIR = $(WOLFBOOTDIR)/lib/wolfssl/
+WOLFTPMDIR = $(WOLFBOOTDIR)/lib/wolfTPM/
+CFLAGS  = -Wall -Wextra -Werror
+CFLAGS  += -I. -DWOLFSSL_USER_SETTINGS -DWOLFTPM_USER_SETTINGS -I$(WOLFDIR) -I$(WOLFBOOTDIR)/include
```
danielinux, line 8: missing `-I$(WOLFTPMDIR)` in CFLAGS

tools/preseal/user_settings.h
```diff
@@ -0,0 +1,107 @@
...
+#define WOLFSSL_IGNORE_FILE_WARN
+#define NO_ERROR_STRINGS
+
+#define BENCH_EMBEDDED
+#define NO_CRYPT_TEST
+#define NO_CRYPT_BENCHMARK
+
+#define WOLFSSL_PUBLIC_MP
+#define WOLFSSL_AES_CFB
+
```
danielinux, line 106: Are we missing some WOLFTPM config here? Can we default to LINUX_DEV for the tool?

tools/preseal/Makefile
danielinux: ... And -I$(WOLFBOOTDIR)/include should be removed, or you are including wolfBoot's `user_settings.h` instead of the local one that seems to be created for the tool

danielinux:
during "preseal" provisioning, why are you extending the PCR with a specific digest from an image? Would this not bind the provisioning to the specific image version? Why does preseal depend on a specific image version?

billphipps:
> during "preseal" provisioning, why are you extending the PCR with a specific digest from an image? Would this not bind the provisioning to the specific image version? Why does preseal depend on a specific image version?

This design is an initial solution that does not depend (yet) on the measured boot feature being enabled.  A better solution would be to measure wolfBoot and the hardware into PCR's and use those measurements to bind the image verification public key into the TPM.  But, this would specifically require wolfBoot to perform the sealing of the TPM during the first boot and/or during updates to the image.  For now, this allows us to provide a tool that can demonstrate how sealing can work using knowledge at compile time.  Once we are done with this initial solution, we'll start to work additional designs that properly integrate measured boot and add additional tools/methods to support different provisioning and update requirements.  There's probably only a few "valid" solutions that we should implement/support.  For now, this is just proof of concept to show what is possible without implying or forcing any other design.  Specifically, the image encryption key should NOT be bound in the current way, which is why it was not demonstrated.

danielinux:
OK thanks for clarifying @billphipps . Please keep in mind the scenario where encryption is off, and provide instructions for both cases.

Going forward, I think it would be OK to assume immutable wolfBoot (no self-update allowed with measured boot), and it would be possible perhaps to use(/add) wolfBoot's digest to the PCR food, which would also help identifying tampering attempts to the bootloader. At leasr for the most generic use cases.

tools/preseal/user_settings.h
billphipps: Funny, the patch I originally supplied used LINUX_DEV as the default.  A later patch put in the Linux SPI HAL instead.  Currently, this has to be done in the Makefile rather than user settings because wolfTPM doesn't include user settings by default.  We'll make sure both of these configs are identified in the Makefile.  We probably need a smarter configuration selection in the future.

PR #315, opened 2023-06-14
fix sim-update-powerfail-resume and header mismatch

jpbland1:
sim-update-powerfail-resume was actually not testing powerfailure at all because the addresses the sim uses changed. fixed that and then fixed the error that the test catches, which is if a power failure happens when the header is being written the header will be corrupt and needs to be re-loaded once the write is resumed and finished

zd 15415


src/update_flash.c
```diff
@@ -492,6 +493,19 @@ static int RAMFUNCTION wolfBoot_update(int fallback_allowed)
                 wolfBoot_set_update_sector_flag(sector, flag);
         }
         sector++;
+
+        /* headers that can be in different positions depending on when the
+         * power fails are now in a known state, re-read and swap fw_size
+         * because the locations are correct but the metadata is now swapped */
+        if (sector == 1) {
+            wolfBoot_open_image(&boot, PART_BOOT);
```
rizlik, line 501: I think this is an incomplete fix:
If a power failure happens when writing the header to the boot partition, the status will be:
- the update header is overwritten with the backup of the boot header
- the swap sector contains the update header
- the boot header is erased/corrupted
In this case, the `total_size` will be calculated using only the v1 image size (header in the update partition). The image v2 size is "hidden" in the swap sector and it isn't checked.
To trigger the use case in a test:
1. modify the size of the v2 image:
```
--- a/tools/test.mk
+++ b/tools/test.mk
@@ -173,7 +173,7 @@ test-sim-internal-flash-with-update: wolfboot.bin test-app/image.elf FORCE
        $(Q)dd if=/dev/urandom of=test-app/image.elf bs=1K count=16 oflag=append conv=notrunc
        $(Q)$(SIGN_TOOL) $(SIGN_OPTIONS) test-app/image.elf $(PRIVATE_KEY) 1
        $(Q)cp test-app/image.bak.elf test-app/image.elf
-       $(Q)dd if=/dev/urandom of=test-app/image.elf bs=1K count=16 oflag=append conv=notrunc
+       $(Q)dd if=/dev/urandom of=test-app/image.elf bs=1K count=32 oflag=append conv=notrunc
        $(Q)$(SIGN_TOOL) $(SIGN_OPTIONS) test-app/image.elf $(PRIVATE_KEY) $(TEST_UPDATE_VERSION)
        $(Q)dd if=/dev/zero bs=$$(($(WOLFBOOT_SECTOR_SIZE))) count=1 2>/dev/null | tr "\000" "\377" > erased_sec.dd
        $(Q)$(SIGN_TOOL) $(SIGN_ARGS) $(DELTA_UPDATE_OPTIONS) \
```
2. test with a *single* failure update when writing to the boot header:
```
# prints 1
./wolfboot.elf update_trigger get_version 
./wolfboot.elf powerfail c0020000
# prints 1 (should print 2)
./wolfboot.elf get_version 
```

Please add this edge case to the tests when fixed

rizlik (changes requested):
good catch.
Is delta update affected?

src/update_flash.c
jpbland1: good point, didn't think about total_size

rizlik (approved):
do you think moving the

jpbland1:
fixed the size problem, should be good to merge

rizlik (commented):
good to me, as a final note, can you consider adding a regression test for the edge case of total size? the actual tests are faking three power failures, which masks the bug.

jpbland1:
I found a bug where the encryption key can be lost if the powerfail happens after the encrypt key is erased since it was sitting in memory until the encrypt key is set again. now I save it to swap and recover it if necessary

src/update_flash.c
```diff
@@ -322,27 +366,24 @@ static int wolfBoot_delta_update(struct wolfBoot_image *boot,
...
-#ifdef EXT_FLASH
-    ext_flash_lock();
-#endif
-    hal_flash_lock();
 
-/* Save the encryption key after swapping */
-#ifdef EXT_ENCRYPTED
-    wolfBoot_set_encrypt_key(key, nonce);
-#endif
+    /* copy the last sector of boot  which has the encrypt key and partition
```
danielinux, line 377: Doesn't this leak the key in clear to the external device (assuming EXT_SWAP=1)? Would this not defeat the purpose of the key itself?

rizlik (changes requested):
Good catch for the new bug, I agree with the commend of @danielinux , external encryption key shouldn't go to swap sector

src/update_flash.c
jpbland1: I wasn't aware the key couldn't be put into a potentially external swap, I thought the point of encrypted partitions was to protect the new partition in transit but I guess this protects it at rest, though I don't see why someone with physical access to the device wouldn't just read it from boot after it was decrypted.

Anyways I'm not going to argue with the referee, I've changed it to the scheme we discussed where the first sector of boot is swapped out and used to store a backup of the key when the final swap takes place. This is only necessary when encrypted partitions are used but using this final swap scheme also protects against the case where the power fails right after erasing the last boot sector. In that scenario the testing flag was never being set and the testing phase never happened.

src/libwolfboot.c
```diff
@@ -974,20 +974,50 @@ int RAMFUNCTION wolfBoot_set_encrypt_key(const uint8_t *key,
     return 0;
 }
 
+int RAMFUNCTION wolfBoot_backup_encrypt_key(const uint8_t* key,
+    const uint8_t* nonce)
+{
+#ifndef MMU
+    uint32_t magic[2] = {WOLFBOOT_MAGIC, WOLFBOOT_MAGIC_TRAIL};
+
+    XMEMCPY((uint8_t*)WOLFBOOT_PARTITION_BOOT_ADDRESS, key, ENCRYPT_KEY_SIZE);
```
danielinux, line 983: not sure how this would work, XMEMCPY straight to NVM? Did you mean to use `hal_flash_write`?

danielinux (changes requested):
Please double check how data is written to internal flash. You should use `hal_flash_unlock/erase/write`.

src/libwolfboot.c
jpbland1: whoops, I've got simulator brain. thought you could read and write that way

jpbland1:
Kind of messy but I found problems with `NVM_WRITE_ONCE` that arise when it's used with external and encrypted partitions. It was trying to call `nvm_select_fresh_sector` which I'm positive will panic if the external partition address doesn't exist in memory space. There were other smaller issues as well including the `addr_align` calculation being incorrect when the header has the encryption key and nonce which lead to it erasing the selected partition instead of the non-selected partition and `partition_magic_write` reading from `base` instead of `addr_read`.

hal/sim.c
```diff
@@ -36,6 +36,8 @@ static uint8_t *ram_base;
 static uint8_t *flash_base;
 
 uint32_t erasefail_address = 0xFFFFFFFF;
+int eraseProgressive = 0;
```
danielinux, line 39: These could possibly be `static` if they are only used by sim.c

include/wolfboot/wolfboot.h
```diff
@@ -219,13 +219,15 @@ extern "C" {
 #ifndef WOLFBOOT_FLAGS_INVERT
 #define IMG_STATE_NEW       0xFF
 #define IMG_STATE_UPDATING  0x70
+#define IMG_STATE_FINAL_SWAP  0xA0
```
danielinux, line 222: if FINAL_SWAP is after UPDATING it should have a value where zeros are added to 0x70 for normal inverted flash allowing multiple writes. This value peraps should be 0x30

danielinux (changes requested):

danielinux:
Please rebase to master+squash.
Please double check the visibility of the symbols (functions/globals being only used in the file where they are declared should be static)

include/wolfboot/wolfboot.h
jpbland1: I see what you mean, NEW 0b1111 -> UPDATING 0b0111 -> FINAL_SWAP 0b0011

PR #329, opened 2023-07-26
update sign to decode keys instead of assuming they're raw

jpbland1:
add a github workflow for testing external partition signing
ZD 16491


tools/keytools/sign.c
```diff
@@ -92,6 +92,8 @@
     #include <wolfssl/wolfcrypt/logging.h>
 #endif
 
+#include "wolfboot/wolfboot.h"
```
dgarske, line 95: Ideally we will want to update the sign.py version as well. If not we should document in Signing.md

dgarske (changes requested):

tools/keytools/sign.c
dgarske: @danielinux please let us know your thoughts on the topic.
If the wolfcrypt-py has the ECC public key decode (I believe it does) I think we should add it.

dgarske (approved):

danielinux (approved):

PR #334, opened 2023-08-03
Simulator fixes and support for using MacOS

dgarske:
* Added simulator support for Mac.
* Fix for simulator to properly assemble wolfboot.bin + signedtestapp + update + swap.
* Fixes for handling 64-bit assigned mmap virtual addresses. Added hal_flash_write and hal_flash_erase support for 64-bit using `ARCH_64BIT` and added haladdr_t for address type.
* Fix simulator conflict with src/libwolfboot.o object in test-app.
* Cleanup test-app linker flags.
* Speedup the key tool builds.


include/hal.h
```diff
@@ -46,8 +46,8 @@ void hal_deinit();
 #endif
 
 void hal_init(void);
-int hal_flash_write(uint32_t address, const uint8_t *data, int len);
-int hal_flash_erase(uint32_t address, int len);
+int hal_flash_write(uintptr_t address, const uint8_t *data, int len);
```
dgarske, line 49: @danielinux , this change of mine seems like it could break some platforms. Perhaps I could just add a HAL build option to support 64-bit addresses?

danielinux (changes requested):
Simulator test gives segmentation fault on linux.

Reproduce with:
`make test-sim-internal-flash-with-update`

then 
`tools/scripts/sim-sunnyday-update.sh`

I suspect two different strategies need to be used on Mac vs GNU due to the differences in mmap _"hint pointer"_ handling.

dgarske:
> Simulator test gives segmentation fault on linux.
> 
> Reproduce with: `make test-sim-internal-flash-with-update`
> 
> then `tools/scripts/sim-sunnyday-update.sh`
> 
> I suspect two different strategies need to be used on Mac vs GNU due to the differences in mmap _"hint pointer"_ handling.

@danielinux  I could not reproduce. What was your .config? I've only seen a set fault if mmap returns a 64-bit pointer and parts of the code assume 32-bit when the address gets passed around.

danielinux:
My bad, I was testing side-by-side, with the same config as `master`. I did not realize the address changed in the example config

danielinux (approved):

PR #335, opened 2023-08-03
wolfBoot TPM improvements

dgarske:
* Added TPM SPI wait state support and debug logging.
* Added platform auth ownership (change platform password to random value before boot).
* Added parameter encryption support.
* Removed the TPM hashing feature (not practical).
* Fixed RSA with wolfTPM build.
* Fixed cleanup wolfTPM objects on make clean.
* TPM based root of trust for public key. A hash of the public key is stored in NV with auth and lock in platform.


danielinux (changes requested):
Nice improvements to wolfTPM support!

Please adjust binary size at the bottom of tools/test.mk to pass size watermark tests

danielinux (approved):

PR #338, opened 2023-08-09
Fixes and improvements for iMXRT1050 QSPI

dgarske:
* Fixes for i.MX RT1050 to support QSPI with IS25WP064A (enabled with `CONFIG_FLASH_IS25WP064A`) or W25Q64JV (enabled with `CONFIG_FLASH_W25Q64JV`). The default is the HyperFlash.
* Add support for iMX RT1050 and RT1060 wolfBoot_printf using DEBUG_UART=1.
* Fix RT1050 example app LED.
* Fix documentation for test-app location. 
* Added GitHub action build tests for NXP parts (iMX RT, Kinetis and LPC).

ZD 16472


docs/Targets.md
```diff
@@ -1053,6 +1053,31 @@ DCP support (hardware acceleration for SHA256 operations) can be enabled by usin
 
 Firmware can be directly uploaded to the target by copying `factory.bin` to the virtual USB drive associated to the device, or by loading the image directly into flash using a JTAG/SWD debugger.
```
bigbrett, line 1055: We should probably add some documentation here about which of the new flash configuration macros to set  (`USE_GET_CONFIG`, `CONFIG_FLASH_IS25WP064A`, `CONFIG_FLASH_W25Q64JV`, `CONFIG_FLASH_QE_ENABLE`), as well as clarify whether or not we support HyperFlash.

hal/imx_rt.c
```diff
@@ -244,44 +274,134 @@ const flexspi_nor_config_t __attribute__((section(".flash_config"))) qspiflash_c
...
-
+/** Flash configuration in the .flash_config section of flash **/
 #ifdef CPU_MIMXRT1052DVJ6B
-    #define CONFIG_FLASH_SIZE (64 * 1024 * 1024)  /* 64MBytes  */
-    #define CONFIG_FLASH_PAGE_SIZE 512UL          /* 512Bytes  */
-    #define CONFIG_FLASH_SECTOR_SIZE (256 * 1024) /* 256KBytes */
-    #define CONFIG_FLASH_BLOCK_SIZE (256 * 1024)  /* 256KBytes */
-    #define CONFIG_FLASH_UNIFORM_BLOCKSIZE true
-    #define CONFIG_SERIAL_CLK_FREQ kFlexSpiSerialClk_30MHz
-const flexspi_nor_config_t __attribute__((section(".flash_config"))) qspiflash_config = {
```
dgarske, line 255: @bigbrett do you think I should keep a version available for HyperFlash also? I think it would be fairly easy to support it still and test.

bigbrett: Yes it is probably a smart idea, since it is the default flash device on the EVKs, and it could see increased adoption over time. Looks like the demo apps  contain the LUT and flash config to make it work too.

danielinux (approved):

PR #340, opened 2023-08-15
Internal only write once

jpbland1:
only call nvm_select_fresh_sector when an internal partition is in use, otherwise it will try to read from a corrupt memory address

ZD 15415


dgarske (approved):
Thanks, this resolved my customer issue.

PR #341, opened 2023-08-15
fix powerfail test to actually trigger, previously

jpbland1:
the test has been doing nothing but trigger the update and then fallback over and over again
ZD 16472


dgarske (approved):
Good find! Confirmed

PR #346, opened 2023-08-17
Document API functions

tmael:
This PR adds function comments to prep for DO178

### DIFF
--- a/src/update_disk.c
+++ b/src/update_disk.c
@@ -64,7 +64,7 @@
 
 #define IMAGE_PRELOAD_ADDRESS 0x5000100
 /**
- * @brief RAM function for starting the boot process.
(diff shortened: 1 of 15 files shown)


src/update_disk.c
```diff
@@ -50,7 +63,13 @@
 #define MAX_FAILURES 4
 
 #define IMAGE_PRELOAD_ADDRESS 0x5000100
-
+/**
+ * @brief RAM function for starting the boot process.
```
danielinux, line 67: It is irrelevant here if the function can be executed from RAM when `RAM_CODE=1`. Please remove "RAM" from "RAM function [...]"

danielinux (changes requested):

danielinux (approved):

PR #351, opened 2023-08-21
handle case where PART_UPDATE_ENDFLAGS % WOLFBOOT_SECTOR_SIZE == 0

jpbland1:
ZD 16472

### DIFF
--- a/src/libwolfboot.c
+++ b/src/libwolfboot.c
@@ -590,10 +590,12 @@ void RAMFUNCTION wolfBoot_update_trigger(void)
     uintptr_t lastSector = PART_UPDATE_ENDFLAGS -
         (PART_UPDATE_ENDFLAGS % WOLFBOOT_SECTOR_SIZE);
 
+#ifndef FLAGS_HOME
     /* if PART_UPDATE_ENDFLAGS stradles a sector, (all non FLAGS_HOME builds)
      * align it to the correct sector */
     if (PART_UPDATE_ENDFLAGS % WOLFBOOT_SECTOR_SIZE == 0)
         lastSector -= WOLFBOOT_SECTOR_SIZE;
+#endif
 #endif
 #ifdef NVM_FLASH_WRITEONCE
     uint8_t selSec = 0;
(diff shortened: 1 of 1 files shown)


src/libwolfboot.c
```diff
@@ -589,6 +589,11 @@ void RAMFUNCTION wolfBoot_update_trigger(void)
 #if defined(NVM_FLASH_WRITEONCE) || defined(WOLFBOOT_FLAGS_INVERT)
     uintptr_t lastSector = PART_UPDATE_ENDFLAGS -
         (PART_UPDATE_ENDFLAGS % WOLFBOOT_SECTOR_SIZE);
+
+    /* if PART_UPDATE_ENDFLAGS stradles a sector, (all non FLAGS_HOME builds)
+     * align it to the correct sector */
+    if (PART_UPDATE_ENDFLAGS % WOLFBOOT_SECTOR_SIZE == 0)
```
dgarske, line 595:
```WOLFBOOT_PARTITION_SWAP_ADDRESS?=0x60050000
590 -> lastSector = 0x60050000
596 -> lastSector = 0x6004f000```

Looks good. Why did we not see this issue previously? Are we missing some test cases?

File: src/libwolfboot.c [hunk 2]
```diff
@@ -589,6 +589,11 @@ void RAMFUNCTION wolfBoot_update_trigger(void)
 #if defined(NVM_FLASH_WRITEONCE) || defined(WOLFBOOT_FLAGS_INVERT)
     uintptr_t lastSector = PART_UPDATE_ENDFLAGS -
         (PART_UPDATE_ENDFLAGS % WOLFBOOT_SECTOR_SIZE);
+
+    /* if PART_UPDATE_ENDFLAGS stradles a sector, (all non FLAGS_HOME builds)
```
dgarske, line 593: Could this be gated on `#ifndef FLAGS_HOME`?

dgarske (changes requested):

src/libwolfboot.c
jpbland1: I think it's because `nvm_select_fresh_sector` defaults to selecting 0 when both sectors are empty and without this PR it was reading the swap sector and then writing to sector 0 so sector 0 had a non-erased byte and was correctly selected the next time around

dgarske (approved):

PR #352, opened 2023-08-21
add hal_init to the imx test-app and trigger update

jpbland1:
ZD 16472

### DIFF
--- a/test-app/app_imx_rt.c
+++ b/test-app/app_imx_rt.c
@@ -124,6 +124,12 @@ void rt1050_init_pins(void)
 
 void main(void)
 {
+    hal_init();
+    if (wolfBoot_current_firmware_version() == 1) {
+        wolfBoot_update_trigger();
+       return;
+    }
+
     imx_rt_init_boot_clock();
 #if defined(CPU_MIMXRT1062DVL6A) || defined(CPU_MIMXRT1064DVL6A)
     rt1060_init_pins();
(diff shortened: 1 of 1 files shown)


test-app/app_imx_rt.c
```diff
@@ -124,6 +124,12 @@ void rt1050_init_pins(void)
 
 void main(void)
 {
+    hal_init();
+    if (wolfBoot_current_firmware_version() == 1) {
```
dgarske, line 128: Can you please move this down after:

```
PRINTF("wolfBoot Test app, version = %d\r\n",
         wolfBoot_current_firmware_version());
```

dgarske (changes requested):

test-app/app_imx_rt.c
```diff
@@ -124,6 +124,12 @@ void rt1050_init_pins(void)
 
 void main(void)
 {
+    hal_init();
```
dgarske, line 127:
```make[1]: Entering directory '/home/davidgarske/GitHub/wolfboot/test-app'
	[CC-ARM] app_imx_rt.o
app_imx_rt.c: In function 'main':
app_imx_rt.c:127:5: warning: implicit declaration of function 'hal_init' [-Wimplicit-function-declaration]
  127 |     hal_init();
      |     ^~~~~~~~```

dgarske (changes requested):

test-app/app_imx_rt.c
jpbland1: it actually wasn't calling `hal_init` that was required, I needed to move the `wolfBoot_update_trigger` to before the SysTick_Config call, it seems to mess with `hal_flash_write`

jpbland1:
it seems SysTick isn't required and the print statements still work without it, I've removed it and cleaned up the test-app

dgarske (approved):

PR #354, opened 2023-08-23
Add powerfail during final sector erase

jpbland1:
Add powerfail during final sector erase, exposing the problems of losing the encryption key when the power fails before the key can be written back, failing to select the correct NVM_FLASH_WRITEONCE sector when a sector is corrupt, and losing the testing state when the power fails before the partition state can be set. these issues are addressed in the next commit
ZD 15415

### DIFF
--- a/.github/workflows/test-powerfail-simulator.yml
+++ b/.github/workflows/test-powerfail-simulator.yml
@@ -181,7 +181,7 @@ jobs:
 
       - name: Build wolfboot.elf
         run: |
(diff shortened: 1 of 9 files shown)


danielinux (dismissed):
LGTM. Off to @rizlik for double checking the logic.

rizlik:
investigate a way to avoid erasing the key until wolfBoot_success

danielinux (approved):

PR #350, opened 2023-08-21
LMS wolfBoot support.

philljj:
# Description

Adds support for LMS/HSS signatures to wolfBoot.

LMS/HSS is a post-quantum stateful hash-based signature scheme.

# Documentation

Added new docs post-quantum readme:
```
docs/PQ.md
```

# Building

See procedure in `docs/PQ.md`.

# Config

Added new LMS sim example
```
config/examples/sim-lms.config
```
The `LMS_LEVELS`, `LMS_HEIGHT`, and `LMS_WINTERNITZ`, `IMAGE_SIGNATURE_SIZE`, and (optionally) `IMAGE_HEADER_SIZE` must be set here.

```
SIGN?=LMS
...
LMS_LEVELS=2
LMS_HEIGHT=5
LMS_WINTERNITZ=8
...
IMAGE_SIGNATURE_SIZE=2644
IMAGE_HEADER_SIZE?=5288
```

In LMS the signature size is a function of the parameters. Use the added helper script `tools/lms/lms_siglen` to calculate your signature length given your LMS parameters:
```
$./tools/lms/lms_siglen 
levels:      3
height:      5
winternitz:  8
#
total_len:   3992
```

# More Info

See these links for more info on LMS and wolfSSL/wolfCrypt:
- https://github.com/wolfSSL/wolfssl-examples/tree/master/pq/stateful_hash_sig
- https://www.wolfssl.com/documentation/manuals/wolfssl/appendix07.html#post-quantum-stateful-hash-based-signatures

PR #360, opened 2023-09-04
[Test action, please ignore] Lms action fix hss dep

danielinux:
Creating draft PR to test github action fixes from PR #350


tools/scripts/renode-test-update.sh
```diff
@@ -19,6 +19,16 @@ quit_renode() {
 
 
 rm -f $RENODE_UART
+
+if (echo $TEST_OPTIONS | grep "LMS" &>/dev/null); then
+    apt install -y git
+    mkdir -p lib/hss-lib/lib
```
philljj, line 25: I think this should be
```
mkdir -p lib/hash-sigs/lib
```

PR #365, opened 2023-09-14
x86: fsp: misc improvements

rizlik:
- support .data and .bss in stage1
- moved SATA memory areas in wolfboot .bss
- improved flash layout
- load app image right after wolfboot


rizlik:
tested

dgarske (approved):

PR #373, opened 2023-09-28
Improvement to keystore and key management tools

danielinux:
- sign: added an option to disable timestamp.
- keygen: added options to host multiple keys with different algorithms and sizes
- keygen: added "--id {list}" option to limit permissions for specific keys


.github/workflows/test-keytools.yml
```diff
@@ -144,3 +144,9 @@ jobs:
       - name: Generate final signed binary
         run: |
           ./tools/keytools/sign --rsa2048 --sha256 --manual-sign test-app/image.elf public-key.der 1 test-app/image_v1.sig
+
+
+    # TODO: Test WOLFBOOT_UNIVERSAL_KEYSTORE
```
dgarske, line 149: Do you want to add these tests now or future PR?

dgarske (dismissed):

.github/workflows/test-keytools.yml
danielinux: good idea, I'll add them now

dgarske (approved):

PR #381, opened 2023-10-13
Fixed keygen ecc key initialization

danielinux:
And use SP_MATH for keytools.


dgarske (approved):

tmael (approved):
@danielinux, Looks good.

PR #383, opened 2023-10-15
Use bash from env for better portability

Luz:
Hi
According to [Stackoverflow](https://stackoverflow.com/questions/10376206/what-is-the-preferred-bash-shebang#10383546) it looks like the portability of the scripts could be improved by having this new style for the bash shebang.

### DIFF
--- a/IDE/Renesas/e2studio/RA6M4/elf2hex.sh
+++ b/IDE/Renesas/e2studio/RA6M4/elf2hex.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 #
 # convert elf to hex file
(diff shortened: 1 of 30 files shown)


danielinux:
Hi @Luz,
May I ask if there is any specific reason for this? The answers on the thread you linked don't seem to agree on a definitive portable method, and the current shebangs seem to work on all systems we tested. Are you trying to run these script in a specific OS? If so, what issue did you encounter?

Thanks,
- - 
Daniele - wolfSSL

Luz:
Well scripts are not executable for me, as I have no /bin/bash :( But I can also easily fix it on my system (nixos), would be no problem :) Feel free to close, if you prefer the current style :)

danielinux:
Thanks for the explanation. We have decided to keep our current `#!` for now, to avoid breaking compatibility for existing users, but we may reconsider this in the future.

PR #398, opened 2023-12-27
Submodules update

danielinux:
- Updated wolfSSL to 5.6.6: Adjusted footprint test to new binary sizes 
- Updated wolfTPM to latest: Added missing include to tpm policy_sign

### DIFF
--- a/tools/tpm/policy_sign.c
+++ b/tools/tpm/policy_sign.c
@@ -29,7 +29,6 @@
 #include <wolfssl/wolfcrypt/ecc.h>
 #include <wolfssl/wolfcrypt/hash.h>
 #include <wolftpm/tpm2_wrap.h>
-#include <wolfssl/error-ssl.h>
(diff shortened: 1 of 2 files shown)


tools/tpm/policy_sign.c
```diff
@@ -29,6 +29,7 @@
 #include <wolfssl/wolfcrypt/ecc.h>
 #include <wolfssl/wolfcrypt/hash.h>
 #include <wolftpm/tpm2_wrap.h>
+#include <wolfssl/error-ssl.h>
```
dgarske, line 32: We don't want to include any TLS/SSL headers.
Instead of adding this header please replace `INPUT_SIZE_E` with `BUFFER_E`. 
See http://github.com/wolfSSL/wolftpm/pull/308

dgarske (changes requested):

dgarske (approved):

PR #400, opened 2024-01-23
API's to support OTP flash read/write on the STM32H7

dgarske:
API's to support OTP flash read/write on the STM32H7. Enabled with `FLASH_OTP_ROT`.

ZD 17063


danielinux:
Superseeded by #449 . Closing.

PR #404, opened 2024-02-12
Fix for NXP T1024 to get MAC address from Integrity OS ethernet config

dgarske:
Fix for NXP T1024 to get MAC address from Integrity OS ethernet configuration area in flash.


danielinux (approved):

dgarske:
See ZD 17479

PR #409, opened 2024-02-22
DISABLE_BACKUP is now powerfail-safe.

danielinux:
By not checking/updating sector flags during copies from UPDATE to BOOT partitions, a copy operation getting interrupted by power failure will be less likely to cause problems to the integrity of the BOOT image. This is because in case of interruptions, the copy operation will be restarted from the first sector.

Idea taken from ZD17551.

### DIFF
--- a/src/update_flash.c
+++ b/src/update_flash.c
@@ -553,23 +553,14 @@ static int RAMFUNCTION wolfBoot_update(int fallback_allowed)
 #endif
     hal_flash_lock();
 #else /* DISABLE_BACKUP */
-#warning "Backup mechanism disabled! Update installation will not be interruptible"
-    /* Read encryption key/IV before starting the update */
 #ifdef EXT_ENCRYPTED
     wolfBoot_get_encrypt_key(key, nonce);
 #endif
 
     /* Directly copy the content of the UPDATE partition into the BOOT partition.
-     * This mechanism is not fail-safe, and will brick your device if interrupted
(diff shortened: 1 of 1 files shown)


tools/scripts/sim-update-powerfail-resume-nobackup.sh
```diff
@@ -0,0 +1,30 @@
...
+    exit 1
+fi
+
+./wolfboot.elf powerfail 0 get_version 2>/dev/null
+./wolfboot.elf powerfail 15000 get_version 2>/dev/null
+./wolfboot.elf powerfail 18000 get_version 2>/dev/null
+./wolfboot.elf powerfail 1a000 get_version 2>/dev/null
+# fail on the last sector to stop the encrypt key save and state update
+./wolfboot.elf powerfail 3e000 get_version 2>/dev/null
+# may not trigger on non NVM_FLASH_WRITEONCE
```
jpbland1, line 14: does `NVM_FLASH_WRITEONCE` get used with `DISABLE_BACKUP`? Also missing combo cases for things like `HOME_FLAGS`, not sure if it's relevant for this flag but otherwise looks good.

danielinux:
Thanks @jpbland1 - I've added two more test cases to check the combination of DISABLE_BACKUP with the two options mentioned above. Please consider for merging.

jpbland1 (approved):
Reviewed and tested, looks good

PR #412, opened 2024-02-28
Added support for custom TLVs in manifest header

danielinux:
Added support in `sign.c` for custom TLVs in the manifest header. Numeric values are supported (valid len: 1, 2, 4 or 8). 
[edit: custom length buffers are also supported via `--custom-tlv-buffer`]

The numbers are stored in little-endian format (according to the current format of the other fields). The TLVs are part of the signed image, so their value is secured against modifications.

The custom fields stored in the TLV can be retrieved at runtime in wolfboot by using `wolfBoot_find_header()`.
Documentation is updated accordingly.


dgarske (approved):
Thank you so much for adding this feature! Excellent work.

PR #414, opened 2024-03-01
Added support for SIGN=ECC521

danielinux:


dgarske:
This is great! Love how little changes were required, since most of the 521 support was already added.

danielinux:
I guess we never "activated" ECC521 because before #395 tests were 100% failing

dgarske (dismissed):

dgarske (approved):

PR #427, opened 2024-04-05
Replace CMAKE_*_DIR to PROJECT_*_DIR

opkaizen:
This allows to reuse the project-level CMakeLists into another top-level CMakeLists for integration in other projects.


lealem47:
Hi @opkaizen,

Why not just compile wolfBoot once then link with your application? Instead of compiling it along with your project?

kaizensparc:
I wanted to avoid having multiple stages before I can get the factory image, and to have only one build system where I can place all my configuration at once (linker script addresses, etc.)

kaizensparc:
I wanted to avoid having multiple stages before I can get the factory image, and to have only one build system where I can place all my configuration at once (linker script addresses, etc.)

opkaizen:
I wanted to avoid having multiple stages before I can get the factory image, and to have only one build system where I can place all my configuration at once (linker script addresses, etc.)

lealem47:
Hi @opkaizen ,
Thank you for the patch. I agree it would be helpful. However, this diff is very small. We here at wolfSSL would tend to treat small fixes as a bug report as it is not worth going through the process of signing the the contributor agreement to become a contributor.

That said, are you planning to make a larger contribution in the near future? If so, then it might be worth it to sign a contributor agreement. Please let me know.

opkaizen:
I am in the process of selecting a suitable bootloader for a product, so I won't know if there are other contributions until at least 1-2 months. I think the easiest is to avoid signing a contributor agreement for now, and if I end up using WolfBoot, then signing it :)
The agreement is currently being reviewed but it will take some time :/
Should I close this PR and use only issues for now?

lealem47:
Understood. I'll close this PR and treat it like a bug report. I will put up a patch addressing this shortly. Thanks!

PR #428, opened 2024-04-08
Check if power source is already configured on STM32H7

opkaizen:
If the power source is already configured, writing on the register can cause a deadlock if the configuration does not match, which can be tricky to fix. This is the check that is done in the STM32 HAL library to avoid this.

I got this issue because the STM32H757 demo board and the STM32H745XI discovery board use the SMPS, and even though the power on reset configuration is LDO|SMPS, writing on LDO deadlocked my boards.


danielinux:
Hi @opkaizen, thanks for spotting this. Could you please get in touch with support@wolfssl.com to set up a contributor agreement?

danielinux:
hi @opkaizen did you get in touch with us yet regarding the CA?

opkaizen:
Hi, not yet, I will soon :)

danielinux:
The CA process did not go through. See zd17861.

PR #430, opened 2024-04-10
Anti-glitch armor: Added clobbers to assembly

danielinux:
Possible fix for ZD17804.

User reports that  `r7` is being reused for an intermediate assignment translating a `mov` from memory. Proper clobbering should prevent that from happening.


rizlik (commented):
It might be pedantic but I think it's also a good idea to add "cc" to the clobbers list for instructions that modify processor flags (and, cmp, lsls).

rizlik (approved):

PR #431, opened 2024-04-10
Updated submodules: wolfSSL-v5.7.0 + wolfTPM latest

danielinux:


src/xmalloc.c
```diff
@@ -315,6 +319,7 @@ static uint32_t sha_block[HASH_BLOCK_SIZE];
     #if defined(WOLFBOOT_HASH_SHA256) || defined(WOLFBOOT_HASH_SHA384)
         { (uint8_t *)sha_block, HASH_BLOCK_SIZE * sizeof(uint32_t), 0 },
     #endif
+        { asncheck_buf, ASNCHECK_BUF_SIZE, 0 },
```
danielinux, line 322: It appears that this extra buffer is now needed since wolfSSL v5.7.0 due to `CALLOC_ASNGETDATA` in `wc_RsaPublicKeyDecode_ex()`

config/examples/stm32c0.config
```diff
@@ -1,8 +1,9 @@
 ARCH?=ARM
 TARGET?=stm32c0
-#Supports ED25519/SHA384 or RSA2048/SHA256
-SIGN?=RSA2048
-HASH?=SHA256
+#Supports ED25519/SHA384
+# No longer supported: RSA2048/SHA256
```
dgarske, line 4: What happened here? Is this due to the code size you mentioned? Shall we just adjust the default flash layout a bit?

tools/test.mk
```diff
@@ -972,25 +972,25 @@ test-size-all:
 	make keysclean
 	make test-size SIGN=ED25519 LIMIT=11380
 	make keysclean
-	make test-size SIGN=ECC256  LIMIT=17696
+	make test-size SIGN=ECC256  LIMIT=17824
```
dgarske, line 975: Is it possible to add a GitHub action that also does an update to latest master in the submodules to detect these issues earlier?

config/examples/stm32c0.config
danielinux: Yes the RSA version now overflows the 10KB partition. Ed25519 still fits. We do have a LMS configuration file also, but it carries a 2KB header for the application, so it does not improve much the application size. I'm adding multiple configuration files for this target with different flash geometries

tools/test.mk
danielinux: Let's discuss about this, this should perhaps be in the wolfSSL CI instead, so we can spot it while reviewing PRs?

PR #445, opened 2024-04-26
Fix compilation issues when using PKA=1 for i.MX RT targets

MulattoKid:
I encountered issues when compiling with `PKA?=1` in my `.config` file. With these changes it compiles and seems to work (stepping into the code I end up in the `fsl_dcp` driver and the speed increases).


dgarske:
Contributor agreement on file. Changes look good to me. 
Do you know if PKA supported on all the i.MX RT targets? Maybe we should add another CI test for these? 
[diff.patch](https://github.com/wolfSSL/wolfBoot/files/15131488/diff.patch)

bigbrett:
@dgarske re: your question, All the i.MX RT 10xx derivatives that wolfBoot supports use the DCP hardware which provides AES and SHA acceleration. Other (newer) parts in the RT family like the cortex M33-based parts (RT 500, etc) and the RT 11xx also have HW accel but use different IP so would require different MCUXpresso drivers. But as far as the RT platforms we currently support in wolfBoot go, PKA via the DCP module is present on all of them.

bigbrett (dismissed):

MulattoKid:
Checking the ref. manuals for i.MX [RT1010, RT1064], and all have the DCP (data co-processor).

MulattoKid:
@dgarske I added the patch as suggested, and gave the new builds a `_pka` suffix.

MulattoKid:
Not sure if you're OK with having `-DDCP_USE_DCACHE=0` in the config examples for the RTs, or if you'd like to make separate configs for the `PKA=1` builds. Just let me know :)

dgarske:
> Not sure if you're OK with having `-DDCP_USE_DCACHE=0` in the config examples for the RTs, or if you'd like to make separate configs for the `PKA=1` builds. Just let me know :)

Hi @MulattoKid , that is fine with me, but can you expand on the reason the data cache needed to be disabled? Either a comment here in the PR on in the .config.

MulattoKid:
> > Not sure if you're OK with having `-DDCP_USE_DCACHE=0` in the config examples for the RTs, or if you'd like to make separate configs for the `PKA=1` builds. Just let me know :)
> 
> Hi @MulattoKid , that is fine with me, but can you expand on the reason the data cache needed to be disabled? Either a comment here in the PR on in the .config.

When compiling without the DCACHE disabled, compilation fails in the `dcp_port` with an explicit compile time error. I haven't looked into what the implications of this are, but as the port explicitly doesn't support DCACHE I assumed disabling it would be fine. Let me know if I've misunderstood something 👍

dgarske (approved):

PR #446, opened 2024-04-26
Fix for NXP i.MX RT flash configuration to not force .flash_config

dgarske:
Fix for NXP i.MX RT flash configuration to not force .flash_config when building wolfBoot as library. ZD 17850.

### DIFF
--- a/hal/imx_rt.c
+++ b/hal/imx_rt.c
@@ -218,6 +218,16 @@ bootloader_api_entry_t *g_bootloaderTree;
 #define LUT_SEQ_IDX_14                    (0x38U)   /* RESERVED */
 #define LUT_SEQ_IDX_15                    (0x3CU)   /* Dummy */
 
+#ifdef TARGET_library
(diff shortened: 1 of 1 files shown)


hal/imx_rt.c
```diff
@@ -218,6 +218,16 @@ bootloader_api_entry_t *g_bootloaderTree;
 #define LUT_SEQ_IDX_14                    (0x38U)   /* RESERVED */
 #define LUT_SEQ_IDX_15                    (0x3CU)   /* Dummy */
 
+#ifdef TARGET_library
```
bigbrett, line 221: We should probably add a quick blurb about this new macro to the docs

bigbrett (approved):

hal/imx_rt.c
dgarske: Sure can do. It's not a new macro, really. When you use the config/examples/library.config the TARGET=library, which does a CFLAG for `-DTARGET_library`.

danielinux (approved):

PR #460, opened 2024-06-26
[QSPI] Fixed registers and pin config on STM32U5

danielinux:
ZD 18119


dgarske (approved):
Taking these improvements as-is. Thanks @danielinux

PR #463, opened 2024-06-28
add emergency fallback test, currently fails due to

jpbland1:
the sector flags needing to be reset by update trigger

### DIFF
--- a/hal/sim.c
+++ b/hal/sim.c
@@ -45,6 +45,7 @@
 uint8_t *sim_ram_base;
 static uint8_t *flash_base;
 
+int forceEmergency = 0;
 uint32_t erasefail_address = 0xFFFFFFFF;
 
 #define INTERNAL_FLASH_FILE "./internal_flash.dd"
@@ -105,8 +106,29 @@ void hal_prepare_boot(void)
 
 int hal_flash_write(uintptr_t address, const uint8_t *data, int len)
 {
-    /* implicit cast abide compiler warning */
-    memcpy((void*)address, data, len);
+    int i;
+    if (forceEmergency == 1 && address == WOLFBOOT_PARTITION_BOOT_ADDRESS) {
+        /* implicit cast abide compiler warning */
+        memset((void*)address, 0, len);
+        /* let the rest of the writes work properly for the emergency update */
+        forceEmergency = 0;
+    }
+    else {
+        for (i = 0; i < len; i++) {
+#ifdef NVM_FLASH_WRITEONCE
+            if (((uint8_t*)address)[i] != FLASH_BYTE_ERASED) {
+                /* no writing to non-erased page in NVM_FLASH_WRITEONCE */
+                printf("NVM_FLASH_WRITEONCE non-erased write detected!\n");
+                return -1;
+            }
+#endif
+#ifdef WOLFBOOT_FLAGS_INVERT
+            ((uint8_t*)address)[i] |= data[i];
+#else
+            ((uint8_t*)address)[i] &= data[i];
+#endif
+        }
+    }
     return 0;
 }
 
@@ -150,8 +172,11 @@ void hal_init(void)
             erasefail_address = strtol(main_argv[++i], NULL,  16);
             fprintf(stderr, "Set power fail to erase at address %x\n",
                 erasefail_address);
-            break;
         }
+        /* force a bad write of the boot partition to trigger and test the
+         * emergency fallback feature */
+        else if (strcmp(main_argv[i], "emergency") == 0)
+            forceEmergency = 1;
     }
 }
 
--- a/include/wolfboot/wolfboot.h
+++ b/include/wolfboot/wolfboot.h
@@ -257,7 +257,9 @@ extern "C" {
 #ifndef WOLFBOOT_FLAGS_INVERT
 #define IMG_STATE_NEW       0xFF
 #define IMG_STATE_UPDATING  0x70
-#define IMG_STATE_FINAL_FLAGS  0x30
+/* now just an intermediary state, update state will always be either new or
+ * updating before the application boots*/
+#define IMG_STATE_FINAL_FLAGS 0x30
 #define IMG_STATE_TESTING   0x10
 #define IMG_STATE_SUCCESS   0x00
 #define FLASH_BYTE_ERASED   0xFF
@@ -265,8 +267,8 @@ extern "C" {
 #else
 #define IMG_STATE_NEW       0x00
 #define IMG_STATE_UPDATING  0x8F
-#define IMG_STATE_FINAL_FLAGS  0xBF
 #define IMG_STATE_TESTING   0xEF
+#define IMG_STATE_FINAL_FLAGS 0xBF
 #define IMG_STATE_SUCCESS   0xFF
 #define FLASH_BYTE_ERASED   0x00
 #define FLASH_WORD_ERASED   0x00000000UL
--- a/src/image.c
+++ b/src/image.c
@@ -611,8 +611,6 @@ static int image_sha256(struct wolfBoot_image *img, uint8_t *hash)
  */
 static void key_sha256(uint8_t key_slot, uint8_t *hash)
 {
-    int blksz;
-    unsigned int i = 0;
     uint8_t *pubkey = keystore_get_buffer(key_slot);
     int pubkey_sz = keystore_get_size(key_slot);
     wc_Sha256 sha256_ctx;
@@ -621,13 +619,7 @@ static void key_sha256(uint8_t key_slot, uint8_t *hash)
         return;
 
     wc_InitSha256(&sha256_ctx);
-    while (i < (uint32_t)pubkey_sz) {
-        blksz = WOLFBOOT_SHA_BLOCK_SIZE;
-        if ((i + blksz) > (uint32_t)pubkey_sz)
-            blksz = pubkey_sz - i;
-        wc_Sha256Update(&sha256_ctx, (pubkey + i), blksz);
-        i += blksz;
-    }
+    wc_Sha256Update(&sha256_ctx, pubkey, (word32)pubkey_sz);
     wc_Sha256Final(&sha256_ctx, hash);
 }
 #endif /* WOLFBOOT_NO_SIGN */
@@ -700,8 +692,6 @@ static int image_sha384(struct wolfBoot_image *img, uint8_t *hash)
  */
 static void key_sha384(uint8_t key_slot, uint8_t *hash)
 {
-    int blksz;
-    unsigned int i = 0;
     uint8_t *pubkey = keystore_get_buffer(key_slot);
     int pubkey_sz = keystore_get_size(key_slot);
     wc_Sha384 sha384_ctx;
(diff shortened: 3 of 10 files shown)


danielinux (changes requested):
let's modify hal_flash_write as discussed on slack, to ensure we simulate the same behavior as an actual FLASH memory (& operation on existing content). That will make the test fail upon invalid state changes

hal/sim.c
```diff
@@ -105,8 +106,15 @@ void hal_prepare_boot(void)
 
 int hal_flash_write(uintptr_t address, const uint8_t *data, int len)
 {
-    /* implicit cast abide compiler warning */
-    memcpy((void*)address, data, len);
+    if (forceEmergency == 1 && address == WOLFBOOT_PARTITION_BOOT_ADDRESS) {
+        /* implicit cast abide compiler warning */
+        memset((void*)address, 0, len);
```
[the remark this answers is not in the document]
jpbland1: no particular reason, just need to make sure the signature check fails to trigger the emergency fallback

src/update_flash.c
```diff
@@ -194,6 +194,91 @@ static int RAMFUNCTION wolfBoot_copy_sector(struct wolfBoot_image *src,
...
+#ifdef EXT_ENCRYPTED
+        + ENCRYPT_KEY_SIZE + ENCRYPT_NONCE_SIZE
+#endif
+    ];
+    /* open boot */
+    wolfBoot_open_image(boot, PART_BOOT);
+    /* open update */
+    wolfBoot_open_image(update, PART_UPDATE);
+    /* open swap */
+    wolfBoot_open_image(swap, PART_SWAP);
```
danielinux, line 223: ifndef NO_BACKUP before accessing SWAP

jpbland1: wolfBoot_swap_and_final_erase is not called when DISABLE_BACKUP is enabled since it doesn't have the same FSM that requires special handling of the final state setting

src/update_flash.c
```diff
@@ -715,21 +813,22 @@ void RAMFUNCTION wolfBoot_start(void)
...
     /* Check if the BOOT partition is still in TESTING,
      * to trigger fallback.
      */
-    if (bootRet == 0 && bootState == IMG_STATE_TESTING) {
-        /* wolfBoot_update_trigger now erases all the sector flags, only trigger
-         * if we're not already updating */
-        if (updateRet || updateState != IMG_STATE_UPDATING) {
-            wolfBoot_update_trigger();
+    if (resumedFinalErase != 0) {
+        if (bootRet == 0 && bootState == IMG_STATE_TESTING) {
```
danielinux, line 823: nit: parenthesis `((bootRet == 0) && (bootState == IMG_STATE_TESTING))`

src/update_flash.c
```diff
@@ -715,21 +813,22 @@ void RAMFUNCTION wolfBoot_start(void)
...
+            wolfBoot_update(1);
+        /* Check for new updates in the UPDATE partition or if we were
+         * interrupted during the flags setting */
+        }
+        else if (updateRet == 0 && updateState == IMG_STATE_UPDATING) {
```
danielinux, line 828: parenthesis

src/update_flash.c
```diff
@@ -458,20 +545,31 @@ static int RAMFUNCTION wolfBoot_update(int fallback_allowed)
...
-            if ((cur_v == 0) || (cur_v == up_v)) {
+        /* if magic isn't set stateRet will be -1 but that means we're on a
+         * fresh partition and aren't resuming */
+        stateRet = wolfBoot_get_partition_state(PART_UPDATE, &st);
+
+        /* if we've already written a sector or we've mangled the boot partition
+         * header we can't determine the direction by version numbers. instead
+         * use the update partition state, updating means regular, new means
+         * reverting */
+        if (stateRet == 0 && (flag != SECT_FLAG_NEW || cur_v == 0)) {
```
danielinux, line 556: parenthesis around "==" conditions please

src/libwolfboot.c
```diff
@@ -866,13 +860,9 @@ uint16_t wolfBoot_find_header(uint8_t *haystack, uint16_t type, uint8_t **ptr)
...
             break;
         }
-        if (*p == HDR_PADDING) {
-            /* Padding byte (skip one position) */
-            p++;
-            continue;
-        }
         /* Sanity check to prevent dereferencing unaligned half-words */
-        if ((((size_t)p) & 0x01) != 0) {
+        if (*p == HDR_PADDING || (((size_t)p) & 0x01) != 0) {
```
danielinux, line 864: nit: parenthesys around "==" conditions

danielinux (changes requested):
Looks good. Please fix style and squash.

dgarske:
@jpbland1 please work on cleaning up these minor issues and it will be ready for merge.

include/wolfboot/wolfboot.h
```diff
@@ -257,16 +257,16 @@ extern "C" {
 #ifndef WOLFBOOT_FLAGS_INVERT
 #define IMG_STATE_NEW       0xFF
 #define IMG_STATE_UPDATING  0x70
-#define IMG_STATE_FINAL_FLAGS  0x30
+#define IMG_STATE_FINAL_SWAP 0x30
```
dgarske, line 260: This value cannot be renamed because it will break customers. Either use the old name or add a new one.

dgarske (changes requested):

src/libwolfboot.c
```diff
@@ -866,13 +860,9 @@ uint16_t wolfBoot_find_header(uint8_t *haystack, uint16_t type, uint8_t **ptr)
...
         }
-        if (*p == HDR_PADDING) {
-            /* Padding byte (skip one position) */
-            p++;
-            continue;
-        }
         /* Sanity check to prevent dereferencing unaligned half-words */
-        if ((((size_t)p) & 0x01) != 0) {
+        if ((*p == HDR_PADDING) || ((((size_t)p) & 0x01) != 0)) {
+            /* Padding byte (skip one position) */
```
dgarske, line 865: Combing this logic is okay, but please update the comment. This is skipping a byte if its a padding byte or if its not an aligned half word.

src/image.c
```diff
@@ -710,13 +708,7 @@ static void key_sha384(uint8_t key_slot, uint8_t *hash)
         return;
 
     wc_InitSha384(&sha384_ctx);
-    while (i < (uint32_t)(pubkey_sz)) {
```
dgarske, line 713: Why are you removing this logic? Guessing it is not required since the public key buffer is fully available? Is it possible that WOLFBOOT_SHA_BLOCK_SIZE might be useful for other reasons?

src/libwolfboot.c
```diff
@@ -350,12 +350,8 @@ static int RAMFUNCTION trailer_write(uint8_t part, uintptr_t addr, uint8_t val)
     /* Calculate write address */
     addr_write = addr_align - ((!nvm_cached_sector) * NVM_CACHE_SIZE);
 
-    /* Ensure that the destination was erased, or force erase */
-    if (*((uint32_t *)(addr_write + NVM_CACHE_SIZE - sizeof(uint32_t)))
-            != FLASH_WORD_ERASED)
-    {
-        hal_flash_erase(addr_write, NVM_CACHE_SIZE);
-    }
+    /* Ensure that the destination was erased */
```
dgarske, line 353: Are you sure always doing the erase of the trailing here (vs conditional before) is the right assumption?

src/update_flash.c
```diff
@@ -194,6 +194,91 @@ static int RAMFUNCTION wolfBoot_copy_sector(struct wolfBoot_image *src,
...
 
+static int wolfBoot_swap_and_final_erase(int resume)
+{
+    struct wolfBoot_image boot[1];
+    struct wolfBoot_image update[1];
+    struct wolfBoot_image swap[1];
+    uint8_t st;
+    int eraseLen = WOLFBOOT_SECTOR_SIZE
+#ifdef NVM_FLASH_WRITEONCE
+    /* need to erase the redundant sector too */
```
dgarske, line 205: Comment indent should match code below.

src/update_flash.c
```diff
@@ -346,14 +431,9 @@ static int wolfBoot_delta_update(struct wolfBoot_image *boot,
...
         sector++;
     }
-    /* mark that our sector flags aren't reliable in testing mode */
-    st = IMG_STATE_FINAL_FLAGS;
-    wolfBoot_set_partition_state(PART_UPDATE, st);
-    /* mark boot partition as testing */
-    st = IMG_STATE_TESTING;
-    wolfBoot_set_partition_state(PART_BOOT, st);
+    /* start re-entrant final erase */
+    wolfBoot_swap_and_final_erase(0);
```
dgarske, line 435: Why is the return code not checked?

test-app/app_sim.c
```diff
@@ -52,6 +52,9 @@ int do_cmd(const char *cmd)
     if (strcmp(cmd, "powerfail") == 0) {
         return 1;
     }
+    if (strcmp(cmd, "emergency") == 0) {
```
dgarske, line 55: Please update documentation for Sim to explain "emergency"

src/update_flash.c
```diff
@@ -194,6 +194,91 @@ static int RAMFUNCTION wolfBoot_copy_sector(struct wolfBoot_image *src,
     return pos;
 }
 
+static int wolfBoot_swap_and_final_erase(int resume)
```
dgarske, line 197: Is this function always needed? Should it be wrapped with `#ifndef DISABLE_BACKUP`?

tools/test.mk
```diff
@@ -977,29 +977,29 @@ test-all: clean
 
 
 test-size-all:
-	make test-size SIGN=NONE LIMIT=4776
+	make test-size SIGN=NONE LIMIT=4917
```
dgarske, line 980: Are these the actual increased values? Would you like me to recompute updated actuals? On my Ubuntu 24 box I can reproduce the values that match GitHub CI.

dgarske (changes requested):

src/update_flash.c
jpbland1: I had thought it was never called for DISABLE_BACKUP but it's also called in wolfBoot_start, good point

src/update_flash.c
jpbland1: return code is only to tell wolfBoot_start if the final swap was resumed or not

src/libwolfboot.c
jpbland1: this was part of code size reduction, worst case scenario from the old code we call erased on an already erased sector

src/image.c
jpbland1: wolfCrypts internal logic already loops over in the correct block size increments, the keybuffer is fully available so it's just wasted code size

dgarske: This is a different block size value. This is used to break the SHA processing into smaller chunks. This is typically used when the value being hashed comes from flash. This would read (by default) 128 bytes at a time.

src/update_flash.c
dgarske: Then please document this in the code at this line.... why we don't need to check the return code.

src/image.c
jpbland1: good to know, then it's almost certainly being re-adjusted by
```
        if ((i + blksz) > (uint32_t)pubkey_sz)
            blksz = pubkey_sz - i;
```
for ed and ecc keys, maybe rsa has bigger keys but in any case the entire key is still available and `wc_Sha384Update` can handle it all in a single function call

dgarske: I agree and I am okay removing this logic. Please make sure you do it consistently for the other algorithms too.

jpbland1: No fun! I wanted to make an intern do that

dgarske (approved):

PR #468, opened 2024-07-09
Add command line tool to generate otp.bin

danielinux:
Instead of using the `otp-keystore-primer`, an `otp.bin` image can now be generated and flashed separately using the newly introduced `otp-keystore-gen` tool.

### DIFF
--- a/tools/keytools/otp/otp-keystore-gen.c
+++ b/tools/keytools/otp/otp-keystore-gen.c
@@ -1,0 +1,89 @@
+/* otp-keystore-primer.c
+ *
+ * Command line utility to create a OTP image
+ *
+ *
+ * Copyright (C) 2024 wolfSSL Inc.
+ *
+ * This file is part of wolfBoot.
+ *
+ * wolfBoot is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * wolfBoot is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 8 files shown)


tools/keytools/otp/otp-keystore-gen.c
```diff
@@ -0,0 +1,91 @@
+/* otp-keystore-primer.c
```
dgarske, line 1: otp-keystore-gen.c

tools/keytools/otp/otp-keystore-gen.c
```diff
@@ -0,0 +1,91 @@
...
+#include <stdint.h>
+#include <stdio.h>
+#include <string.h>
+#include <stddef.h>
+#include <stdlib.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <errno.h>
+
+#define OTP_SIZE 4096
```
dgarske, line 33: Should this be overridable? Does this vary?

tools/keytools/otp/otp-keystore-gen.c
dgarske: This seems like a useful generic tool not specific to OTP that allows a .bin version of the keystore. This would be useful if someone wanted to store the keystore in a specific location or flash in a different way.

danielinux: "keystore.der" is already a binary version of the keystore. This otp.bin also includes the "magic" number at the beginning to be specifically stored in OTP. Other storage might want to use custom headers, as they will have to provide the keystore functions.

tools/keytools/otp/otp-keystore-gen.c
danielinux: It's just an upper limit, but the actual size of the OTP should be enforced by the flashing tool. It's useful in the "primer" case, because it prevent silently overflowing the actual OTP storage size. For other components, this is part of the hal declarations (see h5/h7 and how this value is set in the primer)

dgarske (approved):

PR #469, opened 2024-07-16
New unit tests for NVM_FLASH_WRITEONCE, fix bug in offset calculation

danielinux:
validating `NVM_FLASH_WRITEONCE` state machine sequences using specific unit tests.

Found a :bug:  that can cause out-of-bound access  when electing the current bank, due to wrong sign in offset. Fixing.

PR #470, opened 2024-07-17
Fixed flag comparison selecting most recent flags

danielinux:
Following up from #469 and expanding unit tests for NVM_FLASH_WRITEONCE, to consider all corner cases.

- Added an explicit check for both uninitialized/invalid trailers, which now overrides 27c079b04 fix for "First time boot".
- Once the headers are already compared, the flags comparison loop can start from offset '4' instead, checking for differences in the partition flags byte and the sector flags nibbles behind.
- Expanded unit tests to cover all cases


danielinux:
Ready to review

dgarske (approved):

PR #476, opened 2024-07-22
Fixes for Renesas RX IRQ on boot and addition of SPI flash full chip erase API

dgarske:
* Fix to make sure wolfBoot leaves IRQ's disabled. The TSIP driver will leave IRQ's on which could cause customer issues. 
* Added new `spi_flash_chip_erase` API for the SPI flash driver. Enabled with `SPI_FLASH_CHIP_ERASE`.

### DIFF
--- a/hal/renesas-rx.c
+++ b/hal/renesas-rx.c
@@ -433,14 +433,7 @@ void hal_init(void)
 
 void hal_prepare_boot(void)
 {
-    /* make sure interrupts are disabled */
-#if defined(__CCRX__)
-    clrpsw_i();
-#elif defined(__GNUC__)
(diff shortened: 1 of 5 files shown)


hal/renesas-rx.c
```diff
@@ -433,7 +433,14 @@ void hal_init(void)
 
 void hal_prepare_boot(void)
 {
-
+    /* make sure interrupts are disabled */
```
danielinux, line 436: This is usually done in `do_boot` because it's in general architecture-specific rather than target-. Check e.g. src/boot_arm.c:357.
Consider moving to `boot_renesas.c`.

danielinux (approved):

PR #478, opened 2024-07-23
Fix the size of trailer status for `NVM_FLASH_WRITEONCE`

dgarske:
Fix the size of trailer status for `NVM_FLASH_WRITEONCE` (was using 1 extra sector).


jpbland1 (commented):
This is not correct, the tmpBootPos is the second to last sector of boot. We use swap the second to last sector to swap, then use it to hold the encryption key and mark with the special flag, then swap the original contents back after the last, or last 2 for writeonce, have been erased.

PR #480, opened 2024-07-29
Add HAB support irrespective of MCUSDK setting for imx_rt target

MulattoKid:
After https://github.com/wolfSSL/wolfBoot/pull/475 moved the selection of linker script from `Makefile` to `arch.mk`, building with `TARGET_IMX_HAB=1` only worked when `MCUSDK` wasn't set to 1. This PR fixes this.

cc @danielinux


dgarske (approved):
Thanks @MulattoKid . Contributor agreement on file.

PR #481, opened 2024-07-31
Refactoring of pkcs11_store.c module

danielinux:
-  [x] add unit tests for `pkcs11_store.c`
-  [x] defects detected by unit tests are fixed
-  [x] refactoring with a new powerfail-safe design

This new design uses two sectors at the beginning of the keyvault partition.
The first sector contains the allocation tables for the nodes in flash, including
a bitmap to keep track of the free slots.
The second sector is a pre-commit 'backup' sector, used to update the content of the
vault in two steps. This guarantees powerfail protection.

All hal_flash_erase / hal_flash_write operations are now on entire sectors.
Sector content is cached to memory, then written to the backup sector, then
committed to the actual destination sector.


danielinux:
Awaiting feedback (zd18366)

src/pkcs11_store.c
```diff
@@ -97,36 +101,39 @@ int wolfPKCS11_Store_Open(int type, CK_ULONG id1, CK_ULONG id2, int read,
...
-    if ((!found) && read) {
+    if ((found == -1) && read) {
         *store = NULL;
         return NOT_AVAILABLE_E;
-    } else if (found) {
-        *store = vault_descriptors[found];
-        obj = vault_descriptors[found];
-        memcpy(&obj->hdr, vault_base + found * KEYVAULT_OBJ_SIZE, sizeof(struct obj_hdr));
+    }
+    obj = XMALLOC(sizeof(struct store_object), NULL, DYNAMIC_TYPE_TMP_BUFFER);
```
dgarske, line 120: Consider cast `(struct store_object*)`.

src/pkcs11_store.c
```diff
@@ -97,36 +101,39 @@ int wolfPKCS11_Store_Open(int type, CK_ULONG id1, CK_ULONG id2, int read,
...
+    if (!obj) {
```
dgarske, line 121: Prefer `obj == NULL`... thus not assuming NULL == 0.

src/pkcs11_store.c
dgarske, line 120: Consider version that doesn't require heap..

dgarske (changes requested):

src/pkcs11_store.c
danielinux: XMALLOC is gone now

src/pkcs11_store.c
danielinux: ditto

src/pkcs11_store.c
danielinux: I have removed the dynamic allocation in the last refactoring, however it seems that wolfPKCS11 still requires a heap, so I re-introduced the _sbrk in the designated pool.

danielinux: Ideally, since wolfPKCS11 uses its own area, this heap can be in a bank that is securely erased on powerdown

dgarske (dismissed):

danielinux:
Ready to review

dgarske (dismissed):

PR #483, opened 2024-08-07
Added unit tests for delta.c

danielinux:


tools/unit-tests/unit-delta.c
```diff
@@ -0,0 +1,160 @@
+#include <check.h>
```
dgarske, line 1: License header?

dgarske (changes requested):

tools/unit-tests/unit-delta.c
danielinux: it's missing in  a few unit tests file. Adding.

PR #488, opened 2024-08-13
Unit test for update_ram.c

danielinux:
More unit tests to verify update_ram.c

- Discovered a loop when falling back in partition election. Fixed.
- Discovered a wrong hal_flash_unlock instead of ext_flash_lock when updating partition states on EXT_FLASH

This PR is on top of #487 - please merge that one first.


dgarske (approved):
Fully tested. Thank you @danielinux !

Issue #489, opened 2024-08-15
Non-RAM functions in hal_flash_erase/write on iMX RT1060


MulattoKid:
Hi,

After increasing the value of `WOLFBOOT_SECTOR_SIZE` to reduce the wear on the flash sectors used for swap during updates, we've started to experience issues with hangs in `hal_flash_erase` and `hal_flash_write`. This happens occasionally when starting from a totally erased flash, and then programming the bootloader and application image. It happens when we call `wolfBoot_success`, as it causes erases and writes to the last two sectors in the BOOT and UPDATE partitions (wolfBoot housekeeping).

Upon investigation, we've narrowed it down to the use of `memcpy` and `DCACHE_InvalidateByRange` in both flash functions. I've previously added disabling of interrupts and invalidating the DCache in these functions to ensure no access over FlexSPI while erasing/writing to flash and data coherency. However, the functions used here should also be in RAM, so as to not perform accesses over FlexSPI.

This has always been an issue for us, but the likelihood of this going bad increases when the sector size increases.

**memcpy**
wolfBoot has its own `memcpy` implementation which is declared with `RAMFUNCTION`. So long as both `__WOLFBOOT` and `RAM_CODE` are defined when compiling wolfBoot, this is fine. However, we define `WOLFBOOT_USE_STDLIBC`, which means our newlib `memcpy` is used, which gets placed in flash, not RAM. This is also true when compiling libwolfBoot for our application. We therefore had to add the following to wolfBoot's linker script:
```
    .text :
    {
      <..>
      *(EXCLUDE_FILE(
          *libc.a:lib_a-memcpy.o
      ) .text)                 /* .text sections (code) */
      *(EXCLUDE_FILE(
          *libc.a:lib_a-memcpy.o
      ) .text*)                /* .text* sections (code) */
      *(EXCLUDE_FILE(
          *libc.a:lib_a-memcpy.o
      ) .rodata)               /* .rodata sections (constants, strings, etc.) */
      *(EXCLUDE_FILE(
          *libc.a:lib_a-memcpy.o
      ) .rodata*)              /* .rodata* sections (constants, strings, etc.) */
      <...>
    } > FLASH

    .data : AT (_stored_data)
    {
      <...>
      KEEP(*libc.a:lib_a-memcpy.o (.text*))
      <...>
    } > RAM
```

and this to our link script to ensure this function is placed in RAM instead of flash:
```
    .text :
    {
      <...>
[...]

PR #492, opened 2024-08-22
Improve flash configs for i.MX RT targets

MulattoKid:
This PR does two things:
1. Enable `RAM_CODE=1` in all i.MX RT target example configs. When running XIP this is needed to ensure `memcpy` (which is used during `hal_flash_write`) is in RAM, and not flash as that could result in reading from flash while writing to flash.
2. Change both `CPU_MIMXRT1061CVJ5B` and `CPU_MIMXRT1042XJM5B` to define the `USE_GET_CONFIG` macro. This makes the flash configuration's LUT a lot smaller, and the MCU will instead query the various settings from the flash chip. Note that the `CPU_MIMXRT1052DVJ6B` target (RT1050 family) doesn't support this option.

Ref: https://github.com/wolfSSL/wolfBoot/issues/489


dgarske:
Okay to test. Contributor agreement on file.

bigbrett:
This looks good, but I'm troubled that you are still experiencing hangs. I'd like to hold of on merging this until I can test on more targets as well as figure out the root cause of the ROM API hanging

MulattoKid:
My first test was to add `RAMFUNCTION` to the definition of `hal_flash_init`, and with that things seem to work. I'm not able to hit the hang. As mentioned in https://github.com/wolfSSL/wolfBoot/pull/493 I do believe adding a `DSB` barrier would still be an additional safety measure that could be added - it makes sense in my head that it's needed.

If it's OK I'll combine the these two PRs into one, as they're quite connected, and adding `DSB` would be needed in both.

MulattoKid:
Moved to new PR: https://github.com/wolfSSL/wolfBoot/pull/495

PR #493, opened 2024-08-22
Move DCACHE_InvalidateByRange to outside the loop in hal_flash_write

MulattoKid:
This PR does two things:
1. Ensure the arguments passed to `DCACHE_InvalidateByRange` both are 32-byte aligned (required by the function as cache lines are 32 bytes)
2. Move the call to `DCACHE_InvalidateByRange` to be outside the for-loop in `hal_flash_write`. This is needed to ensure reads aren't performed to flash (i.e. to read the code for the function) while writes are happening, as that's not allowed behavior. Another option would be to move the function from flash to RAM, but that is a bit more involved so I opted for this solution.

I do wonder if the solution implemented in 2. is OK with you? With this solution the cache isn't invalidated between each iteration of the loop, so should one iteration fail and return early (`return -1`) the DCache won't be invalidated at all. Is this OK wrt. to the rest of wolfBoot's logic when a call to `hal_flash_write` fails? If not, I can move the function to RAM.

Ref: https://github.com/wolfSSL/wolfBoot/issues/489

### DIFF
--- a/hal/imx_rt.c
+++ b/hal/imx_rt.c
@@ -880,17 +880,23 @@ int RAMFUNCTION hal_flash_write(uint32_t address, const uint8_t *data, int len)
         memcpy(wbuf, data + i, CONFIG_FLASH_PAGE_SIZE);
         status = g_bootloaderTree->flexSpiNorDriver->program(0, FLEXSPI_CONFIG,
             (address + i) - FLASH_BASE, wbuf);
-        /**
-         * Flash is memory mapped, so the address range must be invalidated in data cache
-         * to ensure coherency between flash and cache
-         */
-        DCACHE_InvalidateByRange(address + i, sizeof(wbuf));
         if (status != kStatus_Success)
         {
             asm volatile("cpsie i");
             return -1;
         }
     }
+    /**
+     * Flash is memory mapped, so the address range must be invalidated in data cache
+     * to ensure coherency between flash and cache.
+     * 
+     * Also, both the address and size must be 32-byte aligned as cache-lines are 32 bytes
+     * (see definition of DCACHE_InvalidateByRange).
+     * To ensure all data is included we align the address downwards, and the length upwards.
+     */
+    uint32_t aligned_address = address - (address % 32);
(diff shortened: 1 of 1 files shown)


bigbrett (commented):
> Move the call to DCACHE_InvalidateByRange to be outside the for-loop in hal_flash_write. This is needed to ensure reads aren't performed to flash (i.e. to read the code for the function) while writes are happening, as that's not allowed behavior. 

Note there is nothing wrong with this, and invalidating in one-go is probably much more performant, so a good change. But I was under the impression that the ROM API program function waits until the operation has succeeded/failed before returning, so I'm not sure this necessarily would fix anything.. unless a pipelined/speculative fetch happening concurrently with the program operation is what was causing the issues? If that is the case, a fencing instruction like DMB/DSB/ISB etc. would also be warranted. Just wanted to get insight as to what you thought the problem was.

> I do wonder if the solution implemented in 2. is OK with you? With this solution the cache isn't invalidated between each iteration of the loop, so should one iteration fail and return early (return -1) the DCache won't be invalidated at all. Is this OK wrt. to the rest of wolfBoot's logic when a call to hal_flash_write fails? If not, I can move the function to RAM.

@MulattoKid wolfBoot doesn't really error check `hal_flash_write()` return values, as the inability to write to flash is a critical failure and the only real solution would be to panic, so often it just takes  a best effort approach, and all bets are off. Therefore I don't think flushing DCACHE if the write fails in order to get at least a partial write matters. I'm okay with it as written, though one could also flush in the `if (status != kStatus_Success)` case before returning.

MulattoKid:
> unless a pipelined/speculative fetch happening concurrently with the program operation is what was causing the issues? If that is the case, a fencing instruction like DMB/DSB/ISB etc. would also be warranted. Just wanted to get insight as to what you thought the problem was.

Yes, that's what we were thinking could be happening. I'll try with barriers today.

MulattoKid:
Moved to new PR: https://github.com/wolfSSL/wolfBoot/pull/495

PR #494, opened 2024-08-23
Support for Microchip SAMA5D3

danielinux:
Add support for ATSAMA5D3. 
This is wolfBoot's first 32-bit Cortex-A target, using `boot_arm32.c` and a reset vector in assembly `boot_arm32_start.S`.


hal/sama5d3.c
```diff
@@ -0,0 +1,760 @@
...
+  return NULL; /* Not yet supported */
+}
+
+/* QSPI functions */
+void qspi_init(uint32_t cpu_clock, uint32_t flash_freq)
+{
+}
+
+
+void zynq_init(uint32_t cpu_clock)
```
dgarske, line 721: Fairly sure you can remove this.

hal/sama5d3.c
```diff
@@ -0,0 +1,760 @@
...
+}
+
+void* hal_get_dts_update_address(void)
+{
+  return NULL; /* Not yet supported */
+}
+
+/* QSPI functions */
+void qspi_init(uint32_t cpu_clock, uint32_t flash_freq)
+{
```
dgarske, line 717: Might consider TODO comment and `(void)cpu_clock`, etc.

dgarske (changes requested):
Very nice! Just a few minor comments.

hal/sama5d3.c
```diff
@@ -0,0 +1,760 @@
...
+	if (quotient)
+		*quotient = factor;
+
+	if (remainder)
+		*remainder = dividend;
+
+	return 0;
+}
+
+static uint32_t div(uint32_t dividend, uint32_t divisor)
```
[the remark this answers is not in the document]
danielinux: I don't think it'd be generic enough. Let's consider moving to ARM32 port file later, if this comes up again on another 32bit ARM

dgarske (approved):
Very nice!

PR #495, opened 2024-08-23
Fix flash interactions for i.MX RT targets

MulattoKid:
This PR does a few things:
1. Enable `RAM_CODE=1` in all i.MX RT target example configs. When running XIP this is needed to ensure `memcpy` (which is used during `hal_flash_write`) is in RAM, and not flash as that could result in reading from flash while writing to flash.
2. Change both `CPU_MIMXRT1061CVJ5B` and `CPU_MIMXRT1042XJM5B` to define the `USE_GET_CONFIG` macro. This makes the flash configuration's LUT a lot smaller, and the MCU will instead query the various settings from the flash chip. Note that the `CPU_MIMXRT1052DVJ6B` target (RT1050 family) doesn't support this option.
3. Ensure the arguments passed to `DCACHE_InvalidateByRange` both are 32-byte aligned (required by the function as cache lines are 32 bytes)
4. Move the call to `DCACHE_InvalidateByRange` to be outside the for-loop in `hal_flash_write`. This is needed to ensure reads aren't performed to flash (i.e. to read the code for the function) while writes are happening, as that's not allowed behavior. Another option would be to move the function from flash to RAM, but that is a bit more involved so I opted for this solution.
5. Change the return pattern for `hal_flash_write` to only have one return statement (to avoid duplicate code due to below point)
6. Add a data synchronization barrier (`asm volatile ("dsb")`) after accessing flash through the FlexSPI ROM API to ensure speculative pre-fetching doesn't violate the rule about reading from flash while erasing/programming it

I've tested this several times now, and am not able to hit the hang issue we've seen before (see issue ref below).

Ref:
- Based on PRs:
  - https://github.com/wolfSSL/wolfBoot/pull/492
  - https://github.com/wolfSSL/wolfBoot/pull/493
- Issue: https://github.com/wolfSSL/wolfBoot/issues/489


hal/imx_rt.c
```diff
@@ -876,23 +939,33 @@ int RAMFUNCTION hal_flash_write(uint32_t address, const uint8_t *data, int len)
...
+     * (see definition of DCACHE_InvalidateByRange).
+     * To ensure all data is included we align the address downwards, and the length upwards.
+     */
+    uint32_t aligned_address = address - (address % 32);
+    uint32_t aligned_len = len + (32 - (len % 32));
+    DCACHE_InvalidateByRange(aligned_address, aligned_len);
+    /* Re-enable interrupts */
     asm volatile("cpsie i");
-    return 0;
+    return program_success;
```
bigbrett, line 968: Although our return checking for writes is lax, technically this function should return 0 on sucess and -1 on failure, according to the docs. If we could strip the use of `bool` and return these codes that would be great

bigbrett (changes requested):
Looks great, just one quick fix needed on return values. I can fix other nits in a subsequent PR, as I have some fixes for 1064 to add.

Let me know if your testing has turned up any further hanging, or if it still running smoothly with these changes.

MulattoKid:
> Let me know if your testing has turned up any further hanging, or if it still running smoothly with these changes.

Have tested some more, and still no issues :+1:

bigbrett (approved):

PR #496, opened 2024-08-23
Add Raspberry Pi3 Uart

miyazakh:
Add use case on Raspberry Pi 3 B+ HW


hal/raspi3.c
```diff
@@ -18,23 +18,209 @@
...
+char uart_read(void)
+{
+    char c;
+    /* wait until data is comming */
+    do{
+        asm volatile("nop");
+    } while(*UART0_FR & 0x10);
+    /* read it and return */
+    c = (char)(*UART0_DR);
+    /* convert carrige return to newline */
```
danielinux, line 155: spelling: carriage. Is this conversion really required?

hal/raspi3.c
```diff
@@ -18,23 +18,209 @@
...
+}
+
+/**
+ * Send string to UART
+ */
+void uart_write(const char* buf, uint32_t sz) {
+    uint32_t len = sz;
+
+    while (len > 0 && *buf) {
+        /* convert newline to carrige return + newline */
```
danielinux, line 169: spelling, plus conversion from CR to CR+NL, that should be handled by the caller.

danielinux (changes requested):
Good addition. Thanks! Just left a couple of comments on the spelling of "carriage" in "carriage return". In the same context, I'm wondering if these uart API should not handle the CR to CR+LF conversion at all.

hal/raspi3.c
miyazakh: Thanks. Agree. Carriage return code should be handled by the caller.

danielinux (approved):

PR #500, opened 2024-08-30
fix ccrx use case for rx72n

miyazakh:


dgarske (approved):

IDE/Renesas/e2studio/RX72N/include/user_settings.h
```diff
@@ -49,6 +49,9 @@
     /* #define WOLFBOOT_SIGN_ECC256  */
     /* #define WOLFBOOT_SIGN_ECC384  */
     /* #define WOLFBOOT_SIGN_ECC521  */
+    #if !defined(min)
+    	#define min(data1, data2)    _builtin_min(data1, data2)
+	#endif
```
danielinux, line 54: align "endif" to its "if", replace tab with spaces.

danielinux: looks like there's a tab in L53 as well, wguke we use spaces everywhere.

miyazakh: Thanks. e2Studio inserts tab as default.

danielinux (approved):

PR #508, opened 2024-10-09
Fixes for SAMA5D3

danielinux:
- Added support for SHA acceleration via ARMASM (arm-v8). Requires [this fix in wolfcrypt](https://github.com/wolfSSL/wolfssl/pull/8035)
- Generic GPIO interface
- Added MAC PIN initialization sequence
- Enabled DBGU


danielinux:
Depends on #509

tools/test.mk
philljj, line 1: These changes to `tools/test.mk` are already in the merged #509 I think.

danielinux: Just rebased. Thanks!

philljj (approved):

PR #520, opened 2024-11-21
Delta update: check sha digest of base image

danielinux:
Only allow delta update if the sha digest of the base image matches the expected value. This is to prevent the delta update from being applied to an image that is not the expected base image, even if the version matches.

The mismatch should never happen in real-life scenarios, but it is adviseable to check for false positives during integration, testing and development.


src/update_flash.c
```diff
@@ -327,12 +331,34 @@ static int wolfBoot_delta_update(struct wolfBoot_image *boot,
...
+    if (wolfBoot_get_delta_info(PART_UPDATE, inverse, &img_offset, &img_size,
+                &delta_base_hash, &delta_base_hash_sz) < 0) {
         return -1;
     }
     cur_v = wolfBoot_current_firmware_version();
     upd_v = wolfBoot_update_firmware_version();
     delta_base_v = wolfBoot_get_diffbase_version(PART_UPDATE);
+
+    if (delta_base_hash_sz != WOLFBOOT_SHA_DIGEST_SIZE) {
+        wolfBoot_printf("Delta update: Base hash size mismatch"
```
dgarske, line 343: Should we have some type of backwards compatibility when the delta_base_v is not found? `Delta update: Base hash size mismatch (size: 0 expected 20)`

```
git checkout master
git distclean
make distclean
cp config/examples/sim-delta-update.config .config
make -C tools/keytools && make -C tools/bin-assemble
make clean && make test-sim-internal-flash-with-delta-update
git checkout delta-base-hash
make wolfboot.elf
./wolfboot.elf update_trigger get_version
./wolfboot.elf success get_version
```

dgarske (changes requested):

src/update_flash.c
danielinux: Done. Also added a (documented) option `--no-base-sha` for `sign`.
No SHA is not fatal now for wolfBoot.
Also added a test to prove that.

dgarske (approved):

PR #525, opened 2024-12-06
Fix offset updating trailer: swap_and_final_erase

danielinux:
When wolfBoot_swap_and_final_erase() is invoked, it was calculating the wrong offset, ending up erasing the third-last sector of the boot partition.

See ZD19006


danielinux:
This is not the right fix (see zd19006). Closing.

PR #526, opened 2024-12-07
properly unlock flash on re-entry of wolfBoot_swap_and_final_erase

jpbland1:
and move it out of the lock logic of update and delta update


jpbland1:
added hal and ext flash locks to the simulator, fixed test failures caused by flash being accessed while locked. @danielinux any idea why the unit test is failing? is it because I added the locks to hal/sim.c and it's not aware of them?

danielinux:
The hal_flash_lock/unlock functions are mocked and checked, it seems that the unit test is dectecting a double lock.

jpbland1:
I figured out the issue, it was only unlocking flash when encryption was on, I think it was a copy and paste issue. should be good to go now

src/update_flash.c
```diff
@@ -480,15 +494,17 @@ static int wolfBoot_delta_update(struct wolfBoot_image *boot,
...
 out:
 #ifdef EXT_FLASH
     ext_flash_lock();
 #endif
     hal_flash_lock();
+    /* start re-entrant final erase, return code is only for resumption in
+     * wolfBoot_start */
+    if (ret == 0) {
+        wolfBoot_swap_and_final_erase(0);
+    }
```
danielinux, line 507: I think this should still be under `#ifndef DISABLE_BACKUP`, unless it's needed anyway. If that's the case, remove the empty `#ifndef` clause at L 496 then.

danielinux (changes requested):
Looks good. Good idea adding checks for unlock/lock in the simulator as well.

Just a doubt about `DISABLE_BACKUP` calling `wolfBoot_swap_and_final_erase()`

src/update_flash.c
jpbland1: missed that, fixed now

danielinux (approved):

PR #530, opened 2024-12-23
Fix wolfBoot_update_trigger sector calculation

MulattoKid:
`wolfBoot_update_trigger` is called to initiate an update on the next boot. It does this by modifying the metadata in the last sector (selecting one of the two if `NVM_FLASH_WRITEONCE` is defined).

There's a check to handle the case where the `PART_UPDATE_ENDFLAGS` isn't aligned with the `WOLFBOOT_SECTOR_SIZE`. However, this check was wrong, as the modulo of these two macros is 0 when aligned, so the check should be != 0.

See ticket #19133 for more information.

### DIFF
--- a/src/libwolfboot.c
+++ b/src/libwolfboot.c
@@ -742,7 +742,7 @@ void RAMFUNCTION wolfBoot_update_trigger(void)
 
     /* if PART_UPDATE_ENDFLAGS straddles a sector, (all non FLAGS_HOME builds)
      * align it to the correct sector */
-    if (PART_UPDATE_ENDFLAGS % WOLFBOOT_SECTOR_SIZE == 0)
+    if (PART_UPDATE_ENDFLAGS % WOLFBOOT_SECTOR_SIZE != 0)
         lastSector -= WOLFBOOT_SECTOR_SIZE;
 
     /* erase the sector flags */


MulattoKid:
I've tested this, and performing an update now works. Without this change the second and third to last sectors are chosen to work with, instead of the last and second to last sectors, resulting in an invalid SHA integrity check when our image spans the entire available region (excluding the two sectors reserved by wolfBoot at the end of the `BOOT` and `UPDATE` partitions).

dgarske:
Unit test says:

```
83%: Checks: 6, Failures: 1, Errors: 0
unit-common.c:53:F:verify_integrity:test_verify_integrity:0: Assertion 'address + len <= (33 * 1024)' failed: address + len == 94631014433515, (33 * 1024) == 33792
make: *** [Makefile:37: run] Error 1
```

If the fix is valid it's possible a unit test may need updated. @jpbland1 please review.

dgarske:
Closing in favor of https://github.com/wolfSSL/wolfBoot/pull/531

MulattoKid:
Thanks for looking into it!

MulattoKid:
Just noting that #531 seems to fix the issue

PR #540, opened 2025-01-24
Add support for Armored mode with IAR

dgarske:
Add support for Armored mode with IAR. Currently only supports ECDSA and Cortex-M. ZD19190

### DIFF
--- a/IDE/IAR/README.md
+++ b/IDE/IAR/README.md
@@ -90,5 +90,6 @@ Using the ST-LINK Utility, perform the following steps:
 
 If you are using a STM32F407-discovery board, a red LED will turn on upon application boot.
 
+##  Armored Mode (Glitch Resistance)
 
-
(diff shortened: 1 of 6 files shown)


danielinux (changes requested):
Looks equivalent to GCC. Please double check wolfBoot_panic() which does not seem to be taken into account here.

dgarske:
> Looks equivalent to GCC. Please double check wolfBoot_panic() which does not seem to be taken into account here.

@danielinux I've confirmed the armored version of wolfBoot_panic() for armored builds and works correctly in IAR. That code builds correctly in IAR with WOLFBOOT_ARMORED set and is being compiled in. This PR is good as-is.

danielinux (approved):

danielinux:
Thanks for double checking.

PR #541, opened 2025-01-26
Add support for STM32F103 (Blue Pill)

Risca:
The Blue Pill board is typically fitted with a STM32F103C8T6, but
reports online seem to indicate that it sometimes can come with its
bigger sibling the STM32F103C8TB as is the case with my board.

This patch adds support for all low- to medium-density STM32F103
devices. All of these devices have their flash divided in 1 KB pages, up
to 128 KB. The high-density and connectivity line of devices, sometimes
called XL devices, use 2 KB pages and are not supported.

Similar to the STM32F4 support, the code will spin up the PLL for
maximum speed during init and turn it off just before booting the
application. the maxium speed is 72 MHz.

Booting with the example config takes around 6 seconds.
Updating takes about 15.


hal/stm32f103.c
```diff
@@ -0,0 +1,324 @@
...
+        if (*(uint32_t*)address != (uint32_t)~0)
+            return -1;
+
+        address += 4;
+    }
+
+    return 0;
+}
+
+int RAMFUNCTION hal_flash_write(uint32_t address, const uint8_t *data, int len)
```
danielinux, line 148: Don't worry at all about erasing the page before write. This task is performed by wolfBoot, that will always call erase before writing. If the flash memory does not support consecutive writes after erase, just add NVM_FLASH_WRITEONCE=1 to the example config, like for similar STM32 targets.

Please remove any erase commands issued in hal_flash_write.

arch.mk
```diff
@@ -139,6 +139,12 @@ ifeq ($(ARCH),ARM)
     ARCH_FLASH_OFFSET=0x08000000
   endif
 
+  ifeq ($(TARGET),stm32f103)
```
danielinux, line 142: I think we should rename the target (and the associated files) to `stm32f1` for consistency with the other ports, to cover the entire family.

Comments in the files, referring to STM32F103, are OK, and even mentioning the BluePill in the test-app as you did is actually helpful as reference if anyone is testing on a different STM32F1.

danielinux (dismissed):
Thanks @Risca for the proposed port. This looks good so far.
- Please do get in touch with us at support@wolfssl.com to fill our Contributor Agreement form, if you haven't already, mentioning this PR.
- The hal_flash_write function is not supposed to erase sectors. Please remove all erase commands from `hal_flash_write`.
- Please rename the TARGET and file names to `stm32f1` for consistency with other ports.
- Our ports usually include also one paragraph of instructions in docs/Targets.md, and a non-regression build test in github actions, but I could also commit those on top while testing, once the above points are solved.

hal/stm32f103.c
Risca: Ah, I see! I couldn't figure out if this was the case so I opted for the safe approach. This makes the implementation much simpler!

arch.mk
Risca: Sure!

Risca:
> Thanks @Risca for the proposed port. This looks good so far.
> 
>     * Please do get in touch with us at [support@wolfssl.com](mailto:support@wolfssl.com) to fill our Contributor Agreement form, if you haven't already, mentioning this PR.
> 
>     * The hal_flash_write function is not supposed to erase sectors. Please remove all erase commands from `hal_flash_write`.
> 
>     * Please rename the TARGET and file names to `stm32f1` for consistency with other ports.
> 
>     * Our ports usually include also one paragraph of instructions in docs/Targets.md, and a non-regression build test in github actions, but I could also commit those on top while testing, once the above points are solved.

I've fixed the rename and (hopefully) the build test. The rest will be a few days

danielinux:
Rebasing on master would get rid of the failing tests, which are caused by the deprecation of some v3 APIs we already fixed.

hal/stm32f103.c
Risca: So, the STM32F1 do support consecutive writes after erase, but it only supports 16 bit writes. I'm trying out an idea to see if a read-modify-write approach to unaligned writes would work but will take some more testing to verify. I currently have a single LED as my only debugging tool for Wolfboot, but I might be able to do some tests with Zephyr or arduino to test it out. It will take a little more time, unfortunately

Risca:
Rebased without changes to latest master to see if automatic builds now works

dgarske:
Tracking in ZD 19403

hal/stm32f103.c
Risca: I've done some testing using Zephyr, and I can confirm that writing to flash only works when:

- the write is a 2 byte operation
- the area is 0xFFFF (erased)

Writing 2 bytes to an area that is _not_ erased does nothing, even if it's `0xFFaa` or `0xbbFF` and you write `0xccaa` or `0xccFF`.

I realize that I misunderstood what `NVM_FLASH_WRITEONCE=1` meant. I thought it meant that writes to a whole _sector_ had to be written at once. After reading the docs better I realize that it's meant to address the exact issue I just described. I'll clean up the code and add `NVM_FLASH_WRITEONCE=1` to the example config.

Risca: The code should be good for review now.

I've received the Contributor Agreement form, but haven't got around to reading or signing it yet.

Risca:
Fixed an issue in the config file: the wolfboot partition was configured much larger than the space available before the first (boot) partition

Risca:
> Fixed an issue in the config file: the wolfboot partition was configured much larger than the space available before the first (boot) partition

Sorry for the noise - I misread the docs. Config is fixed now, and the partition sizes trimmed a bit

hal/stm32f103.c
Risca: The form should be signed now

dgarske:
Contributor agreement approved and on file. Thank you @Risca . Daniele please finalize and merge when you are happy with it.

dgarske (approved):

hal/stm32f1.c
```diff
@@ -0,0 +1,342 @@
+/* stm32f1.c
+ *
+ * Copyright (C) 2021 wolfSSL Inc.
+ *
+ * This file is part of wolfBoot.
```
danielinux, line 5: nit: fix copyright year

hal/stm32f1.c
```diff
@@ -0,0 +1,342 @@
...
+        if (*(uint32_t*)address != (uint32_t)~0)
+            return -1;
+
+        address += 4;
+    }
+
+    return 0;
+}
+
+int RAMFUNCTION flash_w16(volatile uint16_t *dst, const uint16_t value)
```
danielinux, line 148: please make this 'static' to avoid potential namespace collisions

hal/stm32f1.c
```diff
@@ -0,0 +1,342 @@
...
+    *dst = value;
+    DMB();
+    flash_wait_complete();
+
+    /* verify */
+    return (*dst == value) ? 0 : -1;
+}
+
+int RAMFUNCTION hal_flash_write(uint32_t address, const uint8_t *data, int len)
+{
```
danielinux, line 160: nit: please reduce the number of empty lines in this function.

danielinux (changes requested):
Thanks! The hal implementation looks much better now. I hope you don't mind fixing a few last little details.

hal/stm32f1.c
Risca: Sure!

hal/stm32f1.c
Risca: Absolutely

hal/stm32f1.c
Risca: Done!

dgarske (approved):

danielinux (approved):

danielinux:
Merged! Thanks @Risca

PR #544, opened 2025-01-30
Revert simple start-up for Raspi3b

miyazakh:
Guard UART
Added to handle NO_QNX definition


dgarske (approved):
I'll allow this through for now, but longer term I want to combine the startup assembly.

PR #560, opened 2025-03-26
 libwolfboot.a improvements / documentation updates and Renesas RX minor fixes 

dgarske:
* Support for building libwolfboot.a (Library target used to just build a test-lib, but that's not very portable).
* Added test cases for wolfboot as library
* Fix for malformed GNUC macro causing clock to not be setup properly when building with RX-GCC compiler. 
* Fix warnings with using `%zu` on older GCC.
* Fix Renesas RX TSIP build from source to use all .c files.


.github/workflows/test-library.yml
```diff
@@ -0,0 +1,38 @@
...
+        with:
+          submodules: true
+
+      - name: make clean
+        run: |
+          make keysclean && make -C tools/keytools clean && rm -f include/target.h
+
+      - name: Build test-lib
+        run: |
+          cp config/examples/library.config .config
```
danielinux, line 24: `&& make include/target.h`

danielinux (dismissed):

danielinux (approved):

PR #562, opened 2025-04-07
Scatter/gather ELF files according to segments `paddr`

danielinux:
This PR adds the option `ELF_SCATTERED`, which allows for an ELF file to be stored in different partitions on the target flash memory.
Before starting, wolfboot will check all the scattered fragments for integreity, by calculating the digest to match the one in the header.

Also added: simulator config files to test. Still unable to stage the scattered image in the simulator, but the current wolfboot.elf application shows the working scattered store/gathered digest procedures before `do_boot()`.


bigbrett (changes requested):
Great stuff so far. In addition to my review comments, did you have integration into the signing keytool that wasn't committed? You must have in order to be able to generate images right?

src/elf.c
```diff
@@ -165,11 +149,141 @@ int elf_load_image_mmu(uint8_t *image, uintptr_t *entry, elf_mmu_map_cb mmu_cb)
...
+                continue;
+            paddr = (unsigned long)ph[i].paddr;
+            offset = (unsigned long)ph[i].offset;
+            filesz = (unsigned long)ph[i].file_size;
+            printf("Writing section at address %lx offset %lx\n", paddr, offset);
+#ifdef EXT_FLASH
+            if (ext_flash) {
+                ext_flash_unlock();
+                ext_flash_erase(paddr + ARCH_FLASH_OFFSET, filesz);
+                ext_flash_write(paddr + ARCH_FLASH_OFFSET, image + offset, filesz);
```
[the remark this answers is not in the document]
danielinux: yes, possibly this was needed only because of the simulator, which has a dynamic base.

danielinux:
@bigbrett all the remarks have been addressed. I'm available to discuss further on optimizing the update time. I think it's a valid point but I'd like to prioritize design simplicity in this case. I'm open to better approaches if we find one.

I've also added a simple build-verify test case using the simulator since your last review.

bigbrett (approved):
This is great work Daniele. I wish we had a better way to actually verify the boot, but given the dynamic library shenanigans we would need to wrangle to get this running in the simulator, I think seeing the hash verify is good enough for me to get this in.

I think we should ideally figure out a way to test the actual boot in CI at some point, whether it be on renode or on QEMU. ~Theoretically we should be able to load any of the elfs curently used for testing on renode via this mechanism, and they should still boot.~ Edit: JK no we can't, since it is linked to run from the boot partition

dgarske (approved):
All tests out here! Nice work!

config/examples/sim-elf-scattered.config
```diff
@@ -0,0 +1,28 @@
+ARCH=sim
```
dgarske, line 1: Could there be build tests for these new example configs?

dgarske (changes requested):

bigbrett (changes requested):
@danielinux I just spotted something. This new feature breaks [wolfBoot_erase_bootloader()](https://github.com/danielinux/wolfBoot/blob/c841113ea7a48e07e88f23f1ae181adce504ee84/src/update_flash.c#L66-L71), which assumes the bootloader occupies all memory between `ARCH_FLASH_OFFSET` and `WOLFBOOT_PARTITION_BOOT_ADDRESS`. This is no longer an appropriate assumption when scattered elf loading is turned on, as this could be perfectly valid memory to load to. Not sure if we have a good way currently to get just the size of wolfBoot itself....would need to think about this. Thoughts? Maybe we need to introduce an additional "scatter region" or maybe just a dedicated partition region for wolfBoot itself?

Edit: At least internally, this seems to only be used by self-update. I guess that would mean the self update could potentially destroy the "active image" e.g. scattered elf, however I believe that would just cause wolfBoot to re-scatter the elf in the boot partition on the next boot?

bigbrett (changes requested):
also found a few lingering printfs

billphipps:
Couple of questions:
1. I think I'm missing some code: elf_check_image_scattered?  This is where I assumed there would be a loop of hashes against each of the individual sections.  It was unclear to me if each section would have a stored hash or if a single hash would be updated and only finalized at the end.
2. Is there a reason that scattered support isn't default within the ELF support?  Seems like an elf file that has a single segment should measure the same when scattered support is enabled.
3. I think the simulator should be able test this.  I'll take a quick swing at pinning memory maps and see if that can make static linkage work.

config/examples/sim-elf-scattered.config
danielinux: good idea. I'm adding a test now. Thanks @dgarske

danielinux:
> @danielinux I just spotted something. This new feature breaks [wolfBoot_erase_bootloader()](https://github.com/danielinux/wolfBoot/blob/c841113ea7a48e07e88f23f1ae181adce504ee84/src/update_flash.c#L66-L71), 

It would make sense to me to make the two features incompatible

bigbrett:
@danielinux @dgarske I'm going to be making some changes to this functionality as I integrate and test on TC3xx (it wont quite work written as-is at the moment). That said, since this is working in Sim, should we go ahead an merge and I will add additional changes in a subsequent PR? Or would you prefer we close this PR and I'll add additional commits on top of @danielinux 's work and create a new PR once working on TC3xx? Slight preference to just merging the work as-is, since it isn't in use by anyone yet, but up to you

dgarske (approved):

src/elf.c
```diff
@@ -165,11 +149,148 @@ int elf_load_image_mmu(uint8_t *image, uintptr_t *entry, elf_mmu_map_cb mmu_cb)
...
+                ext_flash_lock();
+            }
+            else
+#endif
+            {
+                hal_flash_unlock();
+                hal_flash_erase(paddr + BASE_OFF, filesz);
+                hal_flash_write(paddr + BASE_OFF, image + offset, filesz);
+                hal_flash_lock();
+            }
```
bigbrett, line 240: This assumes that the HAL implementation of `{ext,hal}_flash_write()` can handle reading from flash as well, which I'm not sure all of our HALs do (at least TC3xx does not). Normally when ext flash is enabled we do this via `wolfBoot_copy_sector()` which copies between flash and swap through a RAM buffer.

bigbrett (commented):
@danielinux I found a few other issues the more I look at this....do you want to fix/work on these or hand it over to me? If the latter, thoughts on closing the PR vs merging work as-is even if it doesn't work on AURIX?

src/elf.c
```diff
@@ -165,11 +149,148 @@ int elf_load_image_mmu(uint8_t *image, uintptr_t *entry, elf_mmu_map_cb mmu_cb)
...
+#   define BASE_OFF ARCH_FLASH_OFFSET
+#else
+#   define BASE_OFF 0
+#endif
+
+int elf_store_image_scattered(const unsigned char *hdr, unsigned long *entry_out, int ext_flash) {
+    const unsigned char *image;
+    int is_elf32;
+    unsigned short entry_count;
+    unsigned short entry_size;
```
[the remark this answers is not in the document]
danielinux: I fixed this one

src/image.c
```diff
@@ -1291,6 +1327,287 @@ int wolfBoot_verify_integrity(struct wolfBoot_image *img)
...
+    int stored_sha_len;
+    int i;
+    uint8_t padding_block[PADDING_BLOCK_SIZE];
+    int entry_out_set = 0;
+
+    
+    wolfBoot_hash_t ctx;
+    if (wolfBoot_open_image(&boot, part) < 0)
+        return -1;
+    p = get_img_hdr(&boot);
```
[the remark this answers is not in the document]
danielinux: Good point. If this is the case, the elf parser should be reworked to use a mechanism similar to `get_sha_block()` instead of assuming the read from mmap'd header. The local buffer storing the elf header should be parsed properly in advance (no dynamic allocation is allowed). Same occurs for the trailing at the end of the image.

Calculating the hash should also use the actual get_sha_block while parsing the chunks, instead of dereferencing 'p'

src/elf.c
danielinux: As discussed, an extra buffer may be needed to handle the writes to the same device where we read.

danielinux:
@bigbrett  I'd agree to merge as is for now, keeping in mind the limitation with non-directly-addressable origin. This is a good working point for this feature and merging in now will make further changes easier to review as well.

bigbrett (approved):

PR #570, opened 2025-05-08
wolfBoot as library portability fixes

dgarske:
* Improvements for portability using older gcc 4.8.2. 
  - Make sure wolfboot.h includes the `wolfBoot_verify_*` API's. 
  - Fixed issue with parsing headers when `#include` is `# include`.
* Don't hide `wolfBoot_open_image` and `wolfBoot_open_image_external` for library build.


include/image.h
```diff
@@ -833,7 +833,6 @@ int wolfBoot_open_image_external(struct wolfBoot_image* img, uint8_t part, uint8
 int wolfBoot_open_image_address(struct wolfBoot_image* img, uint8_t* image);
 int wolfBoot_verify_integrity(struct wolfBoot_image *img);
 int wolfBoot_verify_authenticity(struct wolfBoot_image *img);
-int wolfBoot_get_partition_state(uint8_t part, uint8_t *st);
```
danielinux, line 836: Not sure why this declaration was removed?

dgarske: Because its in `include/wolfboot/wolfboot.h:375`.

src/image.c
```diff
@@ -1500,7 +1500,7 @@ static int update_hash_flash_fwimg(wolfBoot_hash_t*       ctx,
 {
     uint32_t current_offset = offset;
     uint32_t remaining_size = size;
-    uint8_t  read_buf[WOLFBOOT_SHA_BLOCK_SIZE]; /* Use local buffer */
+    uint8_t read_buf[WOLFBOOT_SHA_BLOCK_SIZE] XALIGNED(4); /* Use local buffer*/
```
danielinux: Please use XALIGNED_STACK for aligning local variables. IAR does not like it, so we have a way to turn that off specifically for the stack.

src/image.c
```diff
@@ -1529,7 +1529,7 @@ static int update_hash_flash_fwimg(wolfBoot_hash_t*       ctx,
 static int update_hash_flash_addr(wolfBoot_hash_t* ctx, uintptr_t addr,
                                   uint32_t size, int src_ext)
 {
-    uint8_t   buffer[WOLFBOOT_SHA_BLOCK_SIZE];
+    uint8_t   buffer[WOLFBOOT_SHA_BLOCK_SIZE] XALIGNED(4);
```
danielinux: XALIGNED_STACK

src/image.c
```diff
@@ -1568,7 +1568,7 @@ int wolfBoot_check_flash_image_elf(uint8_t part, unsigned long* entry_out)
     size_t                ph_size           = 0;
     size_t                current_ph_offset = 0;
     int64_t               final_offset      = -1;
-    uint8_t               calc_digest[WOLFBOOT_SHA_DIGEST_SIZE];
+    uint8_t               calc_digest[WOLFBOOT_SHA_DIGEST_SIZE] XALIGNED(4);
```
danielinux: XALIGNED_STACK

src/update_flash.c
```diff
@@ -172,8 +172,8 @@ static int RAMFUNCTION wolfBoot_copy_sector(struct wolfBoot_image *src,
         while (pos < WOLFBOOT_SECTOR_SIZE)  {
           if (src_sector_offset + pos <
               (src->fw_size + IMAGE_HEADER_SIZE + FLASHBUFFER_SIZE)) {
-              /* bypass decryption, copy encrypted data into swap */
-              if (dst->part == PART_SWAP) {
+              /* bypass decryption, copy encrypted data into swap (and its external) */
+              if (dst->part == PART_SWAP && SWAP_EXT) {
```
dgarske, line 176: Note: This is the location for fixing `NO_SWAP_EXT=1` with encryption enabled.

src/image.c
```diff
@@ -1913,7 +1913,14 @@ int wolfBoot_verify_authenticity(struct wolfBoot_image *img)
          * TSIP encrypted key is installed at
          *    RENESAS_TSIP_INSTALLEDKEY_ADDR
          */
+        extern int hal_renesas_init(void);
```
danielinux, line 1916: Why is the tsip init happening this late, instead of in hal_init() ? There must be a reason for having to export a function for such hw-specific feature, from the hal, but I'm failing to see it

src/libwolfboot.c
```diff
@@ -1335,6 +1335,14 @@ int wolfBoot_fallback_is_possible(void)
 #ifdef EXT_ENCRYPTED
 #include "encrypt.h"
 
+#if defined(WOLFBOOT_RENESAS_TSIP)
```
danielinux, line 1338: If this is needed, can we move to the same ifdef at L.1368?

src/image.c
dgarske: Because libwolfboot.a doesn't guarantee it gets called.

src/libwolfboot.c
dgarske: I prefer these includes be outside the function.

src/image.c
```diff
@@ -1354,7 +1354,7 @@ static int copy_flash_buffered(uintptr_t src_addr, uintptr_t dst_addr,
 
 #ifndef BUFFER_DECLARED
 #define BUFFER_DECLARED
-    static uint8_t buffer[FLASHBUFFER_SIZE];
+    static uint8_t buffer[FLASHBUFFER_SIZE] XALIGNED_STACK(4);
```
danielinux: XALIGNED(4) -- not in stack

src/update_flash.c
```diff
@@ -166,14 +166,14 @@ static int RAMFUNCTION wolfBoot_copy_sector(struct wolfBoot_image *src,
     if (PART_IS_EXT(src)) {
 #ifndef BUFFER_DECLARED
 #define BUFFER_DECLARED
-        static uint8_t buffer[FLASHBUFFER_SIZE];
+        static uint8_t buffer[FLASHBUFFER_SIZE] XALIGNED_STACK(4);
```
danielinux: XALIGNED(4) -- not in stack

PR #576, opened 2025-06-20
Add support for PowerPC 32-bit SHA256 assembly speedups

dgarske:
Add support for PowerPC 32-bit SHA256 assembly speedups. 

* Minor fix for ELF loading to set entry point before copying sections (resolves issue with loading on top of current location). 
* Rename elf_load_image_mmu argument `entry` to `pentry` to avoid confusion with GET_H64 macro's use of header member "entry".
* Added some useful PowerPC timer code and examples in `hal/nxp_p1021.c`


rizlik (approved):
It looks good, my only comment is that moving the parsing of the entry point doesn't solve the general problem of memory overwrite  when loading  the ELF.

In other archs application and wolfBoot agreed on separating completely the areas where can ELF can be loaded from the memory used by wolfBoot. This way, as the ELF is verified, we can put the responsibility of not using any memory area used by wolfBoot on the ELF.

danielinux (approved):

dgarske:
The wolfHSM tests are having issues (hanging). @bigbrett is looking into and will followup with a separate PR.

src/elf.c
```diff
@@ -133,20 +136,29 @@ int elf_load_image_mmu(uint8_t *image, uintptr_t *entry, elf_mmu_map_cb mmu_cb)
...
+            if (mem_size > file_size) {
+                memset((void*)(uintptr_t)(vaddr + file_size), 0,
+                    mem_size - file_size);
+            }
+        #ifdef ARCH_PPC
+            flush_cache(paddr, mem_size);
+        #endif
+        }
+    #ifdef DEBUG_ELF
+        else {
```
dgarske, line 153: Example output during testing of "skip":
```
Firmware Valid
Loading elf at 0x1E000000
Found valid elf32 (big endian)
Program Headers 2 (size 32)
Load 584 bytes (offset 0x10000) to 0x1E000000 (p 0x1E000000)
Section would collide with headers! Skipping
Entry point 0x1E000000
Booting at 0x1E000000
```

Example working:
```
Firmware Valid
Loading elf at 0x1D000000
Found valid elf32 (big endian)
Program Headers 2 (size 32)
Load 584 bytes (offset 0x10000) to 0x1E000000 (p 0x1E000000)
Entry point 0x1E000000
Booting at 0x1E000000
Test App

0x00000001
0x00000002
0x00000003
```

PR #578, opened 2025-06-25
Fix for documented option bytes on STM32H5 with TrustZone enabled

dgarske:
Fix for documented option bytes on STM32H5 with TrustZone enabled. Flash bank 2 watermark (SECWM2_STRT=0x0 and SECWM2_END=0x1F). ZD 20037


danielinux (approved):
It seems OK to suggest to set SECWM2 range even when DUALBANK is off (`stm32h5-tz.config`). In the case with DUALBANK off the option should be ignored as there is no BANK 2.

PR #580, opened 2025-07-02
Use correct VTOR_NS location when TrustZone enabled

mattia-moffa:
Fixes a regression introduced in [6681e54](https://github.com/wolfSSL/wolfBoot/commit/6681e54) which sets `VTOR` to `0xE000ED08` (`VTOR`) rather than `0xE002ED08` (`VTOR_NS`) when TrustZone is enabled. Issue found by Daniele while I was debugging wolfBoot on the STM32H5 for ZD#20037. Will ask them to verify whether this fixes their first problem.

### DIFF
--- a/docs/STM32-TZ.md
+++ b/docs/STM32-TZ.md
@@ -48,6 +48,13 @@ bytes (256×4). As a result, in this case `IMAGE_HEADER_SIZE` must be set to
 This detail is already taken care of in the configuration files provided in
 `config/examples`.
 
+In addition to this, when using the signing tool standalone the appropriate
+image header size must be supplied as an environment variable. For example:
+
+```
+IMAGE_HEADER_SIZE=1024 ./tools/keytools/sign --sha256 --ecc256 myapp.bin wolfboot_signing_private_key.der 1
+```
+
 ### Example using STM32L552
 
   - Copy the example configuration for STM32-L5 with support for wolfCrypt in


src/boot_arm.c
```diff
@@ -416,8 +416,7 @@ void isr_empty(void)
 
 #ifdef TZEN
 #include "hal.h"
-//#define VTOR (*(volatile uint32_t *)(0xE002ED08))
-#define VTOR (*(volatile uint32_t *)(0xE000ED08))
+#define VTOR (*(volatile uint32_t *)(0xE002ED08))
```
dgarske, line 419: I will test here right now.

Can we add some code to allow a custom VTOR like this?

```
#ifdef VTOR_ADDR
#define VTOR (*(volatile uint32_t *)(VTOR_ADDR))
#else
...
#endif
```

dgarske (changes requested):

src/boot_arm.c
dgarske: That's much better:

```
STM32H5 wolfBoot demo Application
Copyright 2024 wolfSSL Inc
GPL v3
Version : 0x1
========================

cmd> 
```

However I am not able to send any commands. Not sure if something else is broken on the H5 UART RX side.

mattia-moffa: I'm not able to send commands either, that's their second issue. I tried checking out wolfBoot v2.2.0 and this seems to occur there too.

danielinux: I don't think a custom VTOR is needed, when TZEN=1 the VTOR_NS should always be set. That one was a blunder on my side when porting to the rp2350.

mattia-moffa:
This last commit restores the functionality of the UART console on my side. The issue was a misalignment of the interrupt vector table of the boot image. According to the [ARM Cortex-M33 documentation](https://developer.arm.com/documentation/100235/0004/the-cortex-m33-processor/exception-model/vector-table):

> The silicon vendor must configure the required alignment of the vector tables, which depends on the number of interrupts implemented. The minimum alignment is 32 words, enough for up to 16 interrupts. For more interrupts, adjust the alignment by rounding up to the next power of two. For example, if you require 21 interrupts, the alignment must be on a 64-word boundary because the required table size is 37 words, and the next power of two is 64.

The STM32H5 has 149 (133 + 16) interrupts; the next power of two is 256 words, i.e. 1024 bytes. The default image header size was previously 256 bytes, causing the interrupt vector to be 256-aligned and the ISR used for UART to be offset in respect to the correct one.

config/examples/stm32h5-tz.config
```diff
@@ -27,3 +27,4 @@ FLAGS_HOME=0
 DISABLE_BACKUP=0
 WOLFCRYPT_TZ=1
 WOLFCRYPT_TZ_PKCS11=1
+IMAGE_HEADER_SIZE?=1024
```
dgarske, line 30: @mattia-moffa please update the documentation for docs/STM32-TZ.md to include the details from https://github.com/wolfSSL/wolfBoot/pull/580#issuecomment-3046200291

dgarske (changes requested):

config/examples/stm32h5-tz.config
mattia-moffa: Done, do you think that's appropriate?

docs/STM32-TZ.md
```diff
@@ -27,6 +27,27 @@ non-secure domain can access wolfCrypt through a standard PKCS11 interface and
...
+> adjust the alignment by rounding up to the next power of two. For example, if
+> you require 21 interrupts, the alignment must be on a 64-word boundary
+> because the required table size is 37 words, and the next power of two is 64.
+
+For example, all the STM32H5 series boards have at least 146 interrupt
+channels; since the next power of two is 256, they require an alignment of 1024
+bytes (256×4). As a result, in this case `IMAGE_HEADER_SIZE` must be set to
+`1024` or a multiple of it.
+
+This detail is already taken care of in the configuration files provided in
```
dgarske, line 48: You might add a note about IMAGE_HEADER_SIZE using the signing tool. When running standalone you must provide the IMAGE_HEADER_SIZE as an environment variable: Example: `IMAGE_HEADER_SIZE=1024 ./tools/keytools/sign --sha256 --ecc256 myapp.bin wolfboot_signing_private_key.der 1`

dgarske (changes requested):
Looks great otherwise.

danielinux (approved):

PR #581, opened 2025-07-08
Fixed unaligned access report by user (Cortex-A5)

danielinux:
As reported in ZD 20152, this flag is needed to prevent unaligned access when running integrity check.


danielinux:
@dgarske do you think it's worth considering adding this flag for other Cortex-A CPUs, (e.g. 64bit) as well?

dgarske (approved):

Issue #586, opened 2025-08-05
CMake STM32H7 Example is Broken


kwalker-plxs:
The example [here](https://github.com/wolfSSL/wolfBoot?tab=readme-ov-file#cmake) for the stm32h7 does not compile as it stands. 

It looks to be an issue with the `isr_tim2` usage in startup_arm.c:
```
vscode ➜ /workspaces/stm32-cpp-project-template/wolfBoot/build (h757) $ make
[  2%] Building C object CMakeFiles/wolfboothal.dir/hal/stm32h7.c.o
[  4%] Linking C static library libwolfboothal.a
[  4%] Built target wolfboothal
[  7%] Building signing tool
[  9%] Building keygen tool
[  9%] Built target keytools
[ 11%] Generating keystore.c and signing private key
Keystore size: 2608
Keytype: ECC256
Generating key (type: ECC256)
Associated key file:   /workspaces/stm32-cpp-project-template/wolfBoot/build/wolfboot_signing_private_key.der
Partition ids mask:   ffffffff
Key type   :           ECC256
Public key slot:       0
Done.
[ 14%] Building C object CMakeFiles/public_key.dir/keystore.c.o
[ 16%] Linking C static library libpublic_key.a
[ 19%] Built target public_key
[ 21%] Building C object lib/CMakeFiles/wolfcrypt.dir/wolfssl/wolfcrypt/src/integer.c.o
[ 23%] Building C object lib/CMakeFiles/wolfcrypt.dir/wolfssl/wolfcrypt/src/tfm.c.o
[ 26%] Building C object lib/CMakeFiles/wolfcrypt.dir/wolfssl/wolfcrypt/src/ecc.c.o
[ 28%] Building C object lib/CMakeFiles/wolfcrypt.dir/wolfssl/wolfcrypt/src/memory.c.o
[ 30%] Building C object lib/CMakeFiles/wolfcrypt.dir/wolfssl/wolfcrypt/src/wc_port.c.o
[ 33%] Building C object lib/CMakeFiles/wolfcrypt.dir/wolfssl/wolfcrypt/src/wolfmath.c.o
[ 35%] Building C object lib/CMakeFiles/wolfcrypt.dir/wolfssl/wolfcrypt/src/hash.c.o
[ 38%] Building C object lib/CMakeFiles/wolfcrypt.dir/wolfssl/wolfcrypt/src/sha256.c.o
[ 40%] Linking C static library libwolfcrypt.a
[...]

PR #587, opened 2025-08-07
Fixed build with Cmake on STM32H7 with TEST_APPS.

danielinux:
As reported in #586


Looking at the code, I notice that in the STM32H7 section, there's a line
kareem-wolfssl, line 1: What is this file supposed to be?  Is this intentional?

danielinux: Lol thanks for spotting this. Clearly accidental

dgarske (approved):

Issue #589, opened 2025-09-05
Last sector of the firmware in boot partition gets corrupted in `wolfBoot_swap_and_final_erase`


reza-hdd:
Scenario:
Boot partition: INTERNAL
Update partition: EXTERNAL
Swap Partition: EXTERNAL

When Firmware is large enough to consume all the allocated boot partition sectors, in the the final swap and erase operations (_wolfBoot_swap_and_final_erase_) the last sector of FW in Boot partition, which is used as a temporary sector for a copy operation, gets corrupted after being copied in external Swap partition and copied back. It looks like the contents are copied to the external Swap partition without being encrypted, but get decrypted when it is read back.

dgarske:
Hi @reza-hdd ,

Thank you for the report. We will investigate. Can you tell us the hardware platform and share your .config?

We use the last sector for storing state. Are you using `NVM_FLASH_WRITEONCE=1`? If your onboard flash does not support erase and multiple writes then we must use 2 sectors.

Also if you have another place you'd like to store state you can override the functions using `CFLAGS_EXTRA+="-DCUSTOM_PARTITION_TRAILER"` and supply your own implementation for the trailer state. We do have some examples for this to share if you are interested.

```
/* Custom partition trailer
 * Function implementation externally defined
 */
uint8_t* RAMFUNCTION get_trailer_at(uint8_t part, uint32_t at);
void     RAMFUNCTION set_trailer_at(uint8_t part, uint32_t at, uint8_t val);
void     RAMFUNCTION set_partition_magic(uint8_t part);
```

Thanks,
David Garske, wolfSSL

reza-hdd:
Thanks Daniele for the quick fix. I'll test as soon as I can and will let you know the result.
Regards,
Reza

PR #592, opened 2025-09-10
Fix corruption of last boot sector with encryption

danielinux:
As reported by @reza-hdd in #589:

> When Firmware is large enough to consume all the allocated boot partition sectors, in the the final swap and erase operations (_wolfBoot_swap_and_final_erase_) the last sector of FW in Boot partition, which is used as a temporary sector for a copy operation, gets corrupted after being copied in external Swap partition and copied back. It looks like the contents are copied to the external Swap partition without being encrypted, but get decrypted when it is read back.

This was due to the function wolfBoot_copy_sector() assuming that all copies having an external SWAP as destination would not need any encryption, as the "normal" case during update is to copy already encrypted sectors from UPDATE->SWAP.

In the final state, after the update is successfully applied, the mechanism saves a copy of the last sector from BOOT->SWAP. In this case, if the application is big enough, an extra function is needed to encrypt the content of this last sector before writing it to SWAP.

Restoring the backup is OK (normal case SWAP->BOOT, decryption was already there as also noted by the reporter).

### Testing

I was able to reproduce the issue in the simulator by applying the following patch to the test application:

```
diff --git a/test-app/app_sim.c b/test-app/app_sim.c
index e118e5a9..7446b763 100644
--- a/test-app/app_sim.c
+++ b/test-app/app_sim.c
@@ -41,6 +41,9 @@
 char enc_key[] = "0123456789abcdef0123456789abcdef"
                 "0123456789abcdef";
 
+#define FILLER_SIZE (0x16000 - 0x2000 - 0x400)
+static volatile uint8_t filler_data[FILLER_SIZE] = { 0x01, 0x02, 0x03 };
+
 #ifdef TEST_DELTA_DATA
 static volatile char __attribute__((used)) garbage[TEST_DELTA_DATA] = {0x01, 0x02, 0x03, 0x04 };
 
@@ -50,6 +53,7 @@ void hal_init(void);
 
 int do_cmd(const char *cmd)
 {
+    filler_data[FILLER_SIZE - 1] = 0xAA;
     if (strcmp(cmd, "powerfail") == 0) {
         return 1;
     }
```

FILLER_SIZE is arbitrary, and adjusted on my setup to produce a test image.bin as big to fill up to the third-last sector (sector size in sim= 0x1000, two sectors reserved for redundancy of `NVM_FLASH_WRITEONCE` workaround, 0x200 or 0x400 for manifest header).

I used the following configuration:

```
cp config/examples/sim-encrypt-nvm-writeonce-update.config .config
```

and built the test case for the update:

```
test-sim-external-flash-with-enc-update
```

Running the update test:

```
tools/scripts/sim-sunnyday-update.sh
```

Would fail on current master branch, as reported. With this fix the test is successful, and the last sector is encrypted during the backup step.

### DIFF
--- a/src/update_flash.c
+++ b/src/update_flash.c
@@ -28,6 +28,7 @@
 #include "hal.h"
 #include "spi_flash.h"
 #include "wolfboot/wolfboot.h"
+#include "target.h"
 
 #include "delta.h"
 #include "printf.h"
@@ -152,7 +153,11 @@ static int RAMFUNCTION wolfBoot_copy_sector(struct wolfBoot_image *src,
         dst_sector_offset = 0;
 
 #ifdef EXT_ENCRYPTED
+    if (wolfBoot_initialize_encryption() < 0) {
+        return -1;
+    }
     wolfBoot_get_encrypt_key(key, nonce);
+
     if (src->part == PART_SWAP)
         iv_counter = dst_sector_offset;
     else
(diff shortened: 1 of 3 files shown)


src/update_flash.c
```diff
@@ -202,6 +207,56 @@ static int RAMFUNCTION wolfBoot_copy_sector(struct wolfBoot_image *src,
...
+    iv_counter = src_sector_offset;
+    iv_counter /= ENCRYPT_BLOCK_SIZE;
+    if (wolfBoot_initialize_encryption() < 0)
+        return -1;
+    crypto_set_iv(nonce, iv_counter);
+
+    /* Erase swap space */
+    wb_flash_erase(dst, dst_sector_offset, WOLFBOOT_SECTOR_SIZE);
+    if (PART_IS_EXT(dst)) {
+        uint32_t sz = 0;
```
rizlik: unused var

src/update_flash.c
```diff
@@ -202,6 +207,56 @@ static int RAMFUNCTION wolfBoot_copy_sector(struct wolfBoot_image *src,
...
+        uint32_t step = 0;
```
rizlik: unused var

src/update_flash.c
```diff
@@ -202,6 +207,56 @@ static int RAMFUNCTION wolfBoot_copy_sector(struct wolfBoot_image *src,
...
+        uint8_t *orig = (uint8_t *)(WOLFBOOT_PARTITION_BOOT_ADDRESS) + src_sector_offset; 
+        while (pos < WOLFBOOT_SECTOR_SIZE) {
+            uint32_t len = ENCRYPT_BLOCK_SIZE;
```
rizlik: unused var

rizlik: remove len as well

PR #596, opened 2025-09-25
Refactor Makefiles to allow overriding library  paths

bigbrett:
Allows for the user to build wolfBoot against a different version of internal library dependencies (`wolfHSM`, `wolfTPM`, etc) if they want to live dangerously.

The default behavior remains the same (points to `lib/*`)

### DIFF
--- a/docs/compile.md
+++ b/docs/compile.md
@@ -343,3 +343,17 @@ LC_ALL=
 
 Then run the normal `make` steps.
 
+### Building Against Alternate Library Dependency Versions
+
+wolfBoot includes its internal dependencies (all official wolfSSL projects) as git submodules under `lib/`, ensuring known-compatible versions. You may override these paths with `WOLFBOOT_LIB_XXX` environment variables to point to local copies of the libraries.
+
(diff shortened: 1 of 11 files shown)


dgarske (dismissed):

docs/compile.md
```diff
@@ -343,3 +343,17 @@ LC_ALL=
 
 Then run the normal `make` steps.
 
+### Building Against Alternate Library Dependency Versions
+
+wolfBoot includes its internal dependencies (all official wolfSSL projects) as git submodules under `lib/`, ensuring known-compatible versions. You may override these paths with `WOLFBOOT_LIB_XXX` environment variables to point to local copies of the libraries.
+
+This mechanism exists mainly for internal use and is NOT RECOMMENDED, as mismatched versions can cause build failures or subtle bugs.
```
danielinux: Remove comment at L350

danielinux (changes requested):
Remove excessive warnings from docs/compile.md. Submodule tracking is still the preferred mechanism to link dependencies, and each release is tracked (and tested) against one specific commit of each repository.

docs/compile.md
bigbrett: removed

PR #601, opened 2025-10-06
Fixes for STM32H5 TrustZone and PKCS11 store

dgarske:


hal/stm32_tz.c
```diff
@@ -288,7 +291,7 @@ void hal_tz_sau_init(void)
     /* SAU is set up before staging. Set up all areas as secure. */
 
     /* Non-secure callable: NSC functions area */
-    sau_init_region(0, 0x0C038000, 0x0C040000, 1);
+    sau_init_region(0, 0x0C040000, 0x0C05FFFF, 1);
```
dgarske, line 294: This was an important change to prevent the SG instruction hard fault. This correctly specifies the location for the FLASH_NSC. We can probably improve things to have this come from the linker script.

PR #598, opened 2025-10-06
CMake Improvements

gojimmypi:
### New Features

- New VSCode  IDE Support 
- New [CMakePresets.json](https://github.com/gojimmypi/wolfBoot/blob/pr-cmake-improvements/CMakePresets.json) Proof of concept for a handful of STM32 devices.
- New  [.config cmake reader, see cmake function: `load_dot_config(path)`](https://github.com/gojimmypi/wolfBoot/blob/pr-cmake-improvements/cmake/load_dot_config.cmake)
- New  [.config to cmake presets conversion script](https://github.com/gojimmypi/wolfBoot/blob/pr-cmake-improvements/config2presets.py).
- New  [CMake wolfbuild script](https://github.com/gojimmypi/wolfBoot/blob/pr-cmake-improvements/wolfbuild.sh) to build, clean, flash test-app (working on L4)
- Improved Windows cmake builds, better detection of MSVC
- Visual Studio although [that PR #623](https://github.com/wolfSSL/wolfBoot/pull/623) has been given a lower priority to get things done for the upcoming release.
- Expanded GitHub workflow testing, new testing scripts.
 
### Known issues

- CMake full sim testing is disabled (reference only; the default "no config" setting is the simulation. Applies only to GUI IDE)
- CMake imx_rt testing is disabled (no workflow toolchain)
- CMake presets `stm32h5` and `stm32l5` and `stm32wb` have `"BUILD_TEST_APPS": "OFF"` (toolchain / include file issue)

<br />

---

# CMake reading .config files

This PR adds support for configuring the CMake build directly from a Kconfig-style `.config` file. When a `.config` is present at the repository root and `USE_DOT_CONFIG=ON` is set, CMake reads build options from that file instead of requiring them to be passed manually on the command line.

Example usage:

```
# From the repo root, with .config already in place
cmake -S . -B build-stm32h7 \
    -DUSE_DOT_CONFIG=ON \
    -DWOLFBOOT_TARGET=stm32h7

cmake --build build-stm32h7
```

For convenience, `tools/scripts/cmake_dot_config.sh` demonstrates one way to copy `config/examples/<target>.config` to `.config` and then run the CMake configure step using this mechanism, but it is just an example wrapper around the CMake interface shown above.

# Test Drive

Work remains on CMake files, but the initial VS Code build should be working:

```bash
https://github.com/gojimmypi/wolfBoot.git
cd wolfboot
git checkout pr-cmake-improvements
git pull --recurse-submodules
git submodule update --init --recursive

[...]

PR #602, opened 2025-10-07
CMake: Use find_program instead of `which` to find gcc toolchain

cvanbeek13:
This updates the CMake configuration to use the CMake [`find_program()`](https://cmake.org/cmake/help/latest/command/find_program.html) function to locate the toolchain `gcc` application.  The existing implementation uses `which`, which is only available on Linux.  `find_program()` works cross-platform.  By setting `HINTS` to `$ENV{PATH}`, CMake looks for the program using the `PATH` environment variable, in the same way `which` would on Linux.

### DIFF
--- a/cmake/toolchain_aarch64-none-elf.cmake
+++ b/cmake/toolchain_aarch64-none-elf.cmake
@@ -35,10 +35,10 @@ set(MARCH_FLAGS "-march=armv8-a")
 set(TOOLCHAIN aarch64-none-elf)
 set(CMAKE_CXX_STANDARD 20)
 
-execute_process(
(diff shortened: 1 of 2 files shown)


gojimmypi:
Hi @cvanbeek13 -

I have some cmake improvements, including alternate toolchain locations - in https://github.com/wolfSSL/wolfBoot/pull/598 (very much WIP ATM, suggestions welcome).

If you'd like me to test and put up a PR so you don't need to fuss with the contributor agreement, I'm happy to do that. 

Otherwise, it's always nice to have new contributors.

Cheers

PR #604, opened 2025-10-08
TC3xx: Replace IDE/AURIX project with new HAL module

bigbrett:
## Overview

Major overhaul of AURIX TC3xx support to align with the new internal TC3 low-level driver library.

* New HAL based on the `wolfLLD-tc3xx` driver library (currently NDA-restricted).
* Supports both TriCore and HSM cores.
* Removed dependency on Infineon SDK and vendor IDE projects — AURIX IDE builds are no longer supported.
* Fully integrated into the wolfBoot build system — eliminates the need for the legacy `wbaurixtool.sh` script. Signing, NVM image generation, and certificate creation are now handled directly by wolfBoot Makefiles.
* Updated to support the latest upstream wolfHSM.

## Implications

* **NDA dependency:**
  `wolfLLD-tc3xx` currently includes both HSM and TriCore drivers, requiring NDA access even for TriCore-only builds. All existing customers using wolfBoot on TC3xx already meet this requirement. A public TriCore-only version of `wolfLLD-tc3xx` is planned to remove this barrier in the future.

* **Build integration:**
  TC3xx wolfBoot should currently be built **as part of** the `wolfHSM-infineon-tc3xx` platform release, which includes wolfBoot, NDA components, and a top-level Makefile for easy platform builds.

* **Documentation:**
  Build instructions for TC3xx are now maintained as part of the `wolfHSM-infineon-tc3xx` platform release due to tight coupling between components.

## Note

This PR adds support for the HSM core but contains no non-public material. It is safe to open-source.

### DIFF
--- a/hal/tc3_cfg.h
+++ b/hal/tc3_cfg.h
@@ -1,5 +1,22 @@
-/* tc3_cfg.h */
-
+/* tc3_cfg.h
+ *
+ * Copyright (C) 2014-2025 wolfSSL Inc.
+ *
+ * This file is part of wolfBoot.
+ *
+ * wolfBoot is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * wolfBoot is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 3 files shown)


test-app/app_aurix_tc3xx.c
```diff
@@ -0,0 +1,99 @@
+/* app_aurix_tc3xx_wolf.c */
```
dgarske: These new files all need the standard wolfBoot license header. Check the name here _wolf needs removed.

hal/tc3_cfg.h
```diff
@@ -0,0 +1,16 @@
+/* tc3_cfg.h */
```
dgarske: All new files need wolfBoot license header. Thanks

bigbrett: fixed

test-app/app_aurix_tc3xx.c
bigbrett: fixed

PR #608, opened 2025-10-14
Implement filesystem-based partition state management

julek-wolfssl:
- Add new logic to `libwolfboot.c` for reading and updating partition states using a filesystem, enabled via `WOLFBOOT_PARTITION_FILENAME`.
- Implement functions to set and get partition states (`wolfBoot_update_trigger`, `wolfBoot_success`, `wolfBoot_get_partition_state`)
- Add `boot_status.py` and `hal/library_fs.c` for CLI and library access.
- Update `.gitignore` and build rules for new artifacts.

### DIFF
--- a/config/examples/library_fs.config
+++ b/config/examples/library_fs.config
@@ -12,18 +12,18 @@ SPMATHALL?=0
 NO_LOADER=1
 USE_GCC_HEADLESS=0
 
-# Required to build keytools
-WOLFBOOT_SECTOR_SIZE=0x1000
-
 # Flash Partition Filename
 WOLFBOOT_PARTITION_FILENAME=\"/dev/mtd0\"
+
 # Flash Sector Size
 WOLFBOOT_SECTOR_SIZE=0x2000
 # Application Partition Size
 WOLFBOOT_PARTITION_SIZE=0xA000
-# Location in flash for application
-WOLFBOOT_PARTITION_UPDATE_ADDRESS?=0x8000
 # Location in flash for wolfBoot
+WOLFBOOT_ORIGIN=0x0
+# Location in flash for boot partition
 WOLFBOOT_PARTITION_BOOT_ADDRESS?=0x10000
+# Location in flash for update partition
+WOLFBOOT_PARTITION_UPDATE_ADDRESS?=0x8000
 # Location in flash for swap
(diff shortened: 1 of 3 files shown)


hal/library_fs.c
```diff
@@ -0,0 +1,172 @@
+/* library_fs.c
```
dgarske, line 1: Where is this file built able to be used? Docs? Also this might be better outside the hal directory under tools perhaps? This should be the top level example that leverages the libwolfboot.a.

julek-wolfssl: Added docs. I placed it next to `library.c`. This looks to be the correct path to put it in.

julek-wolfssl:
> Not related, but on 32-bit targets getting these printf warnings. Can you fix these too?

Added explicit casting as the types of those fields are dependent on the build configuration.

hal/library_fs.c
dgarske: It can stay here if needed, but you do need to wrap with a macro like `WOLFBOOT_PARTITION_FILENAME` since some people try and include all wolfBoot .c files...
My preference is to put outside the library in tools and have a separate makefile to build using the libwolfboot.a. @danielinux thoughts?

julek-wolfssl: I'm not sure that customers can do that in the `hal` directory. Other source files are not wrapped and contain platform specific code that would have surely thrown errors if they tried.

Makefile
```diff
@@ -161,7 +161,11 @@ endif
 
 ifeq ($(TARGET),library)
     CFLAGS+=-g
-    MAIN_TARGET:=libwolfboot.a
+    MAIN_TARGET:=test-lib
+endif
+
+ifeq ($(TARGET),library_fs)
+    MAIN_TARGET:=lib-fs
 endif
```
julek-wolfssl, line 169: I'm not sure if this is right since it will build the executables given the target name. If a user wants to build wolfBoot as a library without the executables then it should only output `libwolfboot.a`. Am I right that in both cases this should be `MAIN_TARGET:=libwolfboot.a`?

julek-wolfssl: @dgarske @danielinux

julek-wolfssl: I reverted this.

dgarske:
Closing in favor of https://github.com/wolfSSL/wolfBoot/pull/615

PR #617, opened 2025-10-23
Fix STM32H5 update

mattia-moffa:
Fix for STM32H5 update. ZD 20037

### DIFF
--- a/docs/STM32-TZ.md
+++ b/docs/STM32-TZ.md
@@ -59,7 +59,7 @@ IMAGE_HEADER_SIZE=1024 ./tools/keytools/sign --sha256 --ecc256 myapp.bin wolfboo
 
 wolfBoot provides a few Non-Secure Callable functions to allow a non-secure
 application to perform certain operations that must be run from the secure
-domain. For more information, see [API](docs/API.md#nsc-api).
+domain. For more information, see [API](API.md#nsc-api).
 
 ### Example using STM32L552
 


hal/stm32_tz.c
```diff
@@ -129,13 +131,14 @@ void hal_tz_claim_nonsecure_area(uint32_t address, int len)
...
     address = start_address;
+    page_n = start_page_n;
     while (address < end) {
         /* Erase claimed non-secure page, in secure mode */
 #ifndef TARGET_stm32h5
         reg = FLASH_CR & (~((FLASH_CR_PNB_MASK << FLASH_CR_PNB_SHIFT) | FLASH_CR_PER | FLASH_CR_BKER | FLASH_CR_PG | FLASH_CR_MER1 | FLASH_CR_MER2));
         FLASH_CR = reg | ((page_n << FLASH_CR_PNB_SHIFT) | FLASH_CR_PER);
 #else
-        reg = FLASH_CR & (~((FLASH_CR_PNB_MASK << FLASH_CR_PNB_SHIFT) | FLASH_CR_SER | FLASH_CR_BER | FLASH_CR_PG | FLASH_CR_MER));
+        reg = FLASH_CR & (~((FLASH_CR_PNB_MASK << FLASH_CR_PNB_SHIFT) | FLASH_CR_SER | FLASH_CR_BER | FLASH_CR_PG | FLASH_CR_MER | (1 << 31)));
```
dgarske: Use `FLASH_CR_BKSEL` instead of `(1 << 31)`.

hal/stm32_tz.c
```diff
@@ -162,8 +165,10 @@ void hal_tz_release_nonsecure_area(void)
 {
 #ifndef DUALBANK_SWAP
     int i;
-    for (i = 0; i < FLASH_SECBB_NREGS; i++)
+    for (i = 0; i < FLASH_SECBB_NREGS; i++) {
+        FLASH_SECBB1[i] = 0;
```
dgarske, line 170: Code comment. `Set all banks as non-secure`

dgarske (changes requested):
I cannot get the update to work with DUALBANK 0 or 1. Can you share your steps to reproduce? I tried with trustzone and update would not take even. Without trustzone I couldn't get it to boot.

I also have had trouble getting XMODEM to work with mincom or CoolTerm. It never seems to finish. Would be curious how you are testing.

I even added a command to call `wolfBoot_update_trigger();` manually after flashing a new version to the update partition. But it won't install.

I also tried using the "fake" method with:

```
./tools/keytools/sign --ecc256 test-app/image.bin wolfboot_signing_private_key.der 2
echo -n "pBOOT" > trigger_magic.bin
./tools/bin-assemble/bin-assemble \
  update.bin \
    0x0     test-app/image_v2_signed.bin \
    0x9DFFB trigger_magic.bin
STM32_Programmer_CLI -c port=swd -d update.bin 0x08100000
```

mattia-moffa:
I'm using the default `config/examples/stm32h5-tz.config`.

I'm generating the images with:

```
make # or make DEBUG=1
IMAGE_HEADER_SIZE=1024 tools/keytools/sign --ecc256 --sha256 test-app/image.bin wolfboot_signing_private_key.der 2
```

(notice the `IMAGE_HEADER_SIZE`, that's important).

then flashing:

- `wolfboot.bin` to `0x0C000000`
- `test-app/image_v1_signed.bin` to `0x08060000`
- `test-app/image_v2_signed.bin` to `0x08100000`

then writing `pBOOT` to the end of the update partition. Since the partition size is `0xA0000`, that would be at offset `0x9FFFB` in the partition, i.e. address `0x0819FFFB`. You're using `0x9DFFB` which I don't think is the right address.

It also works if I update via XMODEM in the app, but that's a bit unreliable on minicom on my end for whatever reason, I have to use CoolTerm.

docs/Targets.md
```diff
@@ -975,7 +975,7 @@ To initiate an update, sign a new version of the app and upload the v3 to the up
 on the second bank:
 
 ```sh
-tools/keytools/sign --ecc256 test-app/image.bin wolfboot_signing_private_key.der 3
+IMAGE_HEADER_SIZE=1024 tools/keytools/sign --ecc256 test-app/image.bin wolfboot_signing_private_key.der 3
```
dgarske, line 991: Put `IMAGE_HEADER_SIZE` into the default .config for all H5's since it is important.

dgarske (changes requested):

docs/Targets.md
dgarske: Oh it is already, its just the signing tool step. I see. Thank you

dgarske (dismissed):

dgarske (approved):
Approving and merging as is. This PR actually makes the update partition marked secure flash and uses the NCS callable interface to program the update partition, which is very nice. There is a known issue with Dual bank Mattia is working on and will followup with a new PR.

PR #619, opened 2025-10-30
Fix STM32H5 dualbank update

mattia-moffa:
Also adds a `stm32h5-tz-dualbank.config` which is the same as `stm32h5-tz-dualbank-otp.config` but without OTP.

### DIFF
--- a/.github/workflows/test-configs.yml
+++ b/.github/workflows/test-configs.yml
@@ -320,6 +320,12 @@ jobs:
       arch: arm
       config-file: ./config/examples/stm32h5-tz.config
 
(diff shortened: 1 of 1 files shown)


config/examples/stm32h5-tz-dualbank.config
```diff
@@ -0,0 +1,31 @@
+ARCH?=ARM
```
dgarske, line 1: Can you please add the new configs to the build test `.github/workflows/test-configs.yml`?

dgarske (changes requested):

dgarske (approved):

PR #622, opened 2025-10-31
Added SPI TPM support for STM32H5

dgarske:
Added STM32H5 SPI Driver and support for TPM
Cleanup STM32H5 HAL port.
Refactor `__ARM_FEATURE_CMSE` to `TZ_SECURE()`
Refactor attribute cmse_nonsecure_entry to use new macro `CSME_NSE_API`
Added TPM Non Secure Callable API's
Added TPM support to STM32H5 test application

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,7 +105,7 @@ CFLAGS+= \
 # Setup default optimizations (for GCC)
 ifeq ($(USE_GCC_HEADLESS),1)
   CFLAGS+=-Wall -Wextra -Wno-main -ffreestanding -Wno-unused -nostartfiles
-  CFLAGS+=-ffunction-sections -fdata-sections -fomit-frame-pointer -Wno-unused-variable
+  CFLAGS+=-ffunction-sections -fdata-sections -fomit-frame-pointer
   LDFLAGS+=-Wl,-gc-sections -Wl,-Map=wolfboot.map -ffreestanding -nostartfiles
   # Not setting LDFLAGS directly since it is passed to the test-app
   LSCRIPT_FLAGS+=-T $(LSCRIPT)
@@ -531,7 +531,6 @@ cppcheck:
 	cppcheck -f --enable=warning --enable=portability \
 		--suppress="ctunullpointer" --suppress="nullPointer" \
 		--suppress="objectIndex" --suppress="comparePointers" \
-		--suppress="subtractPointers" --suppress="intToPointerCast" \
 		--check-level=exhaustive \
 		--error-exitcode=89 --std=c89 src/*.c hal/*.c hal/spi/*.c hal/uart/*.c
 


Makefile
```diff
@@ -105,7 +105,7 @@ CFLAGS+= \
 # Setup default optimizations (for GCC)
 ifeq ($(USE_GCC_HEADLESS),1)
   CFLAGS+=-Wall -Wextra -Wno-main -ffreestanding -Wno-unused -nostartfiles
-  CFLAGS+=-ffunction-sections -fdata-sections -fomit-frame-pointer
+  CFLAGS+=-ffunction-sections -fdata-sections -fomit-frame-pointer -Wno-unused-variable
```
danielinux: I would keep the warning if possible?

Makefile
```diff
@@ -531,6 +531,8 @@ cppcheck:
 	cppcheck -f --enable=warning --enable=portability \
 		--suppress="ctunullpointer" --suppress="nullPointer" \
 		--suppress="objectIndex" --suppress="comparePointers" \
+		--suppress="subtractPointers" --suppress="intToPointerCast" \
+		--check-level=exhaustive \
```
danielinux: OK fro the subtractPointers, but I'd keep cast warnings and fix casts instead

Makefile
dgarske: Reverted. I was trying to warn on unused. Switched to `-Wunused-variable`

Makefile
dgarske: I can't get a clean build with `make cppcheck` locally on Ubuntu 24 without this. But since CI isn't throwing this error, I will revert.

```
src/update_ram.c:179:28: portability: Casting non-zero decimal integer literal to pointer. [intToPointerCast]
            load_address = (uint32_t*)WOLFBOOT_LOAD_ADDRESS;
```

Makefile
dgarske: For now I'll leave the `-Wunused-variable` since we even have cases in wolfCrypt where this is happening.

```
/home/davidgarske/GitHub/wolfboot/lib/wolfssl/wolfcrypt/src/asn.c:26470:19: error: 'END_DSA_PRIV' defined but not used [-Werror=unused-const-variable=]
26470 |     static wcchar END_DSA_PRIV     = "-----END DSA PRIVATE KEY-----";
      |                   ^~~~~~~~~~~~
/home/davidgarske/GitHub/wolfboot/lib/wolfssl/wolfcrypt/src/asn.c:26469:19: error: 'BEGIN_DSA_PRIV' defined but not used [-Werror=unused-const-variable=]
26469 |     static wcchar BEGIN_DSA_PRIV   = "-----BEGIN DSA PRIVATE KEY-----";
      |                   ^~~~~~~~~~~~~~
cc1: all warnings being treated as errors
```

I'll put up a PR in wolfSSL for that, but first making sure there are not others.

Issue #614, opened 2025-10-21
SPMATH and WOLFBOOT_SMALL_STACK testing failures?


gojimmypi:
This issue is for reference as related to https://github.com/wolfSSL/wolfBoot/pull/613 (and may be incorrect)

I've recently added some new tests, and it appears the SP Math concurrent with Small Stack settings may be problematic. 

I will disabled the tests for now and include this PR as a TODO:

```
        exclude:
          - math: "SPMATH=1 WOLFBOOT_SMALL_STACK=1"
          - math: "SPMATHALL=1 WOLFBOOT_SMALL_STACK=1"
```

For reference from a [recent workflow](https://github.com/wolfSSL/wolfBoot/actions/runs/18696023301?pr=613):

```
Annotations
27 errors
test-lib (SPMATH=1 WOLFBOOT_SMALL_STACK=1, ecc256, sha384)
Process completed with exit code 1.
test-lib (SPMATH=1 WOLFBOOT_SMALL_STACK=1, ecc256, sha256)
Process completed with exit code 1.
test-lib (SPMATH=1 WOLFBOOT_SMALL_STACK=1, ecc256, sha3)
Process completed with exit code 1.
test-lib (SPMATH=1 WOLFBOOT_SMALL_STACK=1, ecc384, sha3)
Process completed with exit code 1.
test-lib (SPMATH=1 WOLFBOOT_SMALL_STACK=1, ecc384, sha256)
Process completed with exit code 1.
test-lib (SPMATH=1 WOLFBOOT_SMALL_STACK=1, ecc384, sha384)
Process completed with exit code 1.
test-lib (SPMATH=1 WOLFBOOT_SMALL_STACK=1, ecc521, sha256)
Process completed with exit code 1.
test-lib (SPMATH=1 WOLFBOOT_SMALL_STACK=1, ecc521, sha3)
Process completed with exit code 1.
test-lib (SPMATH=1 WOLFBOOT_SMALL_STACK=1, ecc521, sha384)
Process completed with exit code 1.
test-lib (SPMATHALL=1 WOLFBOOT_SMALL_STACK=1, ecc384, sha256)
Process completed with exit code 1.
test-lib (SPMATHALL=1 WOLFBOOT_SMALL_STACK=1, ecc384, sha384)
Process completed with exit code 1.
test-lib (SPMATHALL=1 WOLFBOOT_SMALL_STACK=1, ecc256, sha3)
Process completed with exit code 1.
[...]

PR #625, opened 2025-10-31
Improved image, keygen, sign, test lib messages. Polish

gojimmypi:
This PR improves the diagnostic messages and error checking for various executable app code (image, keygen, sign).

Also polishes for code readability, adding return results, moved arg processing to separate function, and other improvements.

Note: `test-lib` now returns non-zero error code for errors, not just failure text.

No key functionality changes.

Was [failing](https://github.com/wolfSSL/wolfBoot/actions/runs/18988825220/job/54237892133?pr=625) because of  image size difference, perhaps due to new `static inline int key_hash_ok(int id, uint8_t* digest)` and other checks. See size adjustments in https://github.com/wolfSSL/wolfBoot/pull/625/commits/6527db71a2188e016bfdad73fb488531b604ec2a

### DIFF
--- a/src/image.c
+++ b/src/image.c
@@ -2219,7 +2219,7 @@ uint8_t* wolfBoot_peek_image(struct wolfBoot_image *img, uint32_t offset,
 int keyslot_id_by_sha(const uint8_t *hint)
 {
     int id;
-    int ret = -1;
+    int ret = 0; /* TODO default to failure until proven otherwise */
     int ct = 0;
     if (hint == NULL) {
         return -1;


.github/workflows/test-library.yml
```diff
@@ -41,41 +41,119 @@ jobs:
...
-      - name: make clean
+      - name: hard clean
         run: |
-          make keysclean && make -C tools/keytools clean && rm -f include/target.h
+          # Ensure parallel build did not leave behind any debris
+          make clean || true
+          make keysclean || true
+          make -C tools/keytools clean || true
+
+          # The brute-force clean:
```
danielinux: This could be a git clean so we don't have to maintain the list

src/image.c
```diff
@@ -988,20 +988,32 @@ static int image_sha256(struct wolfBoot_image *img, uint8_t *hash)
...
 {
     uint8_t *pubkey = keystore_get_buffer(key_slot);
     int pubkey_sz = keystore_get_size(key_slot);
     wc_Sha256 sha256_ctx;
+    int ret = 0;
+
+    if (!pubkey || (pubkey_sz < 0)) {
+        return -1;
+    }
+#ifndef WOLFBOOT_KEYHASH_HAS_RESULT
```
danielinux: Doesn't this break normal sha256? Why is this only done in Sha256? The purpose of this new option is unclear/undocumented.

danielinux (changes requested):

.github/workflows/test-library.yml
gojimmypi: The helpful thing about the explicit clean is that is also reminds of the generated files, spread over multiple directories. This is the thing that tripped me up at the beginning more than anything else.

Note the brute-force clean also removed the `test-app` directory and related source files.

There's always the potential that one of the files such as `target.h` ends up included in someone's repo.

I hear you on the list maintenance though: perhaps move that to a common script?

If I document and possible move the operation to a script, would you consider keeping?

src/image.c
gojimmypi: This is something I bumped into when targeting other platforms. My goal is to make errors more obvious when plugging wolfBoot into other projects - particularly those using their own build process.

The proposed `WOLFBOOT_KEYHASH_HAS_RESULT` breadcrumb ensures the `wolfboot.h` was included with `WOLFBOOT_HASH_SHA256` defined, and that during the `key_sha256()` all of the [expected hashing config defines](https://github.com/wolfSSL/wolfBoot/blob/a559b750b072281ea472d9859ecba73603225d26/include/wolfboot/wolfboot.h#L179) are there.

>  Why is this only done in Sha256? 

I added only for sha256 at this time to determine interest; if I improve the documentation and add to other signature types, would you be willing to accept in either in this PR and/or a future one?

> Doesn't this break normal sha256? 

More specific? It seems to be working as desired.  I'd like to capture other cases & add to tests.

.github/workflows/test-library.yml
danielinux: please replace with `make distclean`. We don't need all those here. Thanks.

src/image.c
```diff
@@ -2200,18 +2223,37 @@ uint8_t* wolfBoot_peek_image(struct wolfBoot_image *img, uint32_t offset,
...
 {
     int id;
+    int ret = -1;
+    int ct = 0;
+    if (hint == NULL) {
+        return -1;
+    }
+    if (WOLFBOOT_SHA_DIGEST_SIZE <= 0) {
+        return -1;
+    }
```
danielinux: Remove, ~~this case is impossible due to checks in image.h~~

danielinux: errata: I thought we had compile-time checks in headers that at least one SHA was selected. 

Please move to header files

src/image.c
danielinux: I personally don't like this at all, even if you would name it properly with the DEBUG all in capitals it's still looks unnecessarily complex.

Compile time checks should be in headers, so we are sure they do not produce code.



Why would key_sha256() return -1 when this is not defined?

Please find a better place to validate compile time mismatch than the key

How can this work:
```
#ifndef WOLFBOOT_KEYHASH_HAS_RESULT
return -1;
#endif
```

If (as per default) the option is not enabled? Is this check misplaced?


What's the difference between `WOLFBOOT_KEYHASH_HAS_RET` later on and `WOLFBOOT_KEYHASH_HAS_RESULT` ?

Where are the configuration options to enable those with the normal Makefile buiild? any documentation?

gojimmypi: My prior explanation was poor.

This change was in the context of https://github.com/wolfSSL/wolfBoot/issues/614

Note that I updated this `key_sha256` to instead return an `int` result:

```
static void key_sha256(uint8_t key_slot, uint8_t *hash)
```

As I didn't update all of the key hash functions, so I needed to know which returned a value, and which did not.

The `WOLFBOOT_KEYHASH_HAS_RESULT` is to answer the question: 

"_Does the key hash return a value, or is it a void function_". See the new `key_hash_ok` function:

```
static inline int key_hash_ok(int id, uint8_t* digest)
{
#ifdef WOLFBOOT_KEYHASH_HAS_RET
    return key_hash(id, digest);
#else
    key_hash(id, digest);
    return 0;
#endif
}
```

For instance, as the `key_sha384` is still a void function in this PR:

```
static void key_sha384(uint8_t key_slot, uint8_t *hash)
```

If any of these fail, the caller doesn't know:

```
    wc_InitSha384(&sha384_ctx);
    wc_Sha384Update(&sha384_ctx, pubkey, (word32)pubkey_sz);
    wc_Sha384Final(&sha384_ctx, hash);
    wc_Sha384Free(&sha384_ctx);
```

This code then forces an error, as there's a known return value, but the expected breadcrumb not found:

```
#ifndef WOLFBOOT_KEYHASH_HAS_RESULT
    wolfBoot_printf("This hash result must define WOLFBOOT_KEYHASH_HAS_RESULT");
    return -1;
#endif
```

If you are in favor of returning a result for all `key_[hash]` functions, I can update them all and omit the `WOLFBOOT_KEYHASH_HAS_RESULT`  check.

src/image.c
gojimmypi: Moved in [#635](https://github.com/wolfSSL/wolfBoot/pull/635/files#diff-b7980628372a7a2cc7c3b83d6c3d7239a0541aef47410d94c7c93e8bab95282d)

PR #626, opened 2025-10-31
Added Vorago VA416x0 wolfBoot support

dgarske:
Added Vorago VA416x0 wolfBoot support

* Tested on VA41620 and VA41630 MCU's.

```
MCU: Cortex-M4 with Triple-Mode Redundancy (TMR) RAD hardening at up to 100MHz.
FLASH: The VA41630 has 256KB of internal SPI FRAM (for the VA41620 its external). FRAM is Infineon FM25V20A.
SRAM: 64KB on-chip SRAM and 256KB on-chip program memory
```

* All build settings come from .config file. For this platform use `TARGET=va416x0`. See example configuration at `config/examples/vorago_va416x0.config`.

Also:
* Added check for image header size <= sector size.
* Expanded the ML-DSA testing for level 3 and 5.
* Fix to make sure the hal/$(TARGET) is always rebuilt for test-app.


test-app/Makefile
```diff
@@ -552,6 +558,9 @@ delta-extra-data: image.bin
 	@echo "\t[CC-$(ARCH)] $@"
 	$(Q)$(CC) $(CFLAGS) -c $(OUTPUT_FLAG) $@ ../src/libwolfboot.c
 
+../hal/$(TARGET).o: ../hal/$(TARGET).c FORCE
```
dgarske, line 567: CI errors are caused by this change, which forces rebuild of the HAL object. I'll fix those cases. I am using the same HAL for Vorago boot and test application but gating features with `__WOLFBOOT`.

gojimmypi: Appreciate the note. I was wondering if my recent changes may have introduced something undesired.

hal/va416x0.ld
```diff
@@ -0,0 +1,166 @@
...
+    . = ALIGN(8);
+    PROVIDE ( end = . );
+    PROVIDE ( _end = . );
+    PROVIDE ( _start_heap = . );
+    . = . + _Min_Heap_Size;
+    . = . + _Min_Stack_Size;
+    . = ALIGN(8);
+    PROVIDE ( END_STACK = . );
+    PROVIDE ( _end_stack = . );
+  } >RAM0
```
danielinux: Perhaps unnecessary, since wolfBoot will use no heap

hal/va416x0.ld
```diff
@@ -0,0 +1,166 @@
+_Min_Heap_Size  = 0x00000200; /* required amount of heap  */
```
danielinux: Consider removing Heap indications/limits and checks

hal/vorago/board.h
danielinux, line 1: Could these files be in hal/vorago instead?

config/examples/vorago_va416x0.config
```diff
@@ -0,0 +1,79 @@
...
+VORAGO_SDK_DIR?=$(PWD)/../VA416xx_SDK/
+
+# Use Vorago FRAM driver
+USE_HAL_SPI_FRAM=1
+
+#CFLAGS_EXTRA+=-DWOLFBOOT_EDAC_RAM_SCRUB=1000
+#CFLAGS_EXTRA+=-DWOLFBOOT_EDAC_ROM_SCRUB=125
+
+# Optionally restore clock to heart-beat oscillator after boot
+#CFLAGS_EXTRA+=-DWOLFBOOT_RESTORE_CLOCK
```
danielinux: Perhaps outside the scope of this PR, but we could make this a proper option via `options.mk`. By default, we tend to restore the clocks to be "transparent", on other platforms, so this should perhaps be defaulting to =1 in that case. A 'generic' support may skip hal_prepare_boot entirely.

Note: We already skip `hal_prepare_boot` in other situations/platforms (TrustZone-M when we stage a non-secure app that may not have access to PLL / Clock config later, and it's usually aware that some HW configuration is left to bootloader).

hal/va416x0.c
```diff
@@ -0,0 +1,540 @@
...
+
+}
+
+void hal_prepare_boot(void)
+{
+#ifdef DEBUG_UART
+    uart_flush();
+#endif
+
+#ifdef WOLFBOOT_RESTORE_CLOCK
```
danielinux, line 535: See comments above. Currently requires forcing CFLAGS, and target-specific. Might be a proper option that skips the function entirely.

config/examples/vorago_va416x0.config
dgarske: Fixed

hal/va416x0.c
dgarske: Fixed

hal/va416x0.ld
dgarske: Fixed. Removed heap support.

hal/va416x0.ld
dgarske: Fixed, removed heap. Keeping stack. This was just a .ld template from their SDK.

hal/vorago/board.h
dgarske: Fixed

PR #720, opened 2026-03-11
Add wolfBoot port for STM32N6 (NUCLEO-N657X0-Q)

aidangarske:
## Summary

- **Cortex-M55** support added to `arch.mk` (inherits Cortex-M33 settings, adds `-mcpu=cortex-m55`).
- **HAL** in `hal/stm32n6.{c,h,ld}`: clock tree to 600 MHz (HSI -> PLL1 -> IC1/IC2/IC6/IC11), OCTOSPI/XSPI2 in single-SPI mmap mode with NOR `FAST_READ_4B`, I/D-cache control, SAU (TZ and non-TZ), USART1 for `wolfBoot_printf`, and the `hal_flash_*` / `ext_flash_*` API on top of an `octospi_cmd` page-program/sector-erase helper.
- **wolfBoot reset path** (`src/boot_arm.c`): naked `isr_reset` for STM32N6 that sets `SCB->VTOR = WOLFBOOT_ORIGIN`, reloads MSP from word 0 of the FSBL vector table, then tail-calls into `isr_reset_c`. Also writes `MSPLIM_S` to `_end_bss` after BSS init (raw T2 MSR encoding because `-mcmse` is not in the build) so a future stack overflow trips a clean `UsageFault.STKOF`.
- **Linker scripts**: `hal/stm32n6.ld` (wolfBoot in AXISRAM2 at `WOLFBOOT_ORIGIN=0x34180400`, 256 KB region) and `test-app/ARM-stm32n6.ld` (app XIP at `0x70020400`).
- **Test app** (`test-app/app_stm32n6.c`): bare-metal LED blinker that reads partition state, calls `wolfBoot_success()`, blinks blue (v1) or red (v>=2).
- **Flash script** (`tools/scripts/stm32n6_flash.sh`): wraps `STM32_SigningTool_CLI` and OpenOCD for one-shot programming. Supports `--skip-build`, `--app-only`, `--test-update` (signs a v2 with `IMAGE_HEADER_SIZE` matching the wolfBoot build), `--halt`.
- **OpenOCD** config (`config/openocd/openocd_stm32n6.cfg`): brings up XSPI2 single-SPI mmap on `reset-init` so flash programming via stmqspi works without the wolfBoot HAL.
- **Documentation** (`docs/Targets.md`): full STM32N6 section with memory layout, OTP fuse requirements, jumper table, build/flash/debug instructions, TrustZone notes.
- **CI** (`.github/workflows/test-configs.yml`): adds `stm32n6.config` to the build matrix.

## Non-obvious gotchas captured in the port

1. **Boot ROM hands off with CPU on PLL1.** `clock_config()` parks `CPUSW`/`SYSSW` on HSI before clearing `PLL1ON`. Disabling PLL1 while the CPU runs on it stops the CPU clock. OpenOCD-loaded boot masks this because `reset_halt` leaves the CPU on reset-state HSI; cold boot was the only path that hit it.
2. **`STM32_SigningTool_CLI -ep <addr>` writes the address verbatim with no Thumb bit.** The flash script never passes `-ep`/`-la`; the tool reads the entry point with the Thumb bit set from the binary's vector table at offset `0x04`. Wrong settings result in a silent hardfault on the first instruction.
3. **`XSPIM_CR` swaps XSPI1/XSPI2 port routing.** Boot ROM leaves `XSPIM_CR.MODE=1`, sending XSPI2 traffic to Port 1 (nothing wired there). `octospi_init()` disables XSPI1, RCC-resets XSPI2, then writes `XSPIM_CR=0` so XSPI2 reaches Port 2 (where MX25UM51245G is wired).
4. **MX25UM51245G stays in single-SPI mode after Boot ROM hand-off.** The earlier theory that the Boot ROM left the chip in OPI/DTR was wrong; single-SPI `FAST_READ_4B` works directly.
5. **XIP mmap needs D-cache.** Single-word XIP reads or non-cacheable mapping stall OCTOSPI with `BUSY=1` and full FIFO. `octospi_init()` ends with one warmup XIP read of `WOLFBOOT_PARTITION_BOOT_ADDRESS`; without it the first XIP access from `wolfBoot_start` stalls (looks like a prefetch/AXI pipeline cold-start quirk).
6. **Boot ROM does NOT load MSP from the FSBL vector table.** It just `BX`es the entry point with MSP still pointing at the Boot ROM's tiny scratch stack (~`0x341037f4`). Naked `isr_reset` in `src/boot_arm.c` sets `VTOR` and reloads MSP before any C prologue. Without this, wolfBoot's signature-verify path overflows the scratch stack, corrupts a saved LR into a `FNC_RETURN` signature (`0xFEFFFFFx`), the next `BX LR` does an S->NS state transition with garbage PC, the resulting MemManage fault is also stack-busted, double-fault -> LOCKUP.
7. **`octospi_cmd` data pump uses byte-wide DR access driven by `FLEVEL`.** The DR32+`FTHRES` polling path stalls on transfers larger than the FIFO depth (32 bytes). `FLEVEL` is correct for any transfer size and threshold.
8. **`ext_flash_write` / `ext_flash_erase` must invalidate the D-cache for the XIP alias.** Without it, reads of just-modified regions return stale data. The A/B swap and signature re-verify after swap depend on this.
9. **Sign tool reads `IMAGE_HEADER_SIZE` from env.** The flash script's `--test-update` invocation now exports it from `.config` so v2 has the same 1024-byte manifest header as v1 (default is 256, which causes hash verification to fail at update time).

## Validation

- Cold boot from QSPI NOR, no pending update: `wolfBoot Init` -> `Verifying signature...done` -> app banner, blue/red LED.
- Cold boot from QSPI NOR with pending v2 update: 3-way A/B swap completes, v2 boots, `TESTING -> SUCCESS`.
- Subsequent cold boots run v2 with state `SUCCESS`.
- Non-TZ build (`config/examples/stm32n6.config`): builds and runs end-to-end.
- TZ build (`config/examples/stm32n6-tz.config`): config in place but the test-app linker script needs a separate fix; pre-existing issue, not part of this work.


hal/stm32n6.h
```diff
@@ -0,0 +1,444 @@
...
+#define GPIO_MODE_ANALOG        0x3
+
+/* GPIO speed values */
+#define GPIO_SPEED_LOW          0x0
+#define GPIO_SPEED_MEDIUM       0x1
+#define GPIO_SPEED_HIGH         0x2
+#define GPIO_SPEED_VERY_HIGH    0x3
+
+
+/*** XSPI2 (External SPI for NOR flash) ***/
```
dgarske: This driver may already be in wolfBoot. See hal/spi/spi_drv_stm32.h OCTOSPI starting at line 591. The actual implementation is in spi_drv_stm32.c and you can use qspi_transfer from src/qspi_flash.c. It would be great if you could leverage those existing drivers. You just need to set the right base address for the N6. I think so far that driver has only been ported to the H7.

hal/stm32n6.h
```diff
@@ -0,0 +1,444 @@
...
+#define XSPI_MODE_OCTAL        4
+
+
+/*** XSPIM (XSPI I/O Manager) ***/
+#define XSPIM_BASE              (0x5802B400UL)
+#define XSPIM_CR                (*(volatile uint32_t *)(XSPIM_BASE + 0x00))
+
+/*** NOR Flash Commands (Macronix MX25UM51245G) ***/
+/* Single-SPI mode commands (initial boot) */
+#define NOR_CMD_WRITE_ENABLE    0x06
```
dgarske: Same. These are standard QSPI commands and already well supported in src/qspi_flash.c. If you find there are missing features then use this as an opportunity to add them.

hal/stm32n6.h
dgarske: This driver may already be in wolfBoot. See hal/spi/spi_drv_stm32.h OCTOSPI starting at line 591. The actual implementation is in spi_drv_stm32.c and you can use qspi_transfer from src/qspi_flash.c. It would be great if you could leverage those existing drivers. You just need to set the right base address for the N6. I think so far that driver has only been ported to the H7.

hal/stm32n6.h
```diff
@@ -0,0 +1,444 @@
...
+#define NOR_SECTOR_SIZE         0x1000      /* 4KB */
+#define NOR_BLOCK_SIZE          0x10000     /* 64KB */
+#define NOR_DEVICE_SIZE         (64 * 1024 * 1024) /* 64MB */
+#define NOR_DEVICE_SIZE_LOG2    26          /* XSPI DEVSIZE: 2^26 = 64MB */
+
+
+/*** USART1 (Debug UART) ***/
+#define USART1_BASE             (0x52001000UL)
+
+#define USART1_CR1              (*(volatile uint32_t *)(USART1_BASE + 0x00))
```
dgarske: I prefer using `#define USART_CR2(num)` so its easy for a customer to switch UARTs. See the stm32h7 driver since its identical. Also why we need to unify our bare metal drivers with wolfHAL

arch.mk
```diff
@@ -352,9 +355,17 @@ else
     CORTEXM_ARM_EXTRA_CFLAGS+=-DWOLFSSL_ARMASM -DWOLFSSL_ARMASM_NO_HW_CRYPTO \
                               -DWOLFSSL_ARMASM_NO_NEON -DWOLFSSL_ARMASM_THUMB2
   endif
+  ifeq ($(CORTEX_M55),1)
+    CORTEX_M33=1
```
dgarske, line 458: Is it using the same CORTEX_M33 macro for both? Could we add a CORTEX_M55 macro and update the places required?

Makefile
```diff
@@ -653,6 +652,12 @@ stack-usage: wolfboot.bin
 image-header-size: wolfboot.bin
 	$(Q)echo $(IMAGE_HEADER_SIZE) > .image_header_size
 
+## Target-specific flash targets
+ifeq ($(TARGET),stm32n6)
+flash: wolfboot.bin test-app/image_v1_signed.bin
```
dgarske, line 699: Handy. I like it.

dgarske (changes requested):
Great work! I haven't tested on hardware yet, but have looked over each line.

arch.mk
aidangarske: Moved to CORTEX_55 macro

hal/stm32n6.h
aidangarske: stm32n6.h now includes hal/spi/spi_drv_stm32.h directly after setting OCTOSPI_BASE. Only a few N6-specific bits not in the shared header are defined locally (FMODE_MMAP, TEF, DLYBYP).

hal/stm32n6.h
aidangarske: The command values match src/qspi_flash.c but we can't reuse qspi_transfer() becuase it disables/enables the peripheral every transfer which breaks XIP. I think the N6 needs custom transfer code that aborts/restores memory mapped mode

hal/stm32n6.h
aidangarske: Done. UART macros are now use base address (UART_CR1(base), etc.) same pattern as the STM32H7 driver.

docs/release-stm32n6.md
```diff
@@ -0,0 +1,54 @@
+# wolfBoot Secure Boot on the STM32N6
```
dgarske: Does this file need to be checked in? Prefer any of these useful details in docs/Targets.md

dgarske (changes requested):

docs/release-stm32n6.md
aidangarske: No I didnt want this in will remove

dgarske (dismissed):

dgarske (changes requested):
This PR is in good shape, but the boot ROM from QSPI isn't working. @aidangarske please work on it. Not rush on this, it can be post release. I think it does require some OTP registers to be set, but I tried setting OTA124 and it didn't change the results. Note you also have to set BOOT0/BOOT1 jumpers to switch between boot modes.

src/boot_arm.c
```diff
@@ -419,9 +419,10 @@ void isr_empty(void)
  *
  */
 
-#ifdef TZEN
+#if defined(__ARM_FEATURE_CMSE) && (__ARM_FEATURE_CMSE == 3U) && \
```
aidangarske, line 465: **🟠 [Medium] TZEN=1 boot path sets secure VTOR to non-secure application memory**
**Category:** Security

When `TZEN=1` and `CORTEX_M55` are both defined (STM32N6 TrustZone build), the `!defined(CORTEX_M55)` guard causes the VTOR macro to use the secure VTOR address (`0xE000ED08`) instead of the non-secure VTOR (`0xE002ED08`). In `do_boot()`, the code writes `VTOR = ((uint32_t)app_offset)` where `app_offset` is `0x70020400` — an address in SAU region 1 which is marked as Non-Secure. This means the secure vector table pointer references NS-controlled memory.

On other TZEN platforms (stm32h5, stm32l5, stm32u5), the code uses the NS VTOR (`0xE002ED08`) and transitions via `blxns`, leaving the secure VTOR pointing to wolfBoot's own (secure) vector table. The STM32N6 path instead points the secure VTOR to non-secure memory and does a regular `mov pc` jump.

If a secure exception occurs after boot (e.g., SecureFault, NMI, HardFault in secure state from hardware error), the exception handler address would be fetched from NS-controlled memory. In practice, exploitation is limited because: (1) the app runs from XIP NOR flash which is read-only in memory-mapped mode, (2) no secure code remains active after boot. However, this violates the TrustZone security model where secure structures should not reference NS memory.

```c
#if defined(__ARM_FEATURE_CMSE) && (__ARM_FEATURE_CMSE == 3U) && \
    defined(TZEN) && !defined(CORTEX_M55)
#define VTOR (*(volatile uint32_t *)(0xE002ED08)) /* Non-secure VTOR */
#else
#define VTOR (*(volatile uint32_t *)(0xE000ED08))  /* Used for CORTEX_M55+TZEN */
#endif

// In do_boot():
VTOR = ((uint32_t)app_offset);  // Sets SECURE VTOR to NS address 0x70020400
asm volatile("msr msp, %0" ::"r"(app_end_stack));
asm volatile("mov pc, %0":: "r"(app_entry));  // No blxns transition
```

**Recommendation:** For the TZEN=1 STM32N6 boot path, consider: (1) writing the app vector table address to the NS VTOR (`0xE002ED08`) instead of the secure VTOR, and (2) keeping the secure VTOR pointing to wolfBoot's own vector table in SRAM (secure memory). If the STM32N6 truly doesn't need a `blxns` transition, at minimum avoid corrupting the secure VTOR by writing a separate NS VTOR update: `*(volatile uint32_t *)(0xE002ED08) = (uint32_t)app_offset;`

hal/stm32n6.c
```diff
@@ -0,0 +1,686 @@
...
+    while (OCTOSPI_CR & OCTOSPI_CR_ABORT)
+        ;
+    return -1;
+}
+
+static void RAMFUNCTION octospi_write_enable(void)
+{
+    octospi_cmd(0, WRITE_ENABLE_CMD, 0, SPI_MODE_NONE,
+                NULL, 0, SPI_MODE_NONE, 0);
+}
```
aidangarske, line 145: **🟠 [Medium] octospi_write_enable ignores return value allowing silent flash operation failures**
**Category:** Logic

The `octospi_write_enable()` function calls `octospi_cmd()` but discards the return value. If the Write Enable (WREN) command fails due to a transfer error (TEF), the function returns silently. The caller (`nor_flash_write` and `nor_flash_erase`) then proceeds with the page program or sector erase command. Without WREN set, the NOR flash ignores the write/erase command — the operation appears to succeed (no SPI error) but no data is actually written/erased.

In a bootloader context, this means a firmware update swap could silently produce corrupted data in the boot or update partition. While wolfBoot's signature verification would catch this corruption before booting, the update would fail in an unclear way, and repeated WREN failures could leave the device in a state where updates never succeed (boot loop).

```c
static void RAMFUNCTION octospi_write_enable(void)
{
    octospi_cmd(0, WRITE_ENABLE_CMD, 0, SPI_MODE_NONE,
                NULL, 0, SPI_MODE_NONE, 0);  // Return value discarded
}

// Called from nor_flash_write:
octospi_write_enable();  // If this fails silently...
ret = octospi_cmd(0, PAGE_PROG_4B_CMD, ...);  // ...flash ignores the write
```

**Recommendation:** Return the result of `octospi_cmd()` from `octospi_write_enable()` and check it in `nor_flash_write()` and `nor_flash_erase()` before proceeding with the program/erase command. Example:
```c
static int RAMFUNCTION octospi_write_enable(void) {
    return octospi_cmd(0, WRITE_ENABLE_CMD, ...);
}
```
And in callers:
```c
if (octospi_write_enable() < 0) { ret = -1; break; }
```

tools/scripts/stm32n6_flash.sh
```diff
@@ -0,0 +1,169 @@
...
+    echo -e "${CYAN}  image_v2_signed.bin -> NOR ${UPDATE_ADDR}${NC}"
+    OPENOCD_CMDS+="flash write_image erase ${WOLFBOOT_ROOT}/test-app/image_v2_signed.bin ${UPDATE_ADDR}; "
+
+    # Write update trigger: "pBOOT" (0x70 0x42 0x4F 0x4F 0x54) at end of
+    # update partition. wolfBoot reads this trailer to detect a pending update.
+    # Address: UPDATE_ADDR + PARTITION_SIZE - 5
+    PART_SIZE=$(grep -E '^WOLFBOOT_PARTITION_SIZE' "${WOLFBOOT_ROOT}/.config" | head -1 | sed 's/.*=//;s/[[:space:]]//g')
+    TRIGGER_ADDR=$(printf "0x%08x" $(( ${UPDATE_ADDR} + ${PART_SIZE} - 5 )))
+    echo -e "${CYAN}  Update trigger -> NOR ${TRIGGER_ADDR}${NC}"
+    printf 'pBOOT' > /tmp/trigger_magic.bin
```
aidangarske, line 167: **🔵 [Low] Flash script uses predictable temporary file path**
**Category:** Security

The flash script writes the update trigger magic bytes to a hardcoded predictable path `/tmp/trigger_magic.bin`. In a multi-user development environment, another user could pre-create a symlink at this path pointing to a sensitive file, causing the `printf` write to overwrite that file. The impact is limited since this is a development flashing tool, not production code, and the data written is just 5 bytes ('pBOOT').

```c
printf 'pBOOT' > /tmp/trigger_magic.bin
OPENOCD_CMDS+="flash write_image /tmp/trigger_magic.bin ${TRIGGER_ADDR}; "
```

**Recommendation:** Use `mktemp` for a unique temporary file:
```bash
TRIGGER_FILE=$(mktemp /tmp/trigger_magic.XXXXXX)
printf 'pBOOT' > "${TRIGGER_FILE}"
OPENOCD_CMDS+="flash write_image ${TRIGGER_FILE} ${TRIGGER_ADDR}; "
# Clean up after use
trap 'rm -f ${TRIGGER_FILE}' EXIT
```

docs/Targets.md
```diff
@@ -1819,6 +1820,251 @@ c
...
+
+### Build and Flash
+
+Use the example configuration and build:
+
+```sh
+# Without TrustZone:
+cp config/examples/stm32n6.config .config
+make
+make flash
```
aidangarske, line 2167: **⚪ [Info] Documentation describes TZEN=1 secure SRAM alias but code uses non-secure alias for both modes**
**Category:** Logic

The documentation states that with TrustZone enabled, wolfBoot runs from the secure SRAM alias at `0x24180400`, while without TrustZone it runs from `0x34180400`. However, `arch.mk` sets `WOLFBOOT_ORIGIN=0x34180400` unconditionally for both TZEN=0 and TZEN=1, with a comment explaining: 'Use the same address for both TZEN=0 and TZEN=1 since the Boot ROM always uses the non-secure alias.'

The flash script also hardcodes `SRAM_ADDR=0x34180400` without TZEN detection, despite the PR description stating 'auto-detects TZEN for correct SRAM load address.'

This is a documentation inconsistency rather than a code bug — the code and arch.mk comment are self-consistent, but the docs and PR description are misleading.

```c
# arch.mk:
WOLFBOOT_ORIGIN=0x34180400  # Same for both TZEN=0 and TZEN=1

# docs/Targets.md claims:
# AXISRAM2 — with TrustZone (TZEN=1, secure alias):
#   0x24180400  wolfBoot FSBL (secure — IDAU marks 0x24xxxxxx as secure)

# stm32n6_flash.sh:
SRAM_ADDR=0x34180400  # No TZEN detection
```

**Recommendation:** Update the documentation in `docs/Targets.md` to clarify that wolfBoot is always loaded and linked at the non-secure alias `0x34180400`, with SAU configuration in `hal_init()` providing the necessary access control. The memory layout table should note that `0x24180400` is the secure alias of the same physical memory, not a separate load address. Also update the PR description to remove the claim about TZEN auto-detection in the flash script.

aidangarske (commented):
## Skoll Code Review

**Scan type:** `review-security`
**Overall recommendation:** `COMMENT`
**Findings:** 5 total — 4 posted, 1 skipped

### Posted findings
- [Medium] **TZEN=1 boot path sets secure VTOR to non-secure application memory** — `src/boot_arm.c:422-466`
- [Medium] **octospi_write_enable ignores return value allowing silent flash operation failures** — `hal/stm32n6.c:143-147`
- [Low] **Flash script uses predictable temporary file path** — `tools/scripts/stm32n6_flash.sh:142`
- [Info] **Documentation describes TZEN=1 secure SRAM alias but code uses non-secure alias for both modes** — `docs/Targets.md:1855-1860`

### Skipped findings
- [Low] ~~Unaligned uint32_t pointer dereference in octospi_cmd data transfer loop~~

dgarske:
@aidangarske how's this going? If you won't have time in the next week let me know and I can finish it up.

danielinux (dismissed):

dgarske (dismissed):

danielinux (approved):

PR #745, opened 2026-04-08
Fixes 20260408

danielinux:
F/2273 - sign: scrub key buffers before free (`1c824539`)
F/2274 - Scrub RSA keygen private material (`87d5a7fc`)
F/2275 - Zero ECC keygen private buffers (`114140c2`)
F/2276 - Zero EdDSA keygen private data (`6f40afae`)
F/2277 - zero ML-DSA private key buffer (`32783034`)
F/2258 - Add restricted key mask authenticity tests (`8399e3ed`)
F/1892 - Propagate encrypt key flash errors (`afef21fe`)
F/1893 - Propagate erase encrypt key write failures (`13370613`)
F/1894 - Fix policy_create PCR digest validation (`6b8ade6b`)
F/1895 - Fix sign encrypted output open failure (`34438999`)
F/1898 - Check image reopen failures in sign tool (`5fd09eed`)
F/2247 - Use constant-time RSA hash comparison (`9a7930dd`)
F/2255 - Protect bootloader before application boot (`f32c275d`)
F/2259 - Add auth type coverage for unit-image (`644e70e7`)
F/2260 - Add auth-only invalid update test (`ff60286e`)
F/2261 - Add RAM_CODE self-update unit coverage (`3659d210`)
F/2262 - Strengthen same-version RAM update test (`1353e933`)
F/2266 - Fix sign header TLV overflow sizing (`ea6700f1`)
F/2267 - Reject oversized delta source offsets (`fefd74e0`)
F/1897 - fix memmove large-length backward copy (`7526fd23`)
F/2248 - Use constant-time TPM secret checks (`78de4a79`)
F/2249 - Use constant-time encryption key validation (`56c46be7`)
F/2252 - Use fixed-length erased-key check (`830869c3`)
F/2256 - enforce skip-verify prerequisites (`a09babb5`)
F/2264 - Add equal-version update-disk regression test (`395202bb`)
F/2268 - Reject valid zero-size delta images (`29367f81`)
F/2269 - Fix total size type in update flash (`fffd8543`)
F/1888 - zeroize update key material (`05823677`)
F/1889 - zero custom encrypt stack buffers (`7c1c8631`)
F/1890 - zeroize swap trailer key buffer (`a870af20`)
F/1891 - Scrub sign-tool encryption material (`3f1906fa`)
F/2253 - Use constant-time delta base hash compare (`10035dbc`)
F/2257 - Warn when DISABLE_BACKUP is enabled (`13d232c4`)
F/2270 - Add sign/parser roundtrip tests (`fb9faffc`)
F/2271 - Add delta roundtrip edge-case coverage (`08d39cf4`)


src/update_flash.c
```diff
@@ -498,9 +498,18 @@ static int RAMFUNCTION wolfBoot_swap_and_final_erase(int resume)
     wb_flash_erase(boot, WOLFBOOT_PARTITION_SIZE - eraseLen, eraseLen);
 
 #ifdef EXT_ENCRYPTED
+    int ret;
```
dgarske: **🟠 [High] `int ret` declared after statements — C89 violation in embedded target code**
*🚫 BLOCK*  `bug`

The new `int ret;` declaration at line 501 appears after executable statements (lines 441-498: `wolfBoot_open_image`, `wolfBoot_get_partition_state`, `memcpy`, `wb_flash_erase`, etc.). This is valid in C99+ but not in C89/C90. wolfBoot targets embedded compilers including CCRX (this same file has `#pragma section FRAM` for CCRX at line 799), which may enforce C89 rules. The `#ifdef EXT_ENCRYPTED` preprocessor guard does NOT create a C block scope, so the declaration is mid-function from the C compiler's perspective.

**Suggestion:**
```suggestion
Move `int ret;` to the top of the function alongside other locals (around line 439), still guarded by `#ifdef EXT_ENCRYPTED`:

```c
    uint32_t tmpBuffer[TRAILER_OFFSET_WORDS + 1];
#ifdef EXT_ENCRYPTED
    int ret;
#endif
```
```

File: src/libwolfboot.c [hunk 2]
```diff
@@ -1742,17 +1772,15 @@ Aes aes_dec, aes_enc;
...
     uint8_t *stored_nonce;
     uint8_t *key;
 #endif
-    uint8_t ff[ENCRYPT_KEY_SIZE];
-
 #ifdef WOLFBOOT_RENESAS_TSIP
-    int ret;
     wrap_enc_key_t* enc_key;
     devId = RENESAS_DEVID + 1;
     enc_key =(wrap_enc_key_t*)RENESAS_TSIP_INSTALLEDENCKEY_ADDR;
```
dgarske, line 1786: **🟠 [High] `aes_init()` does not check `wolfBoot_get_encrypt_key()` return — inconsistent with `chacha_init()`**
*🚫 BLOCK*  `bug`

In `chacha_init()` (lines 1724-1726), the return value of `wolfBoot_get_encrypt_key(key, stored_nonce)` is checked and on failure the function jumps to `exit`. In `aes_init()` at line 1791, the same call is made but the return value is silently discarded. If the custom key retrieval fails, `aes_init` will proceed with uninitialized stack-allocated `key[]` and `stored_nonce[]` buffers. While `encrypt_key_is_valid()` at line 1810 may catch all-zero buffers, uninitialized stack data could pass the validation. This is a clear inconsistency between the two init functions introduced by this PR.

**Suggestion:**
```suggestion
```c
#elif defined(CUSTOM_ENCRYPT_KEY)
    ret = wolfBoot_get_encrypt_key(key, stored_nonce);
    if (ret != 0)
        goto exit;
```
```

File: tools/keytools/sign.c [hunk 3]
```diff
@@ -1107,6 +1143,114 @@ static int sign_digest(int sign, int hash_algo,
...
+    if (CMD.sign != NO_SIGN) {
+        header_size_align_8(&idx);
+        header_size_append_tag(&idx, CMD.signature_sz);
+
+        if (CMD.hybrid) {
+            header_size_align_8(&idx);
+            header_size_append_tag(&idx, CMD.secondary_signature_sz);
+        }
+
+        if (CMD.policy_sign) {
```
dgarske, line 1239: **🟡 [Medium] `header_required_size()` uses `CMD.policy_sz` before it is initialized**
*💡 SUGGEST*  `bug`

`header_required_size()` is called at line 1292 to determine if the header needs to grow for a cert chain. At that point, `CMD.policy_sz` is still 0 (it gets set to `CMD.signature_sz` at line 1746, deep inside `make_header_ex`). When `CMD.policy_sign` is true, the calculation at line 1247 computes `0 + sizeof(uint32_t) = 4` bytes for the policy TLV instead of `CMD.signature_sz + sizeof(uint32_t)` (e.g., 260 bytes for RSA2048). This undercounting means the header might not be grown sufficiently when both `--policy` and `--cert-chain` are used together. The `header_append_tag()` overflow check at line 232 would catch the overflow at runtime and `exit(1)`, so no corrupt image is produced, but the tool would fail unnecessarily.

**Suggestion:**
```suggestion
Use `CMD.signature_sz` directly (since `policy_sz` is always set to `signature_sz`):
```c
        if (CMD.policy_sign) {
            header_size_align_8(&idx);
            header_size_append_tag(&idx,
                CMD.signature_sz + (uint32_t)sizeof(uint32_t));
        }
```
```

File: tools/keytools/keygen.c [hunk 4]
```diff
@@ -581,6 +592,13 @@ static void keygen_rsa(const char *keyfile, int kbits, uint32_t id_mask)
...
+
+cleanup:
+    wc_ForceZero(priv_der, sizeof(priv_der));
+    if (rsa_init)
+        wc_FreeRsaKey(&k);
+    if (exit_code != 0)
+        exit(exit_code);
 }
 
 #define MAX_ECC_KEY_SIZE 66
```
dgarske, line 605: **🟡 [Medium] `keygen_ecc` and `keygen_ml_dsa` not refactored to zero private material on error paths**
*💡 SUGGEST*  `convention`

The PR adds goto-cleanup patterns with `wc_ForceZero()` for RSA, Ed25519, and Ed448 keygen functions, but `keygen_ecc` (line 620) and `keygen_ml_dsa` (line ~1040) still use direct `exit()` calls on error paths. For `keygen_ecc`, if any error occurs after the private scalar `d[]` is populated (line 633) but before the cleanup code at line 696, the `exit()` bypasses `wc_ForceZero(d, ...)`. Similarly for `keygen_ml_dsa`, if any error occurs after `priv = malloc(...)` (line 1102) but before line 1208, the `free(priv)` with zeroing is skipped. While `exit()` terminates the process (so memory is reclaimed by the OS), this is inconsistent with the hardening treatment given to the other keygen functions in this PR.

**Suggestion:**
```suggestion
Refactor `keygen_ecc` and `keygen_ml_dsa` with the same `exit_code` + `goto cleanup` pattern used for RSA/Ed25519/Ed448.
```

tools/keytools/sign.c
```diff
@@ -1937,6 +2101,12 @@ static int make_header_ex(int is_diff, uint8_t *pubkey, uint32_t pubkey_sz,
         fclose(f);
```
dgarske, line 2112: **🟡 [Medium] Output file `f` not closed on failure path in `make_header_ex`**
*💡 SUGGEST*  `bug`

The output file handle `f` is opened at line 1973 and closed on the success path at lines 2100-2101, but the `failure:` cleanup block (line 2103) does not close `f`. If any of the new `goto failure` paths introduced by this PR fire after `f` is opened (e.g., encryption key read failure at lines 2028-2047), `f` is leaked. The PR already correctly manages `fek` and `fef` in the failure block with NULL checks. This is partly pre-existing but is expanded by the PR's conversion of `exit(1)` calls to `goto failure`.

**Suggestion:**
```suggestion
Add `f` cleanup to the failure block (set `f = NULL` after closing on the success path):
```c
    if (f) {
        fclose(f);
        f = NULL;
    }
failure:
    ...
    if (f)
        fclose(f);
```
```

File: src/update_flash.c [hunk 6]
```diff
@@ -498,9 +498,18 @@ static int RAMFUNCTION wolfBoot_swap_and_final_erase(int resume)
     wb_flash_erase(boot, WOLFBOOT_PARTITION_SIZE - eraseLen, eraseLen);
```
dgarske, line 517: **🟡 [Medium] Missing `wolfBoot_get_encrypt_key()` return check in `wolfBoot_swap_and_final_erase`**
*💡 SUGGEST*  `bug`

At line 488, `wolfBoot_get_encrypt_key()` is called to retrieve the encryption key into `tmpBuffer`, but its return value is not checked. If the key retrieval fails, `tmpBuffer` may contain invalid data, which is then staged into the boot partition (line 494) and later used to reinitialize encryption (line 503). The PR added a return check for `wolfBoot_set_encrypt_key` (line 503) but not for the `get` call. This is an inconsistency with the PR's theme of propagating encrypt key errors.

**Suggestion:**
```suggestion
```c
        ret = wolfBoot_get_encrypt_key((uint8_t*)tmpBuffer,
            (uint8_t*)&tmpBuffer[ENCRYPT_KEY_SIZE/sizeof(uint32_t)]);
        if (ret != 0) {
            /* handle error */
        }
```
```

File: src/update_flash_hwswap.c [hunk 7]
```diff
@@ -105,6 +105,10 @@ void RAMFUNCTION wolfBoot_start(void)
     (void)hal_hsm_disconnect();
 #elif defined(WOLFBOOT_ENABLE_WOLFHSM_SERVER)
     (void)hal_hsm_server_cleanup();
+#endif
+#ifndef TZEN
+    if (hal_flash_protect(WOLFBOOT_ORIGIN, BOOTLOADER_PARTITION_SIZE) < 0)
```
dgarske, line 110: **🔵 [Low] `update_flash_hwswap.c` uses `boot_panic()` without error message — inconsistent**
*🔧 NIT*  `convention`

Three of the four `wolfBoot_start()` variants log an error message with `wolfBoot_printf` before calling `wolfBoot_panic()` when `hal_flash_protect` fails, but `update_flash_hwswap.c` calls the local `boot_panic()` (a `while(1)` spin) with no diagnostic output. This makes debugging harder on hwswap platforms.

**Suggestion:**
```suggestion
Add an error message for consistency:
```c
#ifndef TZEN
    if (hal_flash_protect(WOLFBOOT_ORIGIN, BOOTLOADER_PARTITION_SIZE) < 0) {
        wolfBoot_printf("Error protecting bootloader flash region\n");
        boot_panic();
    }
#endif
```
```

File: src/string.c [hunk 8]
```diff
@@ -288,14 +288,25 @@ void RAMFUNCTION *memcpy(void *dst, const void *src, size_t n)
     !defined(__CCRX__)
 void *memmove(void *dst, const void *src, size_t n)
 {
-    int i;
+    size_t i;
     if (dst == src)
         return dst;
     if (src < dst)  {
         const char *s = (const char *)src;
         char *d = (char *)dst;
```
dgarske, line 293: **⚪ [Info] memmove word-copy fast path not gated behind FAST_MEMCPY**
*🔧 NIT*  `convention`

The new memmove backward copy adds a word-aligned fast path (lines 296-310), but this code is unconditionally compiled. The analogous memcpy forward word-copy (line 263) is gated behind `#ifdef FAST_MEMCPY`. For consistency and to keep code size minimal on constrained targets that don't opt in, consider gating the memmove word path similarly. The memmove logic itself is correct — both byte-by-byte and word-at-a-time paths are verified.

**Suggestion:**
```suggestion
Gate the optimization behind `#ifdef FAST_MEMCPY` for consistency.
```

dgarske (changes requested):
## 🐺 Skoll Code Review

**Overall recommendation:** `REQUEST_CHANGES`
**Findings:** 8 total — 8 posted, 0 skipped

### Posted findings
- [High] **`int ret` declared after statements — C89 violation in embedded target code** — `src/update_flash.c:501`
- [High] **`aes_init()` does not check `wolfBoot_get_encrypt_key()` return — inconsistent with `chacha_init()`** — `src/libwolfboot.c:1791`
- [Medium] **`header_required_size()` uses `CMD.policy_sz` before it is initialized** — `tools/keytools/sign.c:1244-1248`
- [Medium] **`keygen_ecc` and `keygen_ml_dsa` not refactored to zero private material on error paths** — `tools/keytools/keygen.c:620-707`
- [Medium] **Output file `f` not closed on failure path in `make_header_ex`** — `tools/keytools/sign.c:2100-2103`
- [Medium] **Missing `wolfBoot_get_encrypt_key()` return check in `wolfBoot_swap_and_final_erase`** — `src/update_flash.c:488`
- [Low] **`update_flash_hwswap.c` uses `boot_panic()` without error message — inconsistent** — `src/update_flash_hwswap.c:110-111`
- [Info] **memmove word-copy fast path not gated behind FAST_MEMCPY** — `src/string.c:296-310`

*Review generated by Skoll via openclaw*

src/update_flash_hwswap.c
danielinux: wontfix

include/image.h
```diff
@@ -424,8 +424,9 @@ static void __attribute__((noinline)) wolfBoot_image_clear_signature_ok(
         asm volatile("mov r0, #50":::"r0"); \
         asm volatile("mov r0, #50":::"r0"); \
         asm volatile("mov r0, #50":::"r0"); \
-        compare_res = XMEMCMP(digest, img->sha_hash, WOLFBOOT_SHA_DIGEST_SIZE); \
-        /* Redundant checks that ensure the function actually returned 0 */ \
+        compare_res = image_CT_compare(digest, img->sha_hash, \
```
dgarske, line 427: **🔴 [Critical] RSA_VERIFY_HASH armored inline asm checks r0==0 but image_CT_compare returns 1 on match**
*🚫 BLOCK*  `bug`

The `RSA_VERIFY_HASH` macro was changed from `XMEMCMP` (returns 0 on match) to `image_CT_compare` (returns 1 on match, confirmed in src/image.c:71 `return diff == 0`). However, the inline ARM assembly was **not updated** to match the new return value semantics. The asm does `cmp r0, #0` / `bne hnope` which means "if r0 != 0, jump to failure." Since `image_CT_compare` returns 1 (non-zero) on a successful match, the asm will ALWAYS branch to `hnope` on a valid signature, causing RSA hash verification to always fail on `WOLFBOOT_ARMORED` builds. The comment was updated to say "Redundant checks that ensure the function actually returned 1" but the assembly operand is still `#0`. This inverts the security check: valid images are always rejected.

**Suggestion:**
```suggestion
Change all `cmp r0, #0` to `cmp r0, #1` throughout the RSA_VERIFY_HASH macro (both call sites, all 8 redundant check blocks). This makes the `bne hnope` branch when r0 != 1 (i.e., mismatch), correctly rejecting only non-matching hashes.
```

src/libwolfboot.c
```diff
@@ -1667,8 +1695,9 @@ int RAMFUNCTION wolfBoot_erase_encrypt_key(void)
     mem -= (sel_sec * WOLFBOOT_SECTOR_SIZE);
 #endif
     XMEMSET(ff, FLASH_BYTE_ERASED, ENCRYPT_KEY_SIZE + ENCRYPT_NONCE_SIZE);
-    if (XMEMCMP(mem, ff, ENCRYPT_KEY_SIZE + ENCRYPT_NONCE_SIZE) != 0)
-        hal_set_key(ff, ff + ENCRYPT_KEY_SIZE);
+    if (!encrypt_key_is_erased(mem, ENCRYPT_KEY_SIZE + ENCRYPT_NONCE_SIZE))
+        ret = hal_set_key(ff, ff + ENCRYPT_KEY_SIZE);
+    return ret;
```
dgarske, line 1700: **🟡 [Medium] wolfBoot_erase_encrypt_key has unreachable return 0 after conditional return ret**
*💡 SUGGEST*  `bug`

In the `#else` branch (non-TSIP, non-MMU), the function now has `return ret;` at line 1700, followed by `#endif` and then `return 0;` at line 1702. The `return 0` at line 1702 is unreachable when this preprocessor branch is active. In the TSIP and MMU branches it IS reachable and correct. The code compiles and works because preprocessor branches are mutually exclusive, but some compilers may warn about the unreachable code. The structure is inherited from the original code but is now more confusing because the `#else` path returns a potentially non-zero error code.

**Suggestion:**
```suggestion
Restructure so each preprocessor branch has its own return:
```c
#ifdef WOLFBOOT_RENESAS_TSIP
    return 0;
#elif defined(MMU)
    ForceZero(ENCRYPT_KEY, ENCRYPT_KEY_SIZE + ENCRYPT_NONCE_SIZE);
    return 0;
#else
    ...
    return ret;
#endif
```
```

File: tools/keytools/sign.c [hunk 11]
```diff
@@ -221,9 +221,23 @@ static void header_append_u16(uint8_t* header, uint32_t* idx, uint16_t tmp16)
...
+
 static void header_append_tag(uint8_t* header, uint32_t* idx, uint16_t tag,
-    uint16_t len, void* data)
+    uint16_t len, const void* data)
 {
+    const uint32_t append_sz = (uint32_t)(sizeof(tag) + sizeof(len)) + len;
+    const uint32_t header_sz = header_append_limit();
+
+    if ((*idx > header_sz) || (append_sz > (header_sz - *idx))) {
+        fprintf(stderr,
```
dgarske, line 234: **🟡 [Medium] header_append_tag overflow calls exit(1) without running failure cleanup**
*💡 SUGGEST*  `bug`

The new overflow check in `header_append_tag` calls `exit(1)` on detection. This bypasses the `failure:` cleanup label in `make_header_ex` which scrubs encryption key material via `wc_ForceZero(key, ...)` and closes open file handles. In practice, the overflow would be detected early in header construction before encryption keys are loaded, so the risk is minimal. However, the pattern is inconsistent with the PR's security-hardening goal.

**Suggestion:**
```suggestion
Consider returning an error code that causes the caller to `goto failure`, or document that this check always fires before sensitive material is loaded.
```

tools/keytools/sign.c
```diff
@@ -1107,6 +1143,118 @@ static int sign_digest(int sign, int hash_algo,
...
+        header_size_align_8(&idx);
+        header_size_append_tag(&idx, cert_chain_sz);
+    }
+
+    if (digest_sz > 0U) {
+        header_size_align_8(&idx);
+        header_size_append_tag(&idx, digest_sz);
+        header_size_align_8(&idx);
+
+        if (CMD.hybrid && secondary_key_sz > 0U) {
```
dgarske, line 1219: **🔵 [Low] header_required_size over-estimates for SHA384/SHA3 hybrid paths**
*💡 SUGGEST*  `bug`

In `header_required_size`, the hybrid secondary key section inserts `header_size_align_8` between `HDR_SECONDARY_CIPHER` and `HDR_SECONDARY_PUBKEY`. This matches the SHA256 path in `make_header_ex` (which has ALIGN_8), but NOT the SHA384 or SHA3 paths (which have no ALIGN_8 between them). The result is that `header_required_size` over-estimates the size for SHA384/SHA3 hybrid. This is safe (header won't overflow) but could cause unnecessary header enlargement when auto-sizing for certificate chains.

**Suggestion:**
```suggestion
Either add the missing ALIGN_8 to the SHA384/SHA3 hybrid paths in make_header_ex for consistency, or remove it from header_required_size to match the actual behavior.
```

src/update_flash.c
```diff
@@ -799,6 +833,7 @@ static int RAMFUNCTION wolfBoot_update(int fallback_allowed)
     uint32_t total_size = 0;
     const uint32_t sector_size = WOLFBOOT_SECTOR_SIZE;
     uint32_t sector = 0;
+    int ret = 0;
     /* we need to pre-set flag to SECT_FLAG_NEW in case magic hasn't been set
      * on the update partition as part of the delta update direction check. if
      * magic has not been set flag will have an un-determined value when we go
```
dgarske, line 839: **🔵 [Low] wolfBoot_get_total_size error message says 'too large' when condition is 'too small'**
*🔧 NIT*  `style`

The check `if (total_size <= IMAGE_HEADER_SIZE)` catches zero-size or invalid images where total_size is too small. But the error message says "Image total size %u too large!" which is misleading.

**Suggestion:**
```suggestion
Change to: `wolfBoot_printf("Image total size %u invalid!\n", total_size);`
```

tools/keytools/keygen.c
```diff
@@ -1131,16 +1238,26 @@ static void keygen_ml_dsa(const char *priv_fname, uint32_t id_mask)
...
-    wc_MlDsaKey_Free(&key);
-    free(priv);
-    priv = NULL;
+cleanup:
+    if (fpriv != NULL)
+        fclose(fpriv);
+    if (key_init)
+        wc_MlDsaKey_Free(&key);
+    if (priv != NULL) {
+        wc_ForceZero(priv, priv_len);
```
dgarske, line 1255: **🔵 [Low] keygen_ml_dsa: wc_ForceZero uses priv_len which may be 0 at cleanup**
*🔧 NIT*  `bug`

In cleanup, `wc_ForceZero(priv, priv_len)` uses `priv_len` which is initialized to the malloc size at line 1139 by `wc_MlDsaKey_ExportPrivRaw`. If failure occurs between malloc (line 1130) and the export call, `priv_len` would still be at its declared value of `(word32)ml_dsa_priv_len` (set at line 1139). Actually, `priv_len` is set at line 1137 to `(word32)ml_dsa_priv_len`. In the current code, no goto can fire between lines 1130-1137, but this is fragile.

**Suggestion:**
```suggestion
Use `(word32)ml_dsa_priv_len` instead of `priv_len` for the size parameter, since that's the allocation size.
```

tools/keytools/sign.c
```diff
@@ -296,6 +310,28 @@ static struct cmd_options CMD = {
     .hybrid = 0
 };
 
+static uint32_t header_append_limit(void)
```
dgarske, line 313: **🔵 [Low] zero_and_free uses hand-rolled volatile loop instead of wc_ForceZero**
*🔧 NIT*  `convention`

The `zero_and_free` function implements its own volatile-pointer zeroing loop while keygen.c in the same PR uses `wc_ForceZero` (wolfCrypt's tested implementation) for the same purpose.

**Suggestion:**
```suggestion
Replace with `wc_ForceZero(buf, len); free(buf);`
```

tools/keytools/sign.c
```diff
@@ -1133,21 +1283,18 @@ static int make_header_ex(int is_diff, uint8_t *pubkey, uint32_t pubkey_sz,
...
-            const uint32_t max_alignment = 8;
-            /* Required space = tag(2) + length(2) + data + potential alignment
-             * * padding */
-            const uint32_t required_space =
-                tag_len_size + file_stat.st_size + max_alignment;
+            const uint32_t required_space = header_required_size(is_diff,
+                (uint32_t)file_stat.st_size, secondary_key_sz);
 
             /* If the current header size is too small, increase it */
             if (CMD.header_sz < required_space) {
```
dgarske, line 1295: **🔵 [Low] sign.c stale f handle after fclose without NULL assignment**
*🔧 NIT*  `bug`

At line 1337, `fclose(f)` is called without setting `f = NULL`. The PR adds a `failure:` cleanup label that does `if (f) fclose(f)` at line 2114. Currently no `goto failure` can be reached between line 1337 and the next `f = fopen(...)`, so no double-close occurs. But this pattern is fragile for future changes. Several other `fclose(f)` calls (lines 1478, 1569, 1644, 1717, 1745, 1782, etc.) also don't NULL `f`.

**Suggestion:**
```suggestion
Add `f = NULL;` after every `fclose(f)` call in make_header_ex, or at minimum after the intermediate closes.
```

src/image.c
```diff
@@ -58,8 +58,8 @@
 /* Globals */
 static uint8_t digest[WOLFBOOT_SHA_DIGEST_SIZE] XALIGNED(4);
 
-static int image_CT_compare(const uint8_t *expected, const uint8_t *actual,
-    uint32_t len)
+int __attribute__((noinline)) image_CT_compare(
```
dgarske: **⚪ [Info] image_CT_compare uses __attribute__((noinline)) without portability guard**
*🔧 NIT*  `convention`

The function uses `__attribute__((noinline))` directly, which is GCC/Clang-specific. If non-GCC compilers (IAR, MSVC, CCRX) are supported for this file, this syntax may not compile. The codebase already uses some portable macros for attributes.

**Suggestion:**
```suggestion
Use a portable macro wrapper if non-GCC compilers need to compile this file.
```

dgarske (changes requested):
## 🐺 Skoll Code Review

**Overall recommendation:** `REQUEST_CHANGES`
**Findings:** 12 total — 12 posted, 1 skipped

### Posted findings
- [Critical] **RSA_VERIFY_HASH armored inline asm checks r0==0 but image_CT_compare returns 1 on match** — `include/image.h:427-466`
- [Medium] **wolfBoot_erase_encrypt_key has unreachable return 0 after conditional return ret** — `src/libwolfboot.c:1700-1702`
- [Medium] **image_CT_compare and wolfBoot_constant_compare have opposite return conventions** — `src/image.c:61, src/tpm.c:47`
- [Medium] **header_append_tag overflow calls exit(1) without running failure cleanup** — `tools/keytools/sign.c:234`
- [Low] **header_required_size over-estimates for SHA384/SHA3 hybrid paths** — `tools/keytools/sign.c:1225-1229`
- [Low] **keygen_ed25519/ed448 do not close fpriv in cleanup path** — `tools/keytools/keygen.c:782,838`
- [Low] **wolfBoot_local_constant_compare duplicated in three files** — `src/update_flash.c:45, src/x86/ahci.c:112`
- [Low] **wolfBoot_get_total_size error message says 'too large' when condition is 'too small'** — `src/update_flash.c:906`
- [Low] **keygen_ml_dsa: wc_ForceZero uses priv_len which may be 0 at cleanup** — `tools/keytools/keygen.c:1255`
- [Low] **zero_and_free uses hand-rolled volatile loop instead of wc_ForceZero** — `tools/keytools/sign.c:313`
- [Low] **sign.c stale f handle after fclose without NULL assignment** — `tools/keytools/sign.c:1337`
- [Info] **image_CT_compare uses __attribute__((noinline)) without portability guard** — `src/image.c:61`

<details><summary>Skipped findings</summary>

- [Low] keygen_ed25519/ed448 do not close fpriv in cleanup path

</details>

*Review generated by Skoll via openclaw*

src/libwolfboot.c
danielinux: not a bug: the final return is still reachable in TSIP/MMU builds; only the non-TSIP/non-MMU branch returns earlier

tools/keytools/sign.c
danielinux: all calls to `header_append_tag` are currently happening before allocating resources/keys or opening files. No sensitive cleanup is bypassed here.
No fix

tools/keytools/keygen.c
danielinux: priv_len is initialized immediately after successful allocation and before any cleanup path can execute, and cleanup zeroizes only when priv != NULL.

invalid claim

tools/keytools/sign.c
danielinux: switching zero-and-free to the available wc_ForceZero instead of reimplementing twice.

dgarske (approved):

dgarske (approved):

PR #746, opened 2026-04-10
NXP T2080 / CW VPX3-152: VxWorks 7 64-bit boot support

dgarske:
# NXP T2080 / CW VPX3-152: VxWorks 7 64-bit boot support

Adds wolfBoot support for booting **VxWorks 7 SMP 64-bit** (and signed ELF images) on the NXP T2080 (quad e6500) / Curtiss-Wright VPX3-152, verified booting to the VxWorks shell (`CPU Count: 8`, `Target Name: vxTarget`).

## Key fix

The e6500 cluster L2 cache must be brought up in a specific order: set **`L2PE` (ECC) in its own write and poll it set BEFORE enabling `L2E`**, with `L2FI|L2LFC` on the flash-invalidate (the CW/SDK2.0 sequence). wolfBoot previously wrote a bare `L2FI` then `L2E` alone, which left the L2 ECC array uninitialized for the kernel's `0x1E0000` set; once ECC checking became active the first instruction fetch there raised an uncorrectable multi-bit ECC machine check (`MCSR[IF]=0x10000`, `L2ERRDET=0x80000080`) and VxWorks hung at the MC vector. With the corrected order wolfBoot runs L2 ECC on throughout and the OS inherits a clean, ECC-valid L2.

## Fixes

- e6500 cluster-L2 ECC bring-up ordering (the root cause above), in `src/boot_ppc_start.S`.
- ELF in-place loader staging-overlap: the staging buffer overlapped the ELF load span and the collide-guard silently dropped the segment holding the entry point; staging moved above the segment span + a loud error on overlap.
- 64-bit OS hand-off correctness: LAW/TLB transition to 36-bit physical, IVOR exception vectoring, and MSR state aligned with the CW U-Boot `bootm` profile.

## Features

- Boot VxWorks 7 SMP 64-bit via the ePAPR spin-table protocol, releasing all four cores (CPU Count 8).
- Boot signed ELF images directly (in-place PT_LOAD scatter loader; no uImage repack needed).
- DPAA bring-up: LIODN programming + QMan/BMan software-portal init (required before a DPAA-aware OS).
- Board configs for CW VPX3-152 (256 MB NOR @ 0xF0000000, 4 GB DDR3L) and NAII 68PPC2.

## Improvements

- PPC32 `printf("%lld")` support (`PRINTF_LONG_LONG` auto-enabled for the PPC toolchain) in `src/string.c`.
- CI: a T2080 build entry added to `.github/workflows/test-configs.yml`; T2080 build rules in `arch.mk`; T2080 target docs in `docs/Targets.md`.

## New build options

- `BOARD_CW_VPX3152` / `BOARD_NAII_68PPC2` -- select the T2080 board variant (addresses, oscillator, DDR). Default is the T2080 RDB.
- `OS_64BIT=1` -- 64-bit OS hand-off (ePAPR/VxWorks 7 64-bit); MSR[CM]=0 at entry, OS self-promotes.
- `ELF=1` -- enable the in-place ELF image loader.
- `WOLFBOOT_NO_DPAA` -- disable DPAA/QMan/BMan init (define to opt out; default is on for T2080).
- `WOLFBOOT_EARLY_UART` -- compile in the e6500 early-boot UART debug helpers (off by default; OS-porting bring-up aid).

## Testing

Cold-boots VxWorks 7 reliably to the SMP shell. No regression to the existing T2080 RDB / NAII paths (board behavior is gated on `BOARD_CW_VPX3152`).

### DIFF
--- a/hal/nxp_ppc.h
+++ b/hal/nxp_ppc.h
@@ -84,7 +84,7 @@
         #define ENABLE_L2_CACHE /* setup and enable L2 in first stage only */
     #else
         /* relocate to 64-bit 0xF_ */
-        #define CCSRBAR_PHYS_HIGH 0xFULL
+        #define CCSRBAR_PHYS_HIGH 0xF
         #define CCSRBAR_PHYS (CCSRBAR_PHYS_HIGH + CCSRBAR_DEF)
     #endif
 
@@ -95,9 +95,9 @@
 
     #define FLASH_BASE_ADDR      0xEC000000UL
     #ifndef BUILD_LOADER_STAGE1
-    #define FLASH_BASE_PHYS_HIGH 0xFULL /* 36-bit: 0xF_EC000000 */
+    #define FLASH_BASE_PHYS_HIGH 0xF /* 36-bit: 0xF_EC000000 */
     #else
-    #define FLASH_BASE_PHYS_HIGH 0x0ULL /* 32-bit stage1 */
+    #define FLASH_BASE_PHYS_HIGH 0x0 /* 32-bit stage1 */
     #endif
     #define FLASH_LAW_SIZE       LAW_SIZE_64MB
     #define FLASH_TLB_PAGESZ     BOOKE_PAGESZ_64M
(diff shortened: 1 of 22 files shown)


hal/nxp_ppc.h
```diff
@@ -84,7 +84,7 @@
         #define ENABLE_L2_CACHE /* setup and enable L2 in first stage only */
     #else
         /* relocate to 64-bit 0xF_ */
-        #define CCSRBAR_PHYS_HIGH 0xFULL
+        #define CCSRBAR_PHYS_HIGH 0xF
```
danielinux, line 87: Multilple unsigned 64bit constants seem to have lost their ULL suffix here

danielinux (approved):

PR #757, opened 2026-04-22
lpc55s69 hw crypto

twcook86:


danielinux (changes requested):
do not tie RNG to `NO_HWACCEL`. CASPER/HASHCRYPT acceleration can be disabled while the hardware RNG is still required by WOLFCRYPT_SECURE_MODE, e.g. when compiling with trustzone.

You still need something like:
```
  ifeq ($(WOLFCRYPT_TZ),1)
    OBJS+=$(MCUXPRESSO)/drivers/rng_1/fsl_rng.o
  endif
```

when PKA=0.

Then in hal/lpc55s69.c, you should remove the WOLFSSL_NXP_LPC55S69_NO_HWACCEL guards around RNG_Init() and RNG_GetRandomData(), or replace them with a dedicated RNG macro if you need one

twcook86:
> do not tie RNG to `NO_HWACCEL`. CASPER/HASHCRYPT acceleration can be disabled while the hardware RNG is still required by WOLFCRYPT_SECURE_MODE, e.g. when compiling with trustzone.
> 
> You still need something like:
> 
> ```
>   ifeq ($(WOLFCRYPT_TZ),1)
>     OBJS+=$(MCUXPRESSO)/drivers/rng_1/fsl_rng.o
>   endif
> ```
> 
> when PKA=0.
> 
> Then in hal/lpc55s69.c, you should remove the WOLFSSL_NXP_LPC55S69_NO_HWACCEL guards around RNG_Init() and RNG_GetRandomData(), or replace them with a dedicated RNG macro if you need one

reworked the rng code, clleaned up various other areas.  I think all the pr issues should be addressed at this point.

danielinux (approved):

danielinux:
Thanks @twcook86 -- merging.

PR #761, opened 2026-04-28
support for "aux" algorithms

bigbrett:
Adds support for compiling in multiple auxiliary algorithms beyond those required by `SIGN` and `HASH`.

This is primarily intended for certificate chain verification via wolfHSM, where all potential algorithms in the chain must be compiled in, however, the implementation is generic and can support other use cases in the future.

Also updates the dummy certificate chain generation script to:

* Support auxiliary algorithms
* Enable expanded testing for mixed certificate chains and image algorithms


danielinux:
This looks like a good idea, but it should be merged with other engines where we do the same, i.e. configurations that require a broader subset of algorithms than those used by wolfBoot. One example is PKCS11 algorithm selection, and perhaps also with hybrid verification (SECONDARY_ ...), to ensure that mixing/matching does not define things twice. user_settings.h is getting crowded and perhaps this could be an opportunity to simplify, and offload some preselection to .mk (aux.mk ? being included/activated in 'engine' configurations?) I'd like to see things converge in one direction because algorithm selection can be done in too many ways already.

PR #763, opened 2026-04-29
FIT: gzip-compressed kernel + ramdisk (initramfs) support

dgarske:
## Summary

Adds an in-tree DEFLATE/gzip inflater and wires it into the FIT image path so
wolfBoot can boot compressed Linux kernels and ramdisks directly out of a FIT
container, with per-subimage hash verification on top of the outer wolfBoot
signature. Also adds initramfs (ramdisk) extraction and DTB `/chosen` fixup so
the kernel can find the loaded ramdisk.

Default-on (`GZIP=1`) for the FIT-using example targets: ZynqMP (eMMC + SD),
Versal VMK180 (eMMC + SD), PolarFire MPFS250 (SD + QSPI). Builds on the
existing master configs are byte-identical to before this branch.

### gzip inflater (`src/gzip.c`, `include/gzip.h`)
- Clean-room RFC 1951 (DEFLATE) + RFC 1952 (gzip wrapper) implementation.
- Single pass, no heap allocations — output buffer doubles as the LZ77 window.
- CRC32 + ISIZE trailer verified; gated by `WOLFBOOT_GZIP`.
- wolfBoot C style: declarations at top of function, single return per
  function, `/* … */` comments, every callee return code checked.

### FIT integration (`src/fdt.c`, `include/fdt.h`)
- New `fit_load_image_ex(out_max)`; `fit_load_image()` kept as a wrapper so
  existing callers don't change.
- When `WOLFBOOT_GZIP` is on and a subimage carries `compression = "gzip"`,
  the inflater writes straight to the FIT-declared load address.
- After load (compressed or not), `fit_verify_hash` checks the `hash-1`
  subnode against the loaded bytes. SHA-256 / SHA-384 return codes from
  `wc_InitSha*`, `wc_Sha*Update`, `wc_Sha*Final` are now propagated — a
  misbehaving crypto backend can't silently degrade to a no-op.

### Ramdisk / initramfs support
- `fit_find_images()` gains a ramdisk out-arg.
- New `fdt_fixup_initrd()` writes `/chosen/linux,initrd-start` and
  `linux,initrd-end` as 64-bit big-endian cells.
- `src/update_disk.c` and `src/update_ram.c` load the FIT ramdisk node
  (under `WOLFBOOT_FIT_RAMDISK`) and patch the DTB.
- Compressed ramdisks reuse the same `fit_load_image_ex()` decompress path.
- `RAMDISK=1` build switch + `WOLFBOOT_LOAD_RAMDISK_ADDRESS` plumbed
  through `tools/config.mk` → `Makefile` sed → `include/target.h.in`.
  `WOLFBOOT_LOAD_RAMDISK_ADDRESS=0` (the default) keeps the ramdisk
  wherever `fit_load_image` returned it (FIT load addr or in-FIT pointer).
- `hal/zynq.c` and `hal/versal.c` `fdt_totalsize` headroom bumped
  512 → 768 bytes to fit the new `linux,initrd-{start,end}` entries.

### Example configs
- New `config/examples/zynqmp_sdcard_ramdisk.config`
  (load address `0x40000000`).
- `GZIP=1` flipped on by default in: `zynqmp.config`,
  `zynqmp_sdcard.config`, `polarfire_mpfs250.config`,
  `polarfire_mpfs250_qspi.config`, `versal_vmk180.config`,
  `versal_vmk180_sdcard.config`. Pass `GZIP=0` to opt out and keep the
  pre-branch host-side gzip workflow.

### Unit tests (`tools/unit-tests/unit-gzip.c`)
- 6 round-trip corpora (empty, short text, all-zeros, structured text,
  pseudo-random, ~2 MB kernel-sized) — host `gzip(1)` → `wolfBoot_gunzip`.
- 9 negative cases: bad magic, bad CM, reserved FLG bits, truncated
  header, truncated DEFLATE body, CRC32 mismatch, ISIZE mismatch, output
  overflow, NULL parameters.
- All 15 cases pass under libcheck.

### Docs (`docs/Targets.md`)
- Versal PetaLinux walkthrough now documents both `GZIP=1` (default,
  `compression="gzip"` + `Image.gz`) and `GZIP=0` (plain `Image`) flows.
- New ZynqMP "Booting PetaLinux" subsection covering the same mechanism
  plus the `RAMDISK=1` / `WOLFBOOT_LOAD_RAMDISK_ADDRESS` setup and a
  sample ITS layout for kernel + DTB + initramfs.
- PolarFire MPFS250 SD-card section now describes Option A (default,
  `compression="gzip"`, mkimage builds the FIT directly) and Option B
  (host-side `gzip -cdvk`, `compression="none"`).

## Notes for reviewers
- The gzip inflater is the only new crypto-adjacent code path; defense in
  depth is provided by the existing wolfBoot signature on the outer FIT
  plus the per-subimage `hash-1` verify. CRC32 in the gzip trailer is a
  data-integrity check, not a security boundary.
- No public API signatures changed. `fit_load_image` keeps its old
  prototype; the new behavior is reached via `fit_load_image_ex`.
- `WOLFBOOT_LOAD_RAMDISK_ADDRESS=0` is the safe default — when unset the
  ramdisk stays wherever the FIT/loader put it.


docs/Targets.md
```diff
@@ -3390,6 +3428,103 @@ Application running successfully!
...
+```sh
+cp config/examples/zynqmp_sdcard.config .config
+# Uncomment the RAMDISK / WOLFBOOT_LOAD_RAMDISK_ADDRESS / LINUX_BOOTARGS
+# block under "Optional: FIT-bundled initramfs" and comment out the
+# LINUX_BOOTARGS_ROOT line above it.
+make
+```
+
+Key options (in `config/examples/zynqmp_sdcard.config`):
+- `RAMDISK=1` - enables FIT ramdisk extraction (`-DWOLFBOOT_FIT_RAMDISK`).
```
AlexLanzano: I think we should rename `RAMDISK` to `FIT_RAMDISK` in just to make it explicit that we're dealing with a FIT ramdisk image instead of a raw ramdisk image.

Maybe a more explicit name for the `RAMDISK` option? Like `EXTRACT_FIT_RAMDISK`?

docs/Targets.md
```diff
@@ -3390,6 +3428,103 @@ Application running successfully!
 Entering idle loop...
 ```
 
+**Booting PetaLinux**
```
AlexLanzano: We may want to rename this to `Booting Linux via FIT image`. FIT images are more a U-Boot thing than Petalinux.

On a side note: Might be good to mention somewhere that wolfBoot can be a drop in replacement for U-Boot in terms of loading and booting FIT images.

hal/versal.c
```diff
@@ -1276,8 +1276,9 @@ int hal_dts_fixup(void* dts_addr)
     wolfBoot_printf("FDT: Version %d, Size %d\n",
         fdt_version(fdt), fdt_totalsize(fdt));
 
-    /* Expand total size to allow adding/modifying properties */
-    fdt_set_totalsize(fdt, fdt_totalsize(fdt) + 512);
+    /* Expand total size to allow adding/modifying properties (bootargs and,
+     * when WOLFBOOT_FIT_RAMDISK is in play, linux,initrd-{start,end}). */
+    fdt_set_totalsize(fdt, fdt_totalsize(fdt) + 768);
```
AlexLanzano: Could we make that hardcoded value configurable? If not we should name it at least

hal/zynq.c
```diff
@@ -1927,11 +1927,12 @@ int hal_dts_fixup(void* dts_addr)
...
-     * guaranteed by the load-address layout (DTB at WOLFBOOT_LOAD_DTS_ADDRESS,
-     * kernel loaded much higher), so growing the header is safe. Matches
-     * the pattern used in hal/versal.c:hal_dts_fixup. */
-    fdt_set_totalsize(fdt, fdt_totalsize(fdt) + 512);
+     * a new/larger bootargs property and (when WOLFBOOT_FIT_RAMDISK is in
+     * play) the linux,initrd-{start,end} properties. Physical headroom is
+     * already guaranteed by the load-address layout (DTB at
+     * WOLFBOOT_LOAD_DTS_ADDRESS, kernel loaded much higher), so growing
+     * the header is safe. Matches the pattern in hal/versal.c. */
+    fdt_set_totalsize(fdt, fdt_totalsize(fdt) + 768);
```
AlexLanzano: Could we make that hardcoded value configurable? If not we should name it at least

AlexLanzano (dismissed):
Looks great overall. Just some minor naming considerations.

danielinux (commented):
Since the FIT image is verified when it's still compressed (which is good, because we don't want to parse unverified FITs), there's perhaps a potential TOCTOU (verify/uncompress/load/stage) because there's no verification of the uncompressed image afterwards.

I would suggest considering to add one extra check (at least for integrity/expected sha) after decompression on the final payload.

include/gzip.h
```diff
@@ -0,0 +1,116 @@
...
+ *   sym 16: 2 extra bits, repeat previous length 3..6 times
+ *   sym 17: 3 extra bits, repeat zero        3..10 times
+ *   sym 18: 7 extra bits, repeat zero       11..138 times
+ */
+#define GZIP_REPEAT_PREV_EXTRA    2
+#define GZIP_REPEAT_PREV_BASE     3
+#define GZIP_REPEAT_Z3_EXTRA      3
+#define GZIP_REPEAT_Z3_BASE       3
+#define GZIP_REPEAT_Z7_EXTRA      7
+#define GZIP_REPEAT_Z7_BASE       11
```
rizlik: NIT: these are only used by gzip and don't use a WOLF prefix. Consider moving them in gzip.c

src/update_disk.c
```diff
@@ -560,6 +560,45 @@ void RAMFUNCTION wolfBoot_start(void)
...
+                        (uint64_t)(uintptr_t)rd_dst, (uint64_t)rd_size);
+                }
+            }
+            else {
+                wolfBoot_printf("FIT: ramdisk node present but load failed\n");
+            }
+        }
+#else
+        (void)ramdisk;
+#endif
```
rizlik, line 575: NIT: can we refactor out this into an helper to avoid duplication between update_disk and update_ram?

rizlik (dismissed):
Very neat gzip implementation!
I just left two NITs, but feel free to ignore them.

On the TOCTOU attack:

Both this and current implementation allow an attacker who can modify the FIT content after the verification but before parsing/loading to bypass any authentication check.
It would be useful to defend against this scenario, for example when loading from external flash where an attacker can control the SPI lines but cannot modify internal RAM.
To prevent this wolfBoot should copy the FIT image in RAM first, then verify, parse and load only using the copy.
This trades-off with RAM usage, that now need to store a full extra copy of the FIT.
I do think it would be a nice opt-in feature, but outside of this PR.

rizlik (approved):

danielinux (approved):

PR #765, opened 2026-05-01
Fix lms/xmss header includes.

twcook86:
Remove outdated includes.

wolfSSL PR #10292 broke the wolfBoot build, though it's really a wolfBoot bug


danielinux (approved):

danielinux:
Also hit while testing with latest wolfssl. Merging. Thanks!

PR #769, opened 2026-05-01
Add WOLFCRYPT_TZ_WOLFHSM TrustZone engine for STM32H5

aidangarske:
# Description
  - Adds `WOLFCRYPT_TZ_WOLFHSM=1` a fourth TrustZone engine for STM32H5
    alongside PKCS11, PSA, and fwTPM. New config:                                                                                                          
    `config/examples/stm32h5-tz-wolfhsm.config`.
  - Hosts a wolfHSM server in the secure boot image and exposes it to                                                                                      
    the non-secure application through a single NSC veneer                                                                                                 
    (`wcs_wolfhsm_transmit`). The non-secure app uses the standard
    wolfCrypt API; operations through `WH_DEV_ID` route to the secure                                                                                      
    side. Keys never leave the secure world.                                                                                                               
  - Persistent flash-backed keystore via a new `whFlashCb` adapter                                                                                         
    (`src/wolfhsm_flash_hal.c`) over `hal_flash_*`, mirroring                                                                                              
    `psa_store.c`'s sector-cached read-modify-erase-write pattern.
    Two-partition journaling in the existing keyvault region.                                                                                              
  - Mutually exclusive with PKCS11/PSA/fwTPM. Shared
    `WOLFHSM_SERVER_OBJS` / `WOLFHSM_CLIENT_OBJS` variables avoid                                                                                          
    object-list duplication with the existing AURIX/sim wolfHSM lanes.
  - New CI lane in `.github/workflows/trustzone-emulator-tests.yml`:                                                                                       
    m33mu first-boot/second-boot persistence test asserting CommInit,
    RNG, SHA256, AES, and keystore restore across reset.                                                                                                   
  - New host unit test `tools/unit-tests/unit-wolfhsm_flash_hal.c`
    covers bounds, alignment, multi-sector, and write-failure paths.                                                                                       
  - Docs: new `docs/wolfHSM.md` + wolfHSM section in `docs/STM32-TZ.md`.                                                                                   
                                                                                                                                                           
  ## Notes                                                                                                                                                 
  - `lib/wolfHSM` submodule pin is not bumped in this PR. The wolfHSM
    PR (`port/stmicro/stm32-tz` NSC bridge transport) lands first; the                                                                                     
    pin bump will follow in a small standalone PR.
  - [x] Update sub module pin to 384 wolfHSM PR                                                                                                        
                                                                                                                                                            
  ## Test plan                                                                                                                                             
  - [x] m33mu first/second boot persistence (CI)
  - [x] Host unit test for flash adapter (CI)                                                                                                              
  - [x] On-board STM32H5 NUCLEO-H563ZI: `WOLFHSM_TZ_TEST_PASS` over UART                                                                                   
  - [x] On-board first/second boot persistence worked
  - [x] PKCS11/PSA/fwTPM configs still build and pass m33mu
  - [x] Skoll review passes


tools/unit-tests/unit-wolfhsm_flash_hal.c
```diff
@@ -0,0 +1,312 @@
...
+    if (addr < MOCK_FLASH_BASE ||
+        addr + (uintptr_t)len > MOCK_FLASH_BASE + MOCK_FLASH_SIZE) {
+        return -1;
+    }
+    memcpy((void *)addr, data, (size_t)len);
+    return 0;
+}
+
+#include "../../src/wolfhsm_flash_hal.c"
+
```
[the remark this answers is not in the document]
danielinux: Please ignore this. Including .c in unit tests is OK

src/wolfhsm_callable.c
```diff
@@ -0,0 +1,193 @@
...
+    if (g_wolfhsm_ready) {
+        return;
+    }
+
+    memset(&g_srv_tx_ctx, 0, sizeof(g_srv_tx_ctx));
+
+    flash_cfg.base           = (uint32_t)&_flash_keyvault;
+    flash_cfg.size           = (uint32_t)&_flash_keyvault_size;
+    flash_cfg.partition_size = WCS_WOLFHSM_PARTITION_SIZE;
+    g_nvm_flash_cfg.config   = &flash_cfg;
```
danielinux: This is storing a temporary stack value (flash_cfg, declared above) into the global structure.

When `wcs_wolfhsm_init()` returns, this value becomes a dangling pointer to future stack values.

src/wolfhsm_callable.c
```diff
@@ -0,0 +1,193 @@
...
+    if (rsp_capacity == 0U || rsp_capacity > WH_COMM_MTU) {
+        *rspSz = 0;
+        return WH_ERROR_BADARGS;
+    }
+    if (!g_wolfhsm_ready) {
+        *rspSz = 0;
+        return WH_ERROR_NOTREADY;
+    }
+
+    g_srv_tx_ctx.req_buf         = cmd;
```
danielinux, line 177: `cmd` is provided by the caller. It should be validated using cmse_check_address_range. If not checked, the non-secure app is compromised, it could send forged to pass a secure-side pointer here.

options.mk
```diff
@@ -935,12 +974,24 @@ ifeq ($(WOLFCRYPT_TZ_PKCS11),1)
...
   ifeq ($(WOLFCRYPT_TZ_FWTPM),1)
     $(error WOLFCRYPT_TZ_PSA and WOLFCRYPT_TZ_FWTPM are mutually exclusive)
   endif
+  ifeq ($(WOLFCRYPT_TZ_WOLFHSM),1)
+    $(error WOLFCRYPT_TZ_PSA and WOLFCRYPT_TZ_WOLFHSM are mutually exclusive)
+  endif
+endif
+
+ifeq ($(WOLFCRYPT_TZ_FWTPM),1)
+  ifeq ($(WOLFCRYPT_TZ_WOLFHSM),1)
```
danielinux, line 1033: Should also add guard against invalid combo like WOLFCRYPT_TZ_PKCS11 or WOLFCRYPT_TZ_PSA

.github/workflows/trustzone-emulator-tests.yml
```diff
@@ -119,6 +119,12 @@ jobs:
           grep -q "\\[BKPT\\] imm=0x7f" /tmp/m33mu-fwtpm.log
           grep -q "\\[EXPECT BKPT\\] Success" /tmp/m33mu-fwtpm.log
 
+      # The wolfHSM STM32H5 TrustZone integration is exercised by the
+      # nightly cross-repo job in wolfHSM
+      # (.github/workflows/wolfboot-tz-integration.yml), which builds
+      # this demo against the latest wolfHSM main and runs it under
+      # m33mu. It is intentionally not gated on every wolfBoot PR.
+
```
danielinux, line 127: All other TZ engines have m33mu tests: I think we should add one for TZ_WOLFHSM  here too.

src/wolfhsm_callable.c
```diff
@@ -0,0 +1,193 @@
...
+    .comm_config = &g_comm_cfg,
+    .nvm         = &g_nvm_ctx,
+    .crypto      = &g_crypto_ctx,
+#ifdef WOLF_CRYPTO_CB
+    .devId       = INVALID_DEVID,
+#endif
+};
+
+static whServerContext           g_server;
+static int                       g_wolfhsm_ready;
```
danielinux: `volatile`

port/stmicro/stm32h5-tz-wolfhsm/out/manifest.env
```diff
@@ -0,0 +1,3 @@
+BOOT_ADDR=0x08060000
```
danielinux: this file has been probably added by accident. Add the 'out' directory to `.gitignore`

config/examples/stm32h5-tz-wolfhsm.config
```diff
@@ -0,0 +1,35 @@
...
+TZEN?=1
+TARGET?=stm32h5
+SIGN?=ECC256
+HASH?=SHA256
+DEBUG?=0
+VTOR?=1
+CORTEX_M0?=0
+CORTEX_M33?=1
+NO_ASM?=0
+NO_MPU=1
```
danielinux, line 14: NO_MPU set with a TZ setup means that also apps in NS domain will not be able to use it. Please remove NO_MPU=1 if not strictly needed for some reason

danielinux (changes requested):

.github/workflows/trustzone-emulator-tests.yml
aidangarske: @bigbrett and I initially wanted to do that but the runs where taking 10-15 mins in CI to run not including the build so we moved it to wolfHSM and it builds nightly with the masters of each wolfboot and wolfHSM. On failure it opens a tracking issue assigned to wolfSSL-Bot in wolfHSM.

src/wolfhsm_callable.c
aidangarske: Moved the flash config to a file-scope static (g_flash_h5_cfg) so the pointer stays valid after wcs_wolfhsm_init() returns and dropped the stack local.

.github/workflows/trustzone-emulator-tests.yml
danielinux: ACK, my apologies. I was missing the test is already there
https://github.com/wolfSSL/wolfHSM/pull/348/

danielinux (dismissed):

docs/wolfHSM.md
```diff
@@ -239,3 +240,51 @@ When using wolfHSM server mode, no external server is required. wolfBoot include
...
+This is a separate deployment shape from the wolfHSM client/server modes documented above; it does not use `WOLFBOOT_ENABLE_WOLFHSM_CLIENT/SERVER` or the `hsmClientCtx`/`hsmServerCtx` HAL hooks, and is mutually exclusive with the other STM32H5 TrustZone engines (`WOLFCRYPT_TZ_PKCS11`, `WOLFCRYPT_TZ_PSA`, `WOLFCRYPT_TZ_FWTPM`).
+
+### Build
+
+```sh
+cp config/examples/stm32h5-tz-wolfhsm.config .config
+make
+```
+
+For on-board hardware testing, add `WOLFBOOT_TZ_TEST_NO_BKPT=1` so the auto-test prints a UART pass/fail line and idles in `while (1)` instead of issuing `bkpt #0x7f` (which HardFaults on real silicon without a debugger):
```
bigbrett: why is the default option something that hard faults when no debugger is connected? Shouldn't it be the other way around?

options.mk
```diff
@@ -1413,34 +1493,24 @@ ifeq ($(WOLFHSM_SERVER),1)
   WOLFCRYPT_OBJS += \
     $(WOLFBOOT_LIB_WOLFSSL)/wolfcrypt/src/cryptocb.o \
     $(WOLFBOOT_LIB_WOLFSSL)/wolfcrypt/src/coding.o \
-    $(WOLFBOOT_LIB_WOLFSSL)/wolfcrypt/src/random.o
+    $(WOLFBOOT_LIB_WOLFSSL)/wolfcrypt/src/random.o \
+    $(WOLFBOOT_LIB_WOLFSSL)/wolfcrypt/src/hmac.o \
+    $(WOLFBOOT_LIB_WOLFSSL)/wolfcrypt/src/aes.o \
```
bigbrett: why are we default adding `hmac` and `aes` to the object list for all wolfHSM server support if wolfBoot doesn't use/need them? Shouldn't this be gated on the wolfBoot feature that requires them?

bigbrett (commented):
@aidangarske a few questions

docs/wolfHSM.md
aidangarske: That is a great point thank you will fix

options.mk
aidangarske: I removed them thank you, I think I was needed them before we where doing the tests we are now. reverted

bigbrett (dismissed):

danielinux (dismissed):
Non-blocking issue:
The three cmse_check_address_range() calls use `CMSE_AU_NONSECURE`, whereas the sibling veneers fwtpm_callable.c and pkcs11_callable.c use `CMSE_NONSECURE` (= `CMSE_AU_NONSECURE | CMSE_MPU_NONSECURE`). 

All three correctly reject secure-world pointers, so this isn't a bug, also because `NO_MPU=1` . But once an NS MPU is ever configured, this veneer alone won't enforce NS-MPU permissions on the buffer. 

Note for later: consider aligning the flag across the three TZ veneers and add a one-line note on the chosen semantics so it's a deliberate decision rather than an accidental divergence.


Merging as is for now.

danielinux:
@aidangarske This has conflicts with master branch now: please rebase. Also perhaps consider fixing the CMSE_AU_NONSECURE / CMSE_NONSECURE minor remark above

danielinux (dismissed):

dgarske (approved):

PR #775, opened 2026-05-08
STM32 TrustZone improvements (when tested with wolfIP)

dgarske:


danielinux (changes requested):
I'm OK about setting all GPIO banks to non-secure for the purpose of the test application. The non-privilege Copilot "medium" comments make sense to me, I'd keep non-secure+privileged SRAM2 and SRAM3 as default for the OS to decide boundaries later. Please check again and consider Copilot's "medium" comment in stm32_tz.c line 234 to 243

danielinux (approved):

PR #787, opened 2026-06-03
Support for PUF in LPC55S69

twcook86:
This adds a new example config (lpc55s69-hwpuf.config) to wolfBoot, to demonstrate and test the new lpc55s69 hw puf support added in https://github.com/wolfSSL/wolfssl/pull/10584

It also adds an example of device attestation with lpc55s69-tz-psa.config


dgarske (commented):
I see this enabling the wolfCrypt HW PUF options, but I don't see anywhere it actually gets used in wolfBoot? Or even called in wolfCrypt. How's this supposed to work end to end?

danielinux (changes requested):
This adds the useful build plumbing for LPC55 PUF support, but it does not yet connect the HW PUF to the attestation identity.

What we actually need here is the HW PUF to become the UID/UDS source for the boot attestation demo. Today the attestation stack gets its identity through `hal_uds_derive_key()` / `hal_attestation_get_ueid()` in dice.c.

This PR still needs:
 - a new LPC55 attestation config enabling PSA attestation demo options
 - LPC55 hal support that rebuilds identity from PUF exposing it via hal_uds_derive_key / get_ueid. Check ~~existing generic software managed internal flash PUF support in STM32H5~~ existing support for HW PUF in mcxn.c via those API for how to implement
 - The test app showing that the PSA attestation token is actually produced from the PUF-derived identity wired inside wolfBoot.

twcook86:
> I see this enabling the wolfCrypt HW PUF options, but I don't see anywhere it actually gets used in wolfBoot? Or even called in wolfCrypt. How's this supposed to work end to end?

lpc55s69-hwpuf.config will compile in wolfssl test.c, which tests the hwpuf.  But, you have to turn on HWPUF=1 in that config file, and point your wolfssl to https://github.com/wolfSSL/wolfssl/pull/10584

danielinux:
Depends on https://github.com/wolfSSL/wolfssl/pull/10584

danielinux:
This is now properly demonstrating HW PUF on LPC55S69 in DICE.

LGTM after https://github.com/wolfssl/wolfssl/pull/10584 is merged.

PR #798, opened 2026-06-15
Wolfhsm no keystore

bigbrett:
## Main Change

wolfHSM client mode (and server mode with cert-chain verify) now builds without a local wolfBoot keystore. Public keys live on the HSM, referenced by key ID or resolved from a verified certificate chain. The old hybrid path that loaded keys from `keystore.c` and pushed them to the HSM as ephemeral keys is removed.

- **No keystore linked with wolfHSM**: `WOLFHSM_CLIENT=1` (or `WOLFHSM_SERVER=1` with   `CERT_CHAIN_VERIFY=1`) no longer compiles/links `src/keystore.o`. Defines  `WOLFBOOT_NO_KEYSTORE`, which skips the keystore permission-mask check and  pubkey-hash lookups since key authorization is enforced by the HSM.

## Other changes
- **Removed `WOLFBOOT_USE_WOLFHSM_PUBKEY_ID`** and the `WOLFHSM_CLIENT_LOCAL_KEYS`  make option. Clients now always use HSM-resident keys. This simplifies the `#if`spaghetti  in `src/image.c`.

- **Parameterizable client ID**: new `WOLFHSM_CLIENT_ID` make variable (default 1) replaces hardcoded IDs all of the various HALs.

- **Server mode requires cert chain**: `WOLFHSM_SERVER=1` without
  `CERT_CHAIN_VERIFY=1` now fails the build with a clear error (was always disallowed, now just added explicit error)

- **ML-DSA wolfHSM fixes**: added missing server-side ML-DSA key export and post-verify  key eviction for client and server cert-chain modes.

- **Doc updates**


src/image.c
```diff
@@ -747,9 +734,7 @@ static void wolfBoot_verify_signature_ml_dsa(uint8_t key_slot,
     wolfBoot_printf("info: ML-DSA %d verify_signature: pubkey %d, sig %d\n",
         ML_DSA_LEVEL, KEYSTORE_PUBKEY_SIZE, ML_DSA_IMAGE_SIGNATURE_SIZE);
 
-#if !defined WOLFBOOT_ENABLE_WOLFHSM_CLIENT || \
-    (defined WOLFBOOT_ENABLE_WOLFHSM_CLIENT && \
-     !defined(WOLFBOOT_USE_WOLFHSM_PUBKEY_ID))
+#if !defined(WOLFBOOT_ENABLE_WOLFHSM_CLIENT)
     pubkey = keystore_get_buffer(key_slot);
```
bigbrett, line 740: fixed

src/image.c
```diff
@@ -735,9 +724,8 @@ static void wolfBoot_verify_signature_ml_dsa(uint8_t key_slot,
 {
     int         ret = 0;
     wc_MlDsaKey ml_dsa;
-#if !defined WOLFBOOT_ENABLE_WOLFHSM_CLIENT || \
-    (defined WOLFBOOT_ENABLE_WOLFHSM_CLIENT && \
-     !defined(WOLFBOOT_USE_WOLFHSM_PUBKEY_ID))
+#if !defined(WOLFBOOT_ENABLE_WOLFHSM_CLIENT) && \
```
bigbrett, line 727: Server without cert-chain is not a supported use case currently, as mentioned in the docs. Added a build-time check to force this to fail for now.

src/image.c
```diff
@@ -2298,15 +2308,25 @@ int wolfBoot_verify_authenticity(struct wolfBoot_image *img)
             return -1;
         img->sha_hash = digest;
     }
-    key_mask = keystore_get_mask(key_slot);
     image_part = image_type & HDR_IMG_TYPE_PART_MASK;
+#ifdef WOLFBOOT_NO_KEYSTORE
```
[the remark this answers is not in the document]
bigbrett: by design, future work

options.mk
```diff
@@ -1410,6 +1418,13 @@ endif
 
 # wolfHSM server options
 ifeq ($(WOLFHSM_SERVER),1)
+  # Currently, cert chain auth is the only supported auth mechanism for wolfHSM
+  # server mode
+  ifeq ($(CERT_CHAIN_VERIFY),)
```
dgarske, line 1423: **⚪ [Info] $(error) continuation line embeds leading whitespace into the message**

The new server-mode guard uses a backslash line continuation inside `$(error ...)`. Make joins continuation lines literally, so the rendered error text becomes `...server mode             only supports certificate-chain authentication` with the next line's indentation embedded as a run of spaces. Functionally fine, just a slightly ragged message.

**Fix:** Put the message on a single line (or trim the continuation indentation) so the printed error reads cleanly.

options.mk
```diff
@@ -1410,6 +1418,13 @@ endif
 
 # wolfHSM server options
 ifeq ($(WOLFHSM_SERVER),1)
```
dgarske, line 1420: **⚪ [Info] WOLFHSM_SERVER misconfiguration $(error) also aborts clean targets**

Because `options.mk` is included unconditionally by the top-level Makefile, the new `$(error)` for `WOLFHSM_SERVER=1` without `CERT_CHAIN_VERIFY` fires during the make read phase for every target, including `make clean`/`make distclean`. A directory whose `.config` sets `WOLFHSM_SERVER=1` but omits `CERT_CHAIN_VERIFY` could not be cleaned without first editing/removing `.config`. In practice a valid server `.config` already sets `CERT_CHAIN_VERIFY`, so this only affects misconfigured trees; failing fast is also arguably intentional.

**Fix:** Decide whether failing on clean targets is acceptable; if not, exempt the clean goals from the error.

dgarske (commented):
## Skoll Code Review

**Scan type:** `review`**Overall recommendation:** `COMMENT`
**Findings:** 4 total — 4 posted, 0 skipped
**2 finding(s) posted as inline comments** (see file-level comments below)

### Posted findings
- [Low] **Removed make option WOLFHSM_CLIENT_LOCAL_KEYS left dangling; new WOLFHSM_CLIENT_ID not registered** — `tools/config.mk:95,137 (and config/examples/sim-wolfHSM-client-ecc.config:28, sim-wolfHSM-client-mldsa.config:49)`
- [Low] **Potential hybrid-signing build break in wolfHSM no-keystore mode** — `src/image.c:2455-2459`
- [Info] **$(error) continuation line embeds leading whitespace into the message** — `options.mk:1423-1426`
- [Info] **WOLFHSM_SERVER misconfiguration $(error) also aborts clean targets** — `options.mk:1420-1426`

dgarske (changes requested):
https://github.com/wolfSSL/wolfBoot/pull/798#pullrequestreview-4520133860

options.mk
bigbrett: this isnt correct, it renders fine with GNU make

options.mk
bigbrett: Well, any build-time error enforcement in options.mk will result in this. Theres a bunch of them in options.mk already so was just following the pattern. @dgarske I assume you actually don't care about this?

dgarske: Fine with me as-is.

bigbrett:
@dgarske 

> [Low] Potential hybrid-signing build break in wolfHSM no-keystore mode — src/image.c:2455-2459

wolfHSM was not intended to be useable with hybrid signatures, since that would require a secondary public key ID to be registered for all woflHSM HALs. It is definitely something we can do and support, but was not within the scope of this PR. 

> [Low] Removed make option WOLFHSM_CLIENT_LOCAL_KEYS left dangling; new WOLFHSM_CLIENT_ID not registered — tools/config.mk:95,137 (and config/examples/sim-wolfHSM-client-ecc.config:28, sim-wolfHSM-client-mldsa.config:49)

This is the one thing my local skolling didn't seem to find (though innocuous). Fixed in my latest commit. I will re-request review when CI clears.

Also pushed an unrelated commit as I forgot to rename the Makefile variable to match the macro name

dgarske (approved):

PR #799, opened 2026-06-16
Implement port for STM32WBA. Bump wolfHAL lib to latest

AlexLanzano:
# Add STM32WBA55CG Nucleo port (wolfHAL backend)

  Adds a wolfBoot port for the STM32WBA55CG Nucleo (Cortex-M33) on the wolfHAL backend, and bumps lib/wolfHAL to the latest.

## Changes

  - New stm32wba target in arch.mk (Cortex-M33 / ARMv8-M)
  - New board hal/boards/stm32wba55cg_nucleo/ (board.c/.h/.mk) using the upstream wolfHAL flash/gpio/uart drivers
  - Linker script hal/stm32wba.ld (1 MB flash @ 0x08000000, 128 KB RAM)
  - Example config config/examples/stm32wba.config (ECC256/SHA256)
  - Bump lib/wolfHAL
  - Add the new config to CI (test-configs.yml)

  Test plan

  - [x] Builds clean for config/examples/stm32wba.config
  - [x] Flash factory.bin @ 0x08000000; confirm it verifies + boots a signed app

### DIFF
--- a/config/examples/stm32wba.config
+++ b/config/examples/stm32wba.config
@@ -1,0 +1,19 @@
+# STM32WBA55CG Nucleo, wolfBoot on the wolfHAL backend.
+#
+# Flash: 1 MB at 0x08000000, 8 KB pages, 128-bit (16-byte) aligned writes.
+# Non-secure / flat mode (the wolfHAL stm32wba flash driver uses the NS
+# flash registers). Bootloader occupies the first 64 KB; two 128 KB
+# partitions plus an 8 KB swap follow.
+TARGET=stm32wba
+BOARD=stm32wba55cg_nucleo
+WOLFHAL=1
+ARCH=ARM
+SIGN=ECC256
+HASH=SHA256
+NO_MPU=1
+WOLFBOOT_SECTOR_SIZE=0x2000
+WOLFBOOT_PARTITION_SIZE=0x20000
+WOLFBOOT_PARTITION_BOOT_ADDRESS=0x08010000
+WOLFBOOT_PARTITION_UPDATE_ADDRESS=0x08030000
+WOLFBOOT_PARTITION_SWAP_ADDRESS=0x08050000
+NVM_FLASH_WRITEONCE=1
--- a/hal/boards/stm32wba55cg_nucleo/board.c
+++ b/hal/boards/stm32wba55cg_nucleo/board.c
@@ -1,0 +1,133 @@
+/* board.c
+ *
+ * STM32WBA55CG Nucleo board configuration using upstream wolfHAL drivers.
+ * Flash/Gpio/Uart singletons are instantiated by the driver .c files from
+ * the WHAL_CFG_* initializer macros in board.h; board.c uses the
+ * BOARD_*_DEV handles for the rest. RCC is header-inlined and hardcodes
+ * its base — no dev pointer needed.
+ *
+ * Clock target: HSE32 -> PLL1 (M=1, N=25, R=3) -> SYSCLK = 100 MHz, which
+ * requires PWR voltage scaling Range 1 and 3 flash wait states (RM0493).
+ *
+ * Copyright (C) 2026 wolfSSL Inc.
+ *
+ * This file is part of wolfBoot.
+ *
+ * wolfBoot is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * wolfBoot is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1335, USA
+ */
+
+#include <stddef.h>
+#include "hal.h"
+#include "board.h"
+
+/* Flash clock gate — enabled before raising the wait-state count. */
+static const whal_Stm32wba_Rcc_PeriphClk g_flashClock = {WHAL_STM32WBA55_FLASH_CLOCK};
+
+#ifdef DEBUG_UART
+/* GPIO ports A (LED PA9, USART1 RX PA8) and B (USART1 TX PB12) plus USART1. */
+static const whal_Stm32wba_Rcc_PeriphClk g_periphClks[] = {
+    {WHAL_STM32WBA55_GPIOA_CLOCK},
+    {WHAL_STM32WBA55_GPIOB_CLOCK},
+    {WHAL_STM32WBA55_USART1_CLOCK},
+};
+#define PERIPH_CLK_COUNT (sizeof(g_periphClks) / sizeof(g_periphClks[0]))
+#endif
+
+/*
+ * Switch PWR voltage scaling to Range 1 (required for >16 MHz operation).
+ * After reset the device is in Range 2 (max 16 MHz). Must switch to Range 1
+ * before configuring PLL or increasing SYSCLK.
+ *
+ * PWR base: 0x46020800, PWR_VOSR offset 0x00C
+ *   bit 16 VOS: 0=Range2, 1=Range1
+ *   bit 15 VOSRDY: read-only, 1 when stable
+ */
+#define PWR_BASE            0x46020800
+#define PWR_VOSR_REG        0x00C
+#define PWR_VOSR_VOS_Msk    (1UL << 16)
+#define PWR_VOSR_VOSRDY_Msk (1UL << 15)
+
+static void set_vos_range1(void)
+{
+    whal_Reg_Update(PWR_BASE, PWR_VOSR_REG, PWR_VOSR_VOS_Msk, PWR_VOSR_VOS_Msk);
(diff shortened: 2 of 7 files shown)


hal/boards/stm32wba55cg_nucleo/board.mk
```diff
@@ -0,0 +1,42 @@
...
+# underlying stm32wb_gpio.c must be given the WB-prefixed flag directly.
+CFLAGS+=-DWHAL_CFG_STM32WB_GPIO_DIRECT_API_MAPPING
+CFLAGS+=-DWHAL_CFG_STM32WBA_UART_DIRECT_API_MAPPING
+# UART is single-instance — reads its singleton from board.h instead of the
+# passed dev pointer.
+CFLAGS+=-DWHAL_CFG_STM32WBA_UART_SINGLE_INSTANCE
+
+WOLFHAL_OBJS+=$(WOLFHAL_ROOT)/src/reg.o
+WOLFHAL_OBJS+=$(WOLFHAL_ROOT)/src/flash/stm32wba_flash.o
+ifeq ($(DEBUG_UART),1)
```
dgarske, line 28: **🟠 [Medium] DEBUG_UART path (GPIO/UART config macros + drivers) is never built in CI** · _test_

The new CI job stm32wba_wolfhal_test builds config/examples/stm32wba.config without DEBUG_UART=1. In board.mk the GPIO/UART driver objects are only added when DEBUG_UART=1, and in board.c the g_periphClks[] table, whal_Gpio_Init, and whal_Uart_Init calls are all guarded by #ifdef DEBUG_UART. Consequently the GPIO/UART singleton initializer macros in board.h (WHAL_CFG_STM32WBA_GPIO_DEV, WHAL_CFG_STM32WB_UART_DEV) and the pin/AF table are never macro-expanded or compiled in CI. A typo in those macros (wrong pin index, AF number, base symbol, or a missing wolfHAL symbol) would not be caught - only the flash path is exercised. A DEBUG_UART build would also flush out any missing-symbol regressions in the bumped lib/wolfHAL submodule (not checked out in this tree).

**Fix:** Add a second CI variant that builds with DEBUG_UART=1 so the GPIO/UART configuration macros and the wolfHAL gpio/uart driver TUs are compiled at least once, protecting the board.h pin table from silent breakage.

hal/boards/stm32wba55cg_nucleo/board.c
```diff
@@ -0,0 +1,133 @@
...
+#define PWR_VOSR_VOSRDY_Msk (1UL << 15)
+
+static void set_vos_range1(void)
+{
+    whal_Reg_Update(PWR_BASE, PWR_VOSR_REG, PWR_VOSR_VOS_Msk, PWR_VOSR_VOS_Msk);
+    while (!(whal_Reg_Read(PWR_BASE, PWR_VOSR_REG) & PWR_VOSR_VOSRDY_Msk))
+        ;
+}
+
+/* HSE32 -> PLL1 (M=1, N=25, R=3 -> 100 MHz) -> SYSCLK */
```
dgarske: **🔵 [Low] Documented PLL1 configuration / divider comment does not arithmetically yield the stated 100 MHz SYSCLK** · _Logic_

Both modes flagged that the header comment 'HSE32 -> PLL1 (M=1, N=25, R=3) -> SYSCLK = 100 MHz' does not reconcile with a textbook STM32 PLL formula where M/R equal the literal divisor: HSE32/M=1 gives a 32 MHz VCO input (outside the declared '.rge = WHAL_STM32WBA_RCC_PLL1RGE_8_16' 8-16 MHz range), and 32*25/3 ~= 266 MHz, which would also over-clock the 100 MHz-max part. The numbers only reconcile to 100 MHz if the wolfHAL stm32wba_rcc driver encodes M and R as (divisor-1): m=1 -> /2 (16 MHz VCO input, matching PLL1RGE_8_16), r=3 -> /4, i.e. 32/2*25/4 = 100 MHz. This is most likely correct given the PR was hardware-validated, but the wolfHAL submodule is not checked out so the field encoding cannot be confirmed from the tree. Does not affect the build. Severity view differs: 'review' rated this Low (NIT/question), 'review-security' rated it Info; the stricter Low is kept here.

**Fix:** Confirm the achieved SYSCLK against RM0493/the datasheet (must be <= 100 MHz) and the M/N/R/RGE field semantics of whal_Stm32wba_Rcc_Pll1Cfg, then add a one-line comment spelling out the (divisor-1) field encoding (e.g. 'M+1=2 -> 16 MHz VCO in; xN=25 -> 400 MHz VCO; /(R+1=4) -> 100 MHz SYSCLK') so the comment matches the literal field values.

config/examples/stm32wba.config
```diff
@@ -0,0 +1,19 @@
...
+# Non-secure / flat mode (the wolfHAL stm32wba flash driver uses the NS
+# flash registers). Bootloader occupies the first 64 KB; two 128 KB
+# partitions plus an 8 KB swap follow.
+TARGET=stm32wba
+BOARD=stm32wba55cg_nucleo
+WOLFHAL=1
+ARCH=ARM
+SIGN=ECC256
+HASH=SHA256
+NO_MPU=1
```
dgarske, line 13: **⚪ [Info] New config disables the MPU (NO_MPU=1), reducing defense-in-depth** · _Security_

The example config sets NO_MPU=1, disabling wolfBoot's MPU-based region protection for the Cortex-M33. This removes a hardening layer (e.g. marking regions non-executable/read-only while the bootloader runs). This is a deliberate bring-up choice common to several wolfBoot ports and is not itself a vulnerability, but it is worth noting for a security-focused bootloader on an ARMv8-M part that supports an MPU.

**Fix:** Consider enabling the MPU for production builds once the port is stable, or document why it is left disabled for this target.

hal/boards/stm32wba55cg_nucleo/board.c
```diff
@@ -0,0 +1,133 @@
...
+    whal_Stm32wba_Flash_Ext_SetLatency(BOARD_FLASH_DEV, 1);
+}
+
+void hal_init(void)
+{
+    pll_clock_on();
+    whal_Flash_Init(BOARD_FLASH_DEV);
+
+#ifdef DEBUG_UART
+    for (size_t i = 0; i < PERIPH_CLK_COUNT; i++) {
```
dgarske: **⚪ [Info] C99 mid-block for-loop declarations deviate from wolfSSL C89 style** · _Logic_

The new code uses C99-style in-loop variable declarations ('for (size_t i = 0; ...)' and 'for (size_t i = PERIPH_CLK_COUNT; i-- > 0; )') rather than declaring the index at the top of the block per wolfSSL C89 conventions. This matches the existing sibling stm32wb board.c and is not covered by the Makefile's --std=c89 cppcheck (which only scans top-level hal/*.c, not hal/boards/**), so it will not break CI, but it diverges from project style for new code.

**Fix:** For strict C89 conformance, declare 'size_t i;' at the top of each function and use plain 'for (i = 0; ...)'.

dgarske (commented):
## Skoll Multi-Scan Review

**Findings:** 5 total — 5 posted, 0 skipped
**1 finding(s) not tied to a diff line** (full detail below)

### Posted findings
- [Medium] [review] **DEBUG_UART path (GPIO/UART config macros + drivers) is never built in CI** — `hal/boards/stm32wba55cg_nucleo/board.mk:28-31`
- [Low] [review+review-security] **Documented PLL1 configuration / divider comment does not arithmetically yield the stated 100 MHz SYSCLK** — `hal/boards/stm32wba55cg_nucleo/board.c:69-93`
- [Info] [review-security] **New config disables the MPU (NO_MPU=1), reducing defense-in-depth** — `config/examples/stm32wba.config:13`
- [Info] [review-security] **C99 mid-block for-loop declarations deviate from wolfSSL C89 style** — `hal/boards/stm32wba55cg_nucleo/board.c:111,126`

#### Non-ASCII (em-dash U+2014) characters introduced in new source files
**File:** `hal/boards/stm32wba55cg_nucleo/board.c:7,35; hal/boards/stm32wba55cg_nucleo/board.h:70,80,106; hal/boards/stm32wba55cg_nucleo/board.mk:13,22`
**Function:** `(file-level comments)`
**Severity:** Low
**Category:** convention

Both reviewers independently flagged that the new board files introduce U+2014 em-dash characters (-) in comments instead of 7-bit ASCII (7 newly-added lines across board.c, board.h, and board.mk, e.g. board.c:35 '/* Flash clock gate -- enabled before raising the wait-state count. */'). wolfBoot/wolfSSL source follows an ASCII-only convention. This compiles cleanly so it is non-blocking, but the non-ASCII bytes can confuse editors/diff tooling and violate project style. Note: the already-merged sibling stm32wb_nucleo files use the same em-dash style, so there is local precedent; this PR still adds new non-ASCII bytes.

**Recommendation:** Replace all em-dash (U+2014) occurrences with the ASCII hyphen '-' in the new board.c, board.h, and board.mk. Optionally clean up the sibling stm32wb_nucleo files in a separate pass.

---

config/examples/stm32wba.config
AlexLanzano: Looks like the current MPU support is for ARMv7. However, the cortext-M33 is ARMv8 so this needs to remain disabled on this port for now.

hal/boards/stm32wba55cg_nucleo/board.mk
AlexLanzano: Fixed

hal/boards/stm32wba55cg_nucleo/board.c
AlexLanzano: Fixed

hal/boards/stm32wba55cg_nucleo/board.c
AlexLanzano: fixed

dgarske (approved):

danielinux (approved):

PR #800, opened 2026-06-16
Add NXP QorIQ support using wolfIP to test-app

dgarske:


hal/nxp_t2080.c
```diff
@@ -582,8 +582,18 @@ static void hal_cpld_init(void)
...
+ * to 0xF0000000 (256MB) and keeps the ucode near the top; the T2080 RDB and
+ * NAII 68PPC2 have a 128MB NOR at 0xE8000000 with the ucode just below the
+ * wolfBoot region (matches the T1040 RDB). A wrong address here is read in
+ * hal_fman_init() and an out-of-NOR value machine-checks, so the default is
+ * board-gated and bounds-checked at runtime. */
 #ifndef FMAN_FW_ADDR
+#ifdef BOARD_CW_VPX3152
 #define FMAN_FW_ADDR    0xFFE60000UL
+#else
+#define FMAN_FW_ADDR    0xEFF00000UL
```
danielinux, line 591: `FMAN_FW_ADDR` here overlaps the boot partition on T2080 RDB / NAII 68PPC2

On T2080 RDB and NAII targets:

Swap: 0xEFDD0000 – 0xEFDE0000
Update partition: 0xEFDE0000 – 0xEFEE0000
Boot partition (WOLFBOOT_PARTITION_SIZE=0x100000): 0xEFEE0000 – 0xEFFE0000
wolfBoot (BOOTLOADER_PARTITION_SIZE=0x20000): 0xEFFE0000 – 0xF0000000



0xEFF00000 falls at offset 0x20000 inside the boot partition. The FMan microcode region is 64–128 KB (see the T1024/T1040 tables at docs/Targets.md:5660 and :5693), so flashing the microcode corrupts the signed application, and flashing or updating the application erases the microcode. 0xEFDD0000–0xF0000000 is fully allocated, so there is no free space at or near this address.

The runtime bounds check added in hal_fman_init() does not catch this: 0xEFF00000 - 0xE8000000 = 0x7F00000, which is below FLASH_BANK_SIZE (128 MB), so the address passes as in-NOR. The check correctly guards against a wrong-board address; it does not guard against a same-board partition collision.

The matches the T1040 RDB rationale in the comment does not carry over. The T1040 uses 0xEFF00000 safely because its BOOTLOADER_PARTITION_SIZE=0xC0000 places wolfBoot at 0xEFF40000, leaving a 256 KB gap at 0xEFF00000–0xEFF40000 that sits outside every partition. The T2080's 0x20000 bootloader leaves no equivalent gap.

The pre-PR documented address 0xE8020000 remains valid: it sits above the RCW at 0xE8000000 and below WOLFBOOT_DTS_BOOT_ADDRESS=0xE8040000, giving exactly 128 KB of unallocated space. BOARD_CW_VPX3152 is unaffected — its 8 MB boot partition at 0xFF000000 ends at 0xFF800000, well below 0xFFE60000.

Suggested change:

```suggestion
#ifndef FMAN_FW_ADDR
#ifdef BOARD_CW_VPX3152
/* 256 MB NOR at 0xF0000000; boot partition is 8 MB at 0xFF000000..0xFF800000,
 * wolfBoot at 0xFFFE0000 -- 0xFFE60000 is unallocated. */
#define FMAN_FW_ADDR    0xFFE60000UL
#else
/* 128 MB NOR at 0xE8000000. The partitions are contiguous from 0xEFDD0000 to
 * the top of flash, so the ucode cannot live near wolfBoot as it does on the
 * T1040 (whose 0xC0000 bootloader leaves a gap at 0xEFF00000). Place it in the
 * unallocated 128 KB between the RCW and WOLFBOOT_DTS_BOOT_ADDRESS
 * (0xE8040000). */
#define FMAN_FW_ADDR    0xE8020000UL
#endif
#endif
```

Consider also making the collision a build-time error rather than relying on the runtime NOR-window check, since the partition bounds are known at compile time:

```C
#if FMAN_FW_ADDR >= WOLFBOOT_PARTITION_BOOT_ADDRESS && \
    FMAN_FW_ADDR <  (WOLFBOOT_PARTITION_BOOT_ADDRESS + WOLFBOOT_PARTITION_SIZE)
#error "FMAN_FW_ADDR overlaps the wolfBoot boot partition"
#endif
```

docs/Targets.md
```diff
@@ -6196,8 +6211,8 @@ Flash Layout (T2080 RDB / NAII 68PPC2, 128 MB flash):
 | Description | File | Address |
 | ----------- | ---- | ------- |
 | Reset Configuration Word (RCW) | _(board-specific)_ | `0xE8000000` |
-| Frame Manager Microcode | `fsl_fman_ucode_t2080_r1.0.bin` | `0xE8020000` |
-| Signed Application | `test-app/image_v1_signed.bin` | `0xE8080000` |
+| Frame Manager Microcode | `fsl_fman_ucode_t2080_r1.0.bin` | `0xEFF00000` |
```
danielinux, line 6111: Same overlap as the comment on FMAN+FW+ADDR
The flash layout table for the 128 MB boards encodes the overlap described above: the microcode at 0xEFF00000 lies inside the signed application's partition at 0xEFEE0000 (1 MB). Both rows changed in this PR; the application row is correct and the microcode row should revert.

I'd suggest replacing with 0xE8020000 and dropping the T1040 comparison, since the reasoning is specific to the T1040's larger bootloader partition

danielinux (changes requested):

danielinux (approved):

PR #792, opened 2026-06-10
Fenrir fixes 2026 06 10

danielinux:
82142f87 F-4412: add equal-version emergency rollback test to pin >= operator
9e390e8f F-4414: replace strtol with strtoul for fw_version to fix saturation on 32-bit-long platforms
ad3bc885 F-4418: fix BSS-clear collision guard to use mem_size instead of file_size
aebecba5 F-4419: zeroize LmsKey on all error paths in keygen_lms
ef16945b F-4420: zeroize XmssKey on all error paths in keygen_xmss
63b58c54 F-4647: bound strlen to ATA_SECURITY_PASSWORD_LEN in passphrase path
a24f1079 F-4649: erase pkcs11_store object sectors on write-mode open to clear residual key material
2a3cbf2c F-4784: erase psa_store object data on write-mode open to clear residual key material
051c41cd F-4786: fix delta_base_version LE decode in base_diff for big-endian hosts
59581859 F-4788: zeroize stack ed25519_key after Free in keygen_ed25519 cleanup
3349bef3 F-4966: fix nwords=0 truncation on sub-word write to 16-bit NOR flash
793cec45 F-4969: zeroize stack RsaKey after FreeRsaKey in keygen_rsa cleanup
ca3e76a9 F-4970: zeroize stack ecc_key after Free in keygen_ecc cleanup
652febdc F-4971: zeroize stack ed448_key after Free in keygen_ed448 cleanup
f20cd302 F-4972: zeroize stack wc_MlDsaKey after Free in keygen_ml_dsa cleanup
87fc9296 F-5092: fix off-by-one in fwtpm_nv_read/write/erase offset guard
286581cf F-5093: fix OOB memset in x86_paging_setup_ptp and non-looping panic
712f3898 F-5348: always evaluate keyslot_CT_hint_matches for every slot
d3bd09f4 F-5351: test wolfBoot_open_self_address rejects bad magic with hdr_ok cleared
c165905f F-5357: zeroize digest and uuid_be stack buffers in hal_uds_derive_key before every return
2b0fde4b F-5745: fix off-by-one in hal_flash_erase skipping last page when len%PAGE_SIZE!=0
f7e77333 F-5747: guard fdt_data_size_ and fdt_splice_string_ against off_dt_strings+size_dt_strings uint32_t overflow
34231739 F-4258: reject GPT partitions whose LBA extent overflows uint64_t
71545e96 F-4339: widen uart-flash-server bounds checks to 64-bit to stop uint32_t wrap
[...]

PR #774, opened 2026-05-07
PolarFire SoC M-Mode LPDDR4 + minimal SBI runtime

dgarske:
## Summary

wolfBoot runs standalone in M-mode on the E51 monitor core, replacing both HSS and OpenSBI: it brings up LPDDR4 in software, secure-boots a signed Yocto FIT from SD, and stays resident as a minimal SBI runtime so the U54 cores boot Linux. Result: secure-boot 4-CPU SMP Yocto Linux to a login prompt in ~40 s from power-on, with no Microchip firmware in the boot chain.

## Commits

1. **PolarFire SoC: LPDDR4 DDR init for MPFS250T Video Kit (M-Mode)** - software LPDDR4 init/training (all-4-lane WRCALIB gate + retry), SD secure boot of a signed Yocto FIT (SDHCI PIO + PDMA-to-DDR staging with read-back verify), SHA384 + ECC384 verification, FIT parse, L2 dtb fixups, and S-mode hand-off. The licensed Cadence DDR controller logic (controller CSR programming, Memory Test Controller, LPDDR4 mode-register protocol) is a target-independent driver in `src/ddr_cadence.c` + `include/ddr_cadence.h`; the Microchip PHY/PLL/clock/training and the board `LIBERO_SETTING_*` values stay in `hal/mpfs250_ddr.c`.
2. **RISC-V: minimal SBI runtime; PolarFire SoC boots 4-CPU SMP Yocto Linux** - a minimal M-mode SBI runtime (`src/riscv_sbi.c`): BASE/TIME/IPI/RFENCE/HSM/DBCN plus the legacy calls, with per-hart M-mode trap stacks, `rdtime` and misaligned load/store emulation, atomic cross-hart IPI ops with self-directed SSIP delivery, and the trap-entry / M->S hand-off wiring. PolarFire SoC bring-up adds CLINT MTIME enable, per-hart HSM `hart_start` mailboxes, the watchdog policy, and the dtb fixups that bring up 4-CPU SMP Linux.

## Key design points

- **Generic Cadence driver**: the controller half is reusable across Cadence-DDR SoCs (controller base overridable via `DDR_CADENCE_CTRL_BASE`); the SoC-specific PHY/PLL/training is behind the platform HAL. Both compile only when `MPFS_DDR_INIT` is set.
[...]

PR #801, opened 2026-06-16
fix broken unit test

bigbrett:
Fixes broken unit tests on master causing all current PRs to fail. 

Seems that two separate PRs came together to break unit tests on master.

#746 changed the ELF loader's collision guard: when a segment would overwrite program headers that haven't been read yet, the loader after this PR fails with -5 instead of the old behavior of silently skipping the segment and returning 0.

The F-4418 regression test (added in #792 ) still expected the old skip-and-succeed behavior. Neither PR failed individually since they never tested each others changes, but now all PRs are failing.

Fix: update the test to expect the loader to reject the image (ret == -5) rather than return success.

### DIFF
--- a/tools/unit-tests/unit-elf-bss-guard.c
+++ b/tools/unit-tests/unit-elf-bss-guard.c
@@ -84,7 +84,7 @@ START_TEST(test_elf_bss_guard_no_header_corruption)
      *   → write proceeds; memset zeroes ph[1].
      * With the fixed guard (mem_size check):
      *   vaddr + mem_size   = ph[1]_start + 56 > ph[1]_start                  → FALSE
-     *   → write is skipped; ph[1] intact.
+     *   → load is aborted (-5) before any write; ph[1] intact.
      */
     ph[0].type      = ELF_PT_LOAD;
     ph[0].flags     = 0;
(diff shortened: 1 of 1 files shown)


tools/unit-tests/unit-elf-bss-guard.c
```diff
@@ -106,7 +106,10 @@ START_TEST(test_elf_bss_guard_no_header_corruption)
     }
 
     ret = elf_load_image_mmu(g_image, sizeof(g_image), &entry, NULL);
-    ck_assert_int_eq(ret, 0);
+
+    /* The loader should reject this image instead of loading it, because the
+     * segment would overwrite part of the image it has not read yet. */
+    ck_assert_int_eq(ret, -5);
```
padelsbach, line 112: `ck_assert_int_ne(ret, 0);` might be more generic and less fragile should the error code change again

padelsbach (approved):

tools/unit-tests/unit-elf-bss-guard.c
bigbrett: I kinda figured it should be fragile and break if the internals are changed? Otherwise perhaps it would still be overwriting the image and failing for another reason? IDK

dgarske (approved):
Thanks! I also fixed here -> https://github.com/wolfSSL/wolfBoot/pull/774/changes/4310c4b36f9aecc4a76fe11c4083658d00b903d5 
Was getting really annoying and I should have just opened a new PR!

PR #802, opened 2026-06-17
fix: bound unauthenticated image size before RAM load

rizlik:
ZD#21988

### DIFF
--- a/src/libwolfboot.c
+++ b/src/libwolfboot.c
@@ -2412,7 +2412,7 @@ int wolfBoot_ram_decrypt(uint8_t *src, uint8_t *dst)
     len = im2n(len);
 
 #if !defined(WOLFBOOT_FIXED_PARTITIONS) && !defined(WOLFBOOT_RAMBOOT_MAX_SIZE)
-#  error "WOLFBOOT_RAMBOOT_MAX_SIZE required when WOLFBOOT_NO_PARTITIONS=1"
+#  error "WOLFBOOT_FIXED_PARTITIONS or WOLFBOOT_RAMBOOT_MAX_SIZE required to bound the RAM load"
 #endif
     /* Bound the UNAUTHENTICATED image length before it drives the copy into the
      * RAM load region: the image is loaded to RAM before its signature is
--- a/src/libwolfboot.c
+++ b/src/libwolfboot.c
@@ -2406,7 +2406,10 @@ int wolfBoot_ram_decrypt(uint8_t *src, uint8_t *dst)
         wolfBoot_printf("Error decrypting header at %p!\n", src);
         return -1;
     }
-    len = *((uint32_t*)(dec_hdr + sizeof(uint32_t)));
+    /* dec_hdr is a byte buffer: copy the little-endian length field without an
+     * unaligned cast, then convert to native byte order. */
+    XMEMCPY(&len, dec_hdr + sizeof(uint32_t), sizeof(len));
+    len = im2n(len);
 
 #if !defined(WOLFBOOT_FIXED_PARTITIONS) && !defined(WOLFBOOT_RAMBOOT_MAX_SIZE)
 #  error "WOLFBOOT_RAMBOOT_MAX_SIZE required when WOLFBOOT_NO_PARTITIONS=1"
--- a/src/libwolfboot.c
+++ b/src/libwolfboot.c
@@ -2417,15 +2417,17 @@ int wolfBoot_ram_decrypt(uint8_t *src, uint8_t *dst)
     /* Bound the UNAUTHENTICATED image length before it drives the copy into the
      * RAM load region: the image is loaded to RAM before its signature is
      * verified, so this length (read from the not-yet-authenticated header) is
-     * attacker-influenceable and must be range checked first. */
-#ifdef WOLFBOOT_FIXED_PARTITIONS
-    if (WOLFBOOT_PARTITION_SIZE <= IMAGE_HEADER_SIZE ||
-            len > (uint32_t)(WOLFBOOT_PARTITION_SIZE - IMAGE_HEADER_SIZE)) {
+     * attacker-influenceable and must be range checked first. When both are
+     * configured, WOLFBOOT_RAMBOOT_MAX_SIZE takes precedence: it is the explicit
+     * cap on the RAM load region and may be tighter than the partition size. */
+#if defined(WOLFBOOT_RAMBOOT_MAX_SIZE)
+    if (len > WOLFBOOT_RAMBOOT_MAX_SIZE) {
         wolfBoot_printf("Invalid encrypted image size %u at %p\n", len, src);
         return -1;
     }
-#elif defined(WOLFBOOT_RAMBOOT_MAX_SIZE)
-    if (len > WOLFBOOT_RAMBOOT_MAX_SIZE) {
+#elif defined(WOLFBOOT_FIXED_PARTITIONS)
+    if (WOLFBOOT_PARTITION_SIZE <= IMAGE_HEADER_SIZE ||
+            len > (uint32_t)(WOLFBOOT_PARTITION_SIZE - IMAGE_HEADER_SIZE)) {
         wolfBoot_printf("Invalid encrypted image size %u at %p\n", len, src);
         return -1;
     }
--- a/src/libwolfboot.c
+++ b/src/libwolfboot.c
@@ -2408,6 +2408,26 @@ int wolfBoot_ram_decrypt(uint8_t *src, uint8_t *dst)
     }
     len = *((uint32_t*)(dec_hdr + sizeof(uint32_t)));
 
+#if !defined(WOLFBOOT_FIXED_PARTITIONS) && !defined(WOLFBOOT_RAMBOOT_MAX_SIZE)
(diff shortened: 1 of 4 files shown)


dgarske (approved):

tools/unit-tests/unit-update-ram-enc.c
```diff
@@ -0,0 +1,334 @@
...
+
+/* Build a plaintext header (magic + chosen length field), zero-padded to one
+ * IMAGE_HEADER_SIZE, then encrypt it into 'enc'. Enough to reach the length
+ * check in wolfBoot_ram_decrypt(); the body is irrelevant for the reject path. */
+static void make_encrypted_header(uint8_t *enc, uint32_t len)
+{
+    uint8_t plain[IMAGE_HEADER_SIZE];
+    uint32_t magic = WOLFBOOT_MAGIC;
+
+    memset(plain, 0, sizeof(plain));
```
dgarske, line 118: **🟠 [Medium] Bound-check edge cases not exercised: exact max-valid length and FIXED underflow guard** · _Test Coverage_

The new tests bracket the limit well (valid len=1024, oversize len=WOLFBOOT_PARTITION_SIZE, the 32-bit wrap value, and a bit-flip-forged length), but nothing pins the exact accept/reject boundary of the comparison `len > (WOLFBOOT_PARTITION_SIZE - IMAGE_HEADER_SIZE)`. An off-by-one in that expression (`>` vs `>=`, or +/- IMAGE_HEADER_SIZE) would not be caught. The new FIXED-partition underflow guard `WOLFBOOT_PARTITION_SIZE <= IMAGE_HEADER_SIZE` is also untested (it only triggers for a sub-256-byte partition, which is an extreme config and admittedly hard to fixture).

**Fix:** Add a boundary pair: one image with len == (WOLFBOOT_PARTITION_SIZE - IMAGE_HEADER_SIZE) (or RAMBOOT_MAX_SIZE for the nopart build) that must return 0, and one with len one block larger that must return -1, to lock the comparison's off-by-one behavior; the underflow-guard branch can be left untested or noted as defensive-only.

src/libwolfboot.c
```diff
@@ -2408,6 +2408,26 @@ int wolfBoot_ram_decrypt(uint8_t *src, uint8_t *dst)
...
     len = *((uint32_t*)(dec_hdr + sizeof(uint32_t)));
 
+#if !defined(WOLFBOOT_FIXED_PARTITIONS) && !defined(WOLFBOOT_RAMBOOT_MAX_SIZE)
+#  error "WOLFBOOT_RAMBOOT_MAX_SIZE required when WOLFBOOT_NO_PARTITIONS=1"
+#endif
+    /* Bound the UNAUTHENTICATED image length before it drives the copy into the
+     * RAM load region: the image is loaded to RAM before its signature is
+     * verified, so this length (read from the not-yet-authenticated header) is
+     * attacker-influenceable and must be range checked first. */
+#ifdef WOLFBOOT_FIXED_PARTITIONS
```
dgarske: **🔵 [Low] WOLFBOOT_RAMBOOT_MAX_SIZE silently ignored when WOLFBOOT_FIXED_PARTITIONS is also defined** · _Logic_

Both new bounds checks use `#ifdef WOLFBOOT_FIXED_PARTITIONS ... #elif defined(WOLFBOOT_RAMBOOT_MAX_SIZE)`. If a configuration defines BOTH macros, the partition-size branch wins and WOLFBOOT_RAMBOOT_MAX_SIZE is silently ignored, inverting the priority of the BEFORE code in update_ram.c (which preferred the NO_PARTITIONS / RAMBOOT branch first). If someone intentionally set WOLFBOOT_RAMBOOT_MAX_SIZE smaller than the partition to constrain a RAM load region under fixed partitions, the looser partition-size bound would be used instead, potentially overrunning the smaller RAM region. NOT a regression today: every shipping config that sets WOLFBOOT_RAMBOOT_MAX_SIZE (polarfire, raspi3, raspi3-encrypted, versal) also sets WOLFBOOT_NO_PARTITIONS=1, so WOLFBOOT_FIXED_PARTITIONS is undefined and the two branches are mutually exclusive. The same pattern is duplicated in src/update_ram.c:101-112. Severity views: review rated this Low (SUGGEST/question), review-security rated it Info; stricter severity (Low) kept. Reported as a latent maintenance hazard.

**Fix:** Confirm the intended precedence. If supporting both macros simultaneously is ever intended, apply the tighter (minimum) of the two bounds rather than letting the partition branch shadow RAMBOOT_MAX_SIZE. Otherwise document that RAMBOOT_MAX_SIZE is a NO_PARTITIONS-only knob (a comment noting the macros are mutually exclusive). Note: the `#error` text ("...required when WOLFBOOT_NO_PARTITIONS=1") is also slightly stale relative to the new WOLFBOOT_FIXED_PARTITIONS condition.

dgarske: **🔵 [Low] New length bound does not account for ENCRYPT_BLOCK_SIZE rounding of the decrypt copy loop** · _Security_

The new bound limits the unauthenticated `len` to WOLFBOOT_PARTITION_SIZE - IMAGE_HEADER_SIZE (or WOLFBOOT_RAMBOOT_MAX_SIZE), so the intended copy region is at most WOLFBOOT_PARTITION_SIZE bytes. However, the unchanged decrypt loop immediately below writes in fixed ENCRYPT_BLOCK_SIZE granules (64 bytes for ChaCha, 16 for AES): `while (dst_offset < (len + IMAGE_HEADER_SIZE)) { ... XMEMCPY(dst + dst_offset, dec_block, ENCRYPT_BLOCK_SIZE); dst_offset += ENCRYPT_BLOCK_SIZE; }`. When len + IMAGE_HEADER_SIZE is not a multiple of ENCRYPT_BLOCK_SIZE, the final iteration writes up to ENCRYPT_BLOCK_SIZE - 1 bytes PAST len + IMAGE_HEADER_SIZE, i.e. up to roundup(len + IMAGE_HEADER_SIZE, ENCRYPT_BLOCK_SIZE). With len at its maximum, this is roundup(WOLFBOOT_PARTITION_SIZE, ENCRYPT_BLOCK_SIZE). This is safe ONLY because WOLFBOOT_PARTITION_SIZE / WOLFBOOT_RAMBOOT_MAX_SIZE (and IMAGE_HEADER_SIZE) are block-aligned in every shipping config, so the rounded-up value equals the bound and the destination RAM staging region is exactly WOLFBOOT_PARTITION_SIZE. If a target ever defines a non-block-aligned partition or ramboot-max size, the last block decrypt would write past the bound the check is meant to enforce. Defense-in-depth observation about the tightness of the newly introduced bound, not an exploitable overflow in current configurations.

**Fix:** Optionally tighten the bound to account for block rounding, e.g. compare against WOLFBOOT_PARTITION_SIZE after rounding len + IMAGE_HEADER_SIZE up to ENCRYPT_BLOCK_SIZE, or add a compile-time assertion that WOLFBOOT_PARTITION_SIZE % ENCRYPT_BLOCK_SIZE == 0 and WOLFBOOT_RAMBOOT_MAX_SIZE % ENCRYPT_BLOCK_SIZE == 0. No action required for the currently supported (block-aligned) targets.

tools/unit-tests/unit-update-ram-enc.c
```diff
@@ -0,0 +1,334 @@
...
+    tcase_set_timeout(overflow, 5);
+
+    return s;
+}
+
+int main(int argc, char *argv[])
+{
+    int fails;
+    Suite *s;
+    SRunner *sr;
```
dgarske, line 400: **🔵 [Low] Overflow-length tcase is missing its tcase_set_timeout call** · _Test Consistency_

Three of the four new tcases get an explicit 5s timeout (bitflip, valid, oversize), but overflow (test_ram_decrypt_overflow_len_rejected) is created and added to the suite without a matching `tcase_set_timeout(overflow, 5)`. It falls back to the libcheck default timeout. Purely cosmetic/consistency - all four reject-path tests are fast - but the omission looks accidental given the other three are set right next to it.

**Fix:** Add `tcase_set_timeout(overflow, 5);` so all four tcases set their timeout consistently.

dgarske (commented):
## Skoll Multi-Scan Review

**Findings:** 4 total — 4 posted, 0 skipped

### Posted findings
- [Medium] [review] **Bound-check edge cases not exercised: exact max-valid length and FIXED underflow guard** — `tools/unit-tests/unit-update-ram-enc.c:118-300`
- [Low] [review+review-security] **WOLFBOOT_RAMBOOT_MAX_SIZE silently ignored when WOLFBOOT_FIXED_PARTITIONS is also defined** — `src/libwolfboot.c:2418-2429`
- [Low] [review-security] **New length bound does not account for ENCRYPT_BLOCK_SIZE rounding of the decrypt copy loop** — `src/libwolfboot.c:2418-2439`
- [Low] [review] **Overflow-length tcase is missing its tcase_set_timeout call** — `tools/unit-tests/unit-update-ram-enc.c:323-326`

dgarske (changes requested):
https://github.com/wolfSSL/wolfBoot/pull/802#pullrequestreview-4518270572

dgarske (approved):

PR #815, opened 2026-07-02
Fix STM32U5 fallback erase and rollback handling

danielinux:
Fixes two bugs in the STM32U5 dual-bank fallback path.

  When wolfBoot invalidates a failing internal-flash partition, wolfBoot_erase_partition() now unlocks and re-locks internal flash itself. That makes the helper safe for existing callers, including the hwswap
  fallback path and wolfBoot_dualboot_candidate(), and fixes the STM32U5 case where erase operations could silently do nothing while flash was still locked.

  The hwswap boot loop in update_flash_hwswap.c now recomputes boot_v, update_v, and max_v on each retry. That prevents a corrupted higher-version image from leaving behind a stale max_v after erase and
  triggering a false anti-rollback panic when falling back to the remaining valid image.

zd22096

### DIFF
--- a/src/libwolfboot.c
+++ b/src/libwolfboot.c
@@ -773,20 +773,20 @@ int wolfBoot_get_update_sector_flag(uint16_t sector, uint8_t *flag)
  */
 void RAMFUNCTION wolfBoot_erase_partition(uint8_t part)
 {
-    uint32_t address = 0;
+    uintptr_t address = 0;
     int size = 0;
 
     switch (part) {
         case PART_BOOT:
-            address = (uint32_t)WOLFBOOT_PARTITION_BOOT_ADDRESS;
+            address = (uintptr_t)WOLFBOOT_PARTITION_BOOT_ADDRESS;
             size = WOLFBOOT_PARTITION_SIZE;
             break;
         case PART_UPDATE:
-            address = (uint32_t)WOLFBOOT_PARTITION_UPDATE_ADDRESS;
+            address = (uintptr_t)WOLFBOOT_PARTITION_UPDATE_ADDRESS;
             size = WOLFBOOT_PARTITION_SIZE;
             break;
         case PART_SWAP:
-            address = (uint32_t)WOLFBOOT_PARTITION_SWAP_ADDRESS;
+            address = (uintptr_t)WOLFBOOT_PARTITION_SWAP_ADDRESS;
             size = WOLFBOOT_SECTOR_SIZE;
             break;
         default:
--- a/src/libwolfboot.c
+++ b/src/libwolfboot.c
@@ -767,7 +767,10 @@ int wolfBoot_get_update_sector_flag(uint16_t sector, uint8_t *flag)
 /**
  * @brief Erase a partition.
  *
- * This function erases a partition.
+ * This function erases a partition. It manages the flash lock internally:
+ * the target flash (internal or external) is unlocked before the erase and
+ * locked again before returning. Callers do not need to unlock the flash
+ * beforehand, and on return the flash is left locked.
  *
  * @param[in] part Partition number.
  */
--- a/src/libwolfboot.c
+++ b/src/libwolfboot.c
@@ -799,7 +799,9 @@ void RAMFUNCTION wolfBoot_erase_partition(uint8_t part)
             ext_flash_erase(address, size);
             ext_flash_lock();
         } else {
+            hal_flash_unlock();
             hal_flash_erase(address, size);
+            hal_flash_lock();
         }
     }
 }
--- a/src/update_flash_hwswap.c
+++ b/src/update_flash_hwswap.c
@@ -46,11 +46,6 @@ void RAMFUNCTION wolfBoot_start(void)
     int active;
     struct wolfBoot_image fw_image;
     uint8_t p_state;
-#ifndef ALLOW_DOWNGRADE
-    uint32_t boot_v = wolfBoot_current_firmware_version();
-    uint32_t update_v = wolfBoot_update_firmware_version();
-    uint32_t max_v = (boot_v > update_v) ? boot_v : update_v;
-#endif
     active = wolfBoot_dualboot_candidate();
 
     if (active < 0) /* panic if no images available */
@@ -58,6 +53,9 @@ void RAMFUNCTION wolfBoot_start(void)
 
     for (;;) {
 #ifndef ALLOW_DOWNGRADE
+        uint32_t boot_v = wolfBoot_current_firmware_version();
+        uint32_t update_v = wolfBoot_update_firmware_version();
+        uint32_t max_v = (boot_v > update_v) ? boot_v : update_v;
         uint32_t active_v = (active == PART_UPDATE) ? update_v : boot_v;
         if ((max_v > 0U) && (active_v < max_v)) {
             wolfBoot_printf("Rollback to lower version not allowed\n");
--- a/src/update_flash_hwswap.c
+++ b/src/update_flash_hwswap.c
@@ -57,6 +57,13 @@ void RAMFUNCTION wolfBoot_start(void)
         uint32_t update_v = wolfBoot_update_firmware_version();
         uint32_t max_v = (boot_v > update_v) ? boot_v : update_v;
         uint32_t active_v = (active == PART_UPDATE) ? update_v : boot_v;
+        /* Anti-rollback defense-in-depth: wolfBoot_dualboot_candidate() always
+         * selects the max-version partition, and after a failed image is erased
+         * its version reads 0, so the remaining valid partition becomes the new
+         * max. This guard therefore should not normally trigger, but it is kept
+         * as a safety net in case a lower version is ever selected while a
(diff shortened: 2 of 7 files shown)


dgarske (commented):
## Skoll Multi-Scan Review

**Findings:** 5 total — 5 posted, 0 skipped
**1 finding(s) not tied to a diff line** (full detail below)

### Posted findings
- [Low] [review-security] **wolfBoot_erase_partition() now leaves internal flash LOCKED on return; peripheral in-tree caller writes to locked flash** — `src/libwolfboot.c:801-805`
- [Low] [review-security] **hwswap anti-rollback boot_panic() denial branch is now unreachable and no longer covered by any test** — `src/update_flash_hwswap.c:56-64`
- [Low] [review] **Script name sim-dualbank-rollback-denied.sh no longer matches its behavior** — `tools/scripts/sim-dualbank-rollback-denied.sh:1-66`
- [Low] [review] **Convoluted if(locked)/if(!locked) guards around partition_magic_write** — `tools/unit-tests/unit-nvm.c:318-323`

#### Sibling update_ram.c retains the stale max_v pattern this PR fixes in hwswap
**File:** `src/update_ram.c:253-290`
**Function:** `wolfBoot_start (update_ram)`
**Severity:** Low

This finding is NOT introduced by the PR and is outside the changed file set; it is raised only for consistency because it directly parallels the hwswap fix. `update_flash_hwswap.c` now recomputes `boot_v`/`update_v`/`max_v` inside the boot loop so that anti-rollback re-evaluates after a failing image is invalidated. `src/update_ram.c` still computes `boot_v`/`update_v`/`max_v` once before the loop (lines 253-257) and reuses them in the anti-rollback guard (lines 282-290). The mechanics differ (the update_ram fallback at `backup_on_failure` flips `active` without erasing, so versions persist), so the exact stale-erase trigger does not occur; however the same design intent - allowing fallback to a lower valid image when a higher image is corrupt - is not reflected here, and a higher-but-corrupt image would still drive `max_v` high enough to panic when falling back to the lower valid partition. Worth confirming whether the analogous behavior is intended for the RAM updater.

**Recommendation:** Confirm intended behavior for the RAM updater; apply the same recompute if consistency is desired. Out of scope for this PR - track separately.

_Referenced code: `src/update_ram.c:253-262` (10 lines)_

---

src/libwolfboot.c
```diff
@@ -799,7 +799,9 @@ void RAMFUNCTION wolfBoot_erase_partition(uint8_t part)
             ext_flash_erase(address, size);
             ext_flash_lock();
         } else {
```
dgarske, line 804: **🔵 [Low] wolfBoot_erase_partition() now leaves internal flash LOCKED on return; peripheral in-tree caller writes to locked flash** · _Blast Radius_

This PR changes the contract of the public API `wolfBoot_erase_partition()`. BEFORE, the internal-flash path only called `hal_flash_erase(address, size)` and never touched the lock state, so callers were responsible for unlock/lock and flash remained in whatever state the caller established. AFTER, the else-branch does `hal_flash_unlock(); hal_flash_erase(...); hal_flash_lock();`, meaning the function now always returns with internal flash LOCKED. The three core production callers (update_flash.c:236, update_flash_hwswap.c:79, libwolfboot.c:1773) were all made consistent with this and are correct. However, the in-tree caller `test-app/hifive1_write_page.c:write_page()` does `wolfBoot_erase_partition(0x01)` immediately followed by `hal_flash_write(dst, flash_page, PAGESIZE)` with no intervening `hal_flash_unlock()`. With the new self-locking behavior, that `hal_flash_write` now targets locked flash and can silently fail/no-op if the HiFive1 HAL enforces the lock. The unit tests in this same PR confirm the semantic change: they had to insert `hal_flash_unlock()` AFTER every `wolfBoot_erase_partition()` call (e.g. unit-nvm.c:308, unit-enc-nvm.c:87) precisely because erase now leaves flash locked.

**Fix:** Document the new lock-state postcondition in the `wolfBoot_erase_partition` doxygen comment (flash is locked on return). Audit peripheral/out-of-tree callers such as test-app/hifive1_write_page.c and add an explicit `hal_flash_unlock()` before their post-erase `hal_flash_write()`, or have them not rely on the erase leaving flash unlocked.

src/update_flash_hwswap.c
```diff
@@ -46,18 +46,16 @@ void RAMFUNCTION wolfBoot_start(void)
...
-    uint32_t max_v = (boot_v > update_v) ? boot_v : update_v;
-#endif
     active = wolfBoot_dualboot_candidate();
 
     if (active < 0) /* panic if no images available */
         boot_panic();
 
     for (;;) {
 #ifndef ALLOW_DOWNGRADE
+        uint32_t boot_v = wolfBoot_current_firmware_version();
```
dgarske, line 56: **🔵 [Low] hwswap anti-rollback boot_panic() denial branch is now unreachable and no longer covered by any test** · _Missing Tests_

By recomputing `boot_v`/`update_v`/`max_v` at the top of every loop iteration, `active_v` now always equals `max_v` whenever the denial check is evaluated: on entry `active = wolfBoot_dualboot_candidate()` always selects the max-version partition (or, in the TESTING-rollback case, returns the higher candidate without switching), and after a fallback the failed partition is erased so its version reads 0, making the remaining active partition the new max. Consequently the guard `(max_v > 0U) && (active_v < max_v)` can no longer be true, so `boot_panic()` at line 62 has become effectively dead code. BEFORE this PR the branch WAS reachable (stale higher `max_v` on the 2nd iteration after an erase) and was exercised by `test_hwswap_highversion_rollback_denied` (asserting `do_boot_called == 0`). This PR repurposes that test into `test_hwswap_highversion_fallback_boots_lower_version` (asserting `do_boot_called == 1`), so no remaining unit test exercises the anti-rollback denial/boot_panic path in the hwswap updater. Real anti-rollback protection still holds (the loop always targets the max version and only drops to a lower version after the higher one is erased for failing verification), so this is a defense-in-depth / maintainability + coverage concern rather than an exploitable hole.

**Fix:** Either (a) remove the now-unreachable denial branch if the fallback-always-boots-remaining-valid-image semantics is the intended contract, adding a comment that anti-rollback is enforced by dualboot_candidate + loop-targets-max; or (b) if the branch is meant as a real safety net, add a unit test that drives active_v < max_v (e.g. by injecting a state where the lower-version partition is selected while the higher valid one remains) so the security control stays regression-tested.

tools/scripts/sim-dualbank-rollback-denied.sh
```diff
@@ -45,21 +45,22 @@ rollback_output="$(timeout 3s ./wolfboot.elf get_version 2>&1)"
...
-if ! printf '%s\n' "$rollback_output" | grep -q "Rollback to lower version not allowed"; then
-    echo "Rollback denial message not found in output." >&2
+if ! printf '%s\n' "$rollback_output" | grep -q "Simulator do_boot app_offset"; then
+    echo "Fallback boot did not reach the application handoff." >&2
     echo "$rollback_output" >&2
     exit 1
 fi
 
-echo "Dualbank rollback-to-older-version denial verified."
+echo "Dualbank corrupted-update fallback verified."
```
dgarske, line 73: **🔵 [Low] Script name sim-dualbank-rollback-denied.sh no longer matches its behavior**

The script was inverted from verifying a rollback denial (expecting timeout/exit 80 and the 'Rollback to lower version not allowed' message) to verifying a successful corrupted-update fallback (expecting exit 0, absence of the denial message, and 'Simulator do_boot app_offset'). The new logic is correct and matches the fix, and the final echo/comments were updated. However the filename still says 'rollback-denied', which is now misleading. The file is referenced by two CI workflows (test-sunnyday-simulator.yml:59, test-hooks-simulator.yml:94), so a rename has cost, but the stale name will confuse future readers.

**Fix:** Rename the script (and its two CI references) or add a clarifying header comment to avoid the name/behavior mismatch.

tools/unit-tests/unit-nvm.c
```diff
@@ -313,8 +315,12 @@ START_TEST(test_partition_magic_write_stops_on_flash_write_error)
     erased_nvm_bank1 = 0;
     hal_flash_write_fail = 1;
 
+    if (locked)
```
dgarske: **🔵 [Low] Convoluted if(locked)/if(!locked) guards around partition_magic_write**

At this point in the test `locked` is deterministically 0 (line 307 `wolfBoot_erase_partition` self-locks to 1, line 308 `hal_flash_unlock()` sets it to 0), and `partition_magic_write` does not change the lock state. So the guards reduce to: skip the unlock, then take the `!locked` branch to lock - equivalent to the previous single `hal_flash_lock()`. The conditional form plus the trailing `locked = 1;` (line 335, already 1) obscures intent and papers over lock state rather than asserting it, unlike the clear `hal_flash_lock()` used before. Functionally correct and lock-balanced, just less readable than the surrounding tests.

**Fix:** Simplify to an unconditional lock after the call for readability and consistency with the other tests in the file.

dgarske (approved):

PR #818, opened 2026-07-07
Add AArch64 UEFI support (NVIDIA Jetson Orin)

dgarske:
## Summary

Adds a new `aarch64_efi` target that builds wolfBoot as an AArch64 UEFI application (`wolfboot.efi`) -- the direct sibling of the existing Qemu x86-64 UEFI target. It uses only UEFI Boot Services (no SoC-specific registers), so the same binary runs on any AArch64 UEFI platform. The platform firmware launches `wolfboot.efi`, which reads the next-stage image from the EFI Simple File System, authenticates it with wolfCrypt, and boots it via `LoadImage`/`StartImage` (an AArch64 Linux `Image` is itself a PE/COFF EFI-stub application).

Validated end to end on the NVIDIA Jetson Orin Nano Developer Kit (Tegra234): the on-module UEFI firmware (edk2-nvidia) launches wolfBoot after the NVIDIA-signed early chain (BootROM -> MB1 -> MB2 -> UEFI), and wolfBoot verifies and boots NVIDIA Jetson Linux (L4T R36.4.4, `5.15.148-tegra`) to an Ubuntu 22.04 login prompt.

## What it adds

- **UEFI-application HAL** (`hal/aarch64_efi.c`, `hal/aarch64_efi.ld`) and `config/examples/aarch64_efi.config`. No bare-metal init -- UEFI owns DRAM/MMU/GIC; the flash ops are stubs and storage/console come from Boot Services.
- **Measured boot via the firmware TPM** (`MEASURED_BOOT_TCG2=1`): using `EFI_TCG2_PROTOCOL`, wolfBoot extends the verified kernel, its command line, and the platform device tree into a PCR (`MEASURED_PCR_A`, default 9) with `HashLogExtendEvent`, then reads the PCR back. The firmware/fTPM does the hashing and log append, so wolfBoot needs no TPM transport driver and pulls in no wolfTPM.
- **Authenticated kernel command line** (new; both EFI targets): the Linux command line is carried in a signature-covered `HDR_CMDLINE` TLV in the manifest header instead of an unsigned file. wolfBoot reads it from the verified image and passes it to the kernel EFI stub via `LoadOptions`, so kernel arguments (`root=`, `init=`, security flags) cannot be altered without breaking the image signature. Sign with `--cmdline "..."` (a shorthand for the `HDR_CMDLINE` string TLV). The same path is added to `x86_64_efi` (which previously had no command-line mechanism).
- **UEFI Secure Boot enrollment docs**: how to enroll wolfBoot's signing identity into the platform db/KEK/PK so the fused firmware enforces `wolfboot.efi`, which in turn enforces the kernel -- a complete hardware root of trust.

## Notable details

- `open_kernel_image()` now returns `-1` on error rather than a truncated `EFI_STATUS` (a `UINTN` whose error high-bit would become a positive `int`) -- addresses review feedback.
- The command-line TLV read is a single shared helper (`include/wolfboot_efi.h`, `wolfBoot_efi_get_cmdline()`) included by both EFI HALs after their gnu-efi headers -- one auditable copy, no build wiring.
- Command line is ASCII, up to 255 bytes; it must fit within `IMAGE_HEADER_SIZE` (the default 256-byte ED25519/SHA256 header holds roughly 70 bytes; set `IMAGE_HEADER_SIZE=512` for longer lines). If the image carries no `HDR_CMDLINE` TLV, wolfBoot boots with no command line and the kernel uses its built-in `CONFIG_CMDLINE`.

## Testing

- Both `aarch64_efi` and `x86_64_efi` build (`make` with each config; `wolfboot.efi` produced).
- `sign --ed25519 --sha256 --cmdline "root=..." kernel.img key.der 1` produces a `HDR_CMDLINE` (0x0034) TLV that reads back exactly via the wolfBoot header parser.
- On hardware (Jetson Orin Nano): `wolfboot.efi` + a signed `kernel.img` on a FAT ESP boots L4T to an Ubuntu login; the console shows `Firmware Valid` and the TCG2 measurement lines.

## Docs

`docs/Targets.md` gains the "NVIDIA Jetson Orin (aarch64_efi)" section (deploy, boot Linux, Secure Boot enrollment, measured boot); `docs/Signing.md` documents `--cmdline` / `HDR_CMDLINE`.


docs/Targets.md
```diff
@@ -7850,6 +7851,189 @@ You can `Ctrl-C` or login as `root` and power off qemu with `poweroff`
...
+`cmdline.txt` (read by wolfBoot and passed to the kernel via `LoadOptions`):
+
+```
+root=/dev/mmcblk0p2 rw rootwait console=ttyTCU0,115200
+```
+
+On power-up the Jetson UEFI auto-boots `\EFI\BOOT\BOOTAA64.EFI`; wolfBoot verifies `kernel.img` and hands off to the kernel, which receives the real Tegra234 device tree from the UEFI configuration table (`EFI stub: Using DTB from configuration table`), mounts `mmcblk0p2`, and brings up systemd and the login on the J14 debug console (`ttyTCU0`).
+
+Security note: the plaintext `\cmdline.txt` is not covered by the image signature. For a production trust chain the command line should be authenticated -- baked into the signed image, or supplied via the device-tree `/chosen` bootargs. An initramfs-based flow (rather than a direct `root=` mount) would additionally need initrd support via the `LINUX_EFI_INITRD_MEDIA_GUID` LoadFile2 protocol, which this target does not currently implement.
+
```
danielinux: I understand the trade-off here but I think we could rather suggest to compile linux with CONFIG_CMDLINE="..." and the expected cmd line, in combination with CONFIG_CMDLINE_FORCE. With CONFIG_CMDLINE_FORCE the kernel ignores t bootargs set by UEFI and it might be a better fit for a proper secure boot deploy.

danielinux: an alternative could be to force the cmdline inside a custom TLV in the manifest header perhaps?

hal/aarch64_efi.c
```diff
@@ -0,0 +1,683 @@
...
+                               0);
+
+    if (status != EFI_SUCCESS)
+        file_handle = NULL;
+
+    return file_handle;
+}
+
+static int open_kernel_image(EFI_FILE_HANDLE vol, CHAR16 *filename,
+        EFI_PHYSICAL_ADDRESS *_addr, uint32_t *sz)
```
danielinux, line 558: As previously noted in the comment: EFI_STATUS is UINTN and all EFI errors set the high bit but will be truncated in the return statement as a positive (valid) value. The caller checks against the *_addr parameter, but the return value of open_kernel_image() is not indicating errors properly.
Instead of returning `status` we could rather return -1 for real error codes to avoid this, at lines 553 and 569

danielinux (changes requested):

docs/Targets.md
dgarske: I implemented and tested it! https://github.com/wolfSSL/wolfBoot/pull/818/changes/7da4c707ece27fdbf900e1e6ef2f96e4ab9f9f9b

hal/aarch64_efi.c
```diff
@@ -0,0 +1,635 @@
...
+    mem_path_device->EndingAddress =
+        (EFI_PHYSICAL_ADDRESS)((uintptr_t)boot_addr + *size - 1);
+    SetDevicePathNodeLength(&mem_path_device->Header,
+                            sizeof(MEMMAP_DEVICE_PATH));
+
+    SetDevicePathEndNode(&mem_path_device[1].Header);
+
+#ifdef WOLFBOOT_MEASURED_BOOT_EFI_TCG2
+    /* Measure kernel + cmdline + DTB into the firmware TPM before handoff. */
+    tcg2_measure_all(boot_addr, *size);
```
danielinux, line 450: This seems to be called too soon, and the cmdline is not part of the PCR. I think wolfBoot_efi_get_cmdline() should happen before this measure.

hal/aarch64_efi.c
```diff
@@ -0,0 +1,635 @@
...
+#endif
+
+    wolfBoot_printf("Staging kernel at address %p, size: %u\n", (void*)boot_addr, *size);
```
danielinux: "%p" in gnuefi prints 32-bit numbers, so the value here is truncated. Consider "%lx" instead

hal/aarch64_efi.c
```diff
@@ -0,0 +1,635 @@
...
+
+    /* Always leave *_addr well-defined so the caller's "no image" check is
+     * reliable: 0 on any failure, the loaded address only on full success. */
+    *_addr = 0;
+    *sz = 0;
+    file = openFile(filename, vol);
+    if (file == NULL)
+        return -1;
+
+    *sz =  FileSize(file);
```
danielinux: This is the "blocking" issue flagged by copilot, truncating the return value straight into a 32bit. I'd suggest to declare a UINT64 locally and then explictly truncate after detecting error (possibly as UINT64_MAX?)

hal/aarch64_efi.c
```diff
@@ -0,0 +1,635 @@
...
+    pages = (*sz / PAGE_SIZE) + 1;
```
danielinux: This allocates one page extra when sz is a multiple of PAGE_SIZE. Could be improved using `(*sz + PAGE_SIZE - 1) / PAGE_SIZE`

hal/aarch64_efi.c
```diff
@@ -0,0 +1,635 @@
...
+{
+    UINTN i;
+    uint32_t totalsize;
+    const uint8_t *p;
+
+    /* The platform publishes its device tree in a UEFI configuration table
+     * (the kernel boots against it -- "Using DTB from configuration table").
+     * Rather than trust a specific vendor GUID (which varies), scan the tables
+     * for the entry whose contents start with the FDT magic (0xd00dfeed). */
+    for (i = 0; i < gSystemTable->NumberOfTableEntries; i++) {
```
danielinux, line 363: Are you sure vendor GUID varies? EFI_DTB_GUID should be stable (looks like it's the same across other bootloaders which all agree on the constant form)
EFI_DTB_GUID = b1b621d5-f19c-41a5-830b-d9152c69aae0

hal/aarch64_efi.c
```diff
@@ -0,0 +1,635 @@
...
+static void tcg2_extend(WOLFBOOT_TCG2_PROTOCOL *tcg2, const void *addr,
+                        uint32_t size, const char *desc)
+{
+    uint8_t buf[__builtin_offsetof(WOLFBOOT_TCG2_EVENT, Event)
+                + WOLFBOOT_TCG2_DESC_MAX];
+    WOLFBOOT_TCG2_EVENT *evt = (WOLFBOOT_TCG2_EVENT *)buf;
+    EFI_STATUS status;
+    UINTN desclen = 0;
+
+    while (desc[desclen] != '\0' && desclen < WOLFBOOT_TCG2_DESC_MAX - 1)
```
danielinux, line 245: when desc is exactly WOLFBOOT_TCG2_DESC_MAX - 1 or longer, the loop stops at 31, desclen++ becomes 32 and copymem copies 32 bytes including the last character, not being a NUL. Currently it does not seem to be an issue because all callers are passing a short desc but it might become a risk later on.

include/wolfboot_efi.h
```diff
@@ -0,0 +1,76 @@
...
+ * for an EFI LoadedImage's LoadOptions. The TLV is covered by the image
+ * signature, so this must be called only AFTER the image has been verified.
+ *
+ * `manifest` points at the wolfBoot header (i.e. boot_addr - IMAGE_HEADER_SIZE).
+ * Returns NULL and sets *out_bytes = 0 when the image carries no command-line
+ * TLV (the caller then boots with no command line, so the kernel uses its
+ * built-in CONFIG_CMDLINE). The returned buffer is allocated with
+ * BS->AllocatePool (EfiLoaderData) and includes the terminating NUL in
+ * *out_bytes (a byte count). */
+static CHAR16 *wolfBoot_efi_get_cmdline(const uint8_t *manifest, UINTN *out_bytes)
```
danielinux: please make this  static inline so it's not generating a warning when unused

arch.mk
```diff
@@ -2196,6 +2238,19 @@ endif
...
+  # fdt.o/gpt.o (like x86_64_efi). update_ram.o is set in the block above.
+  # DEBUG=1: route wolfBoot_printf to the UEFI console (gnu-efi Print).
+  ifeq ($(DEBUG),1)
+    CFLAGS += -DWOLFBOOT_DEBUG_EFI=1
+  endif
+  # Drop -Werror for this target: WOLFBOOT_DEBUG_EFI pulls gnu-efi headers into
+  # every TU and efidebug.h redefines the -DDEBUG object macro as a function
+  # macro. That cpp macro-redefinition warning has no -W name, so it can't be
+  # scoped with -Wno-error=<name>; our own sources are kept warning-clean.
+  CFLAGS := $(filter-out -Werror,$(CFLAGS))
```
danielinux: Is this possible to move this under ifeq($(DEBUG),1) section above since it only collides with efidebug.h

tools/scripts/aarch64-efi-qemu.sh
```diff
@@ -0,0 +1,58 @@
...
+
+for f in "$CODE" "$VARS_SRC"; do
+    [ -r "$f" ] || { echo "Missing $f -- install qemu-efi-aarch64"; exit 1; }
+done
+command -v qemu-system-aarch64 >/dev/null || { echo "Install qemu-system-arm"; exit 1; }
+
+mkdir -p "$ESP"
+cp "$ROOT/wolfboot.efi" "$ESP"/
+cp "$ROOT/aarch64_efi-stage/kernel.img" "$ESP"/ 2>/dev/null || true
+# optional Linux kernel command line (M3): read by wolfboot as \cmdline.txt
```
danielinux: I think command line is now handled differently, there's no cmdline.txt mechanism anymore. Comment and line below should be dropped.

danielinux (changes requested):

danielinux (approved):

PR #824, opened 2026-07-10
Add Raspberry Pi Compute Module 4 (BCM2711) target + wolfCrypt FIPS 140-3

dgarske:
## Summary

Adds a wolfBoot target for the Raspberry Pi Compute Module 4 (CM4) - Broadcom BCM2711, quad-core Cortex-A72 (AArch64). wolfBoot replaces the second-stage OS loader (e.g. U-Boot): the VideoCore firmware loads wolfBoot as `kernel8.img`, and wolfBoot verifies the signed application image before booting it, extending the platform root of trust into the OS. wolfBoot can boot from a RAM-appended image or read A/B images from the onboard eMMC / microSD, and can optionally perform the verification inside the wolfCrypt FIPS 140-3 module.

## CM4 target and authenticated boot

- New `cm4` HAL (`hal/cm4.c/.h/.ld`) for BCM2711, with EL2 RAM-boot image handoff.
- wolfBoot is built as an AArch64 Linux kernel image: `src/boot_aarch64_start.S` prepends the 64-byte ARM64 image header (`"ARM\x64"` magic) and `hal/cm4.ld` links at `0x200000`. The VideoCore firmware only transfers control to a 64-bit kernel that carries this header, and runs it in place at the 2 MB-aligned `0x200000` (it does not relocate a header image to `0x80000`).
- Console is the BCM2711 **mini-UART** (`0xFE215040`, Linux `ttyS0`) on GPIO14/15, inherited from the firmware; build with `-DCM4_UART_PL011` to use the PL011 instead on boards that route it to the header.
- Cortex-A72 build support in `arch.mk` (`-mcpu=cortex-a72+crypto`, `-mstrict-align` for MMU-off operation).
- Example config `config/examples/cm4.config` (RAM boot, `SIGN=ECC384 HASH=SHA384`).

Root of trust:

```
BCM2711 boot ROM -> SPI EEPROM -> VideoCore firmware -> wolfBoot (kernel8.img)
   -> verify (ECDSA P-384 / SHA-384) -> application
```

## eMMC / SD A/B disk boot

- `config/examples/cm4_emmc.config` (onboard eMMC) and `config/examples/cm4_sdcard.config` (microSD) drive the BCM2711 EMMC2 controller (`0xFE340000`) through the generic SDHCI driver (`src/sdhci.c` + `hal/cm4.c` register glue) to read A/B signed images from GPT partitions. wolfBoot parses the GPT, selects the higher-version image, verifies it, ELF-loads it (`ELF=1`) to `WOLFBOOT_LOAD_ADDRESS`, and boots.
- The eMMC path is hardware-validated (below). The microSD path shares the same driver but needs a CM4 Lite - an eMMC module disables the carrier's microSD slot.

## wolfCrypt FIPS 140-3

- `FIPS=1` build option (`options.mk`) rebuilds the wolfcrypt object list as the validated module boundary in link order (`wolfcrypt_first` ... `fips`/`fips_test` ... `wolfcrypt_last`), pointing `WOLFBOOT_LIB_WOLFSSL` at a FIPS wolfSSL tree. The block guards impossible builds early (`FIPS=1` requires `NO_ARM_ASM=1` and a FIPS-approved `SIGN`).
- `include/user_settings.h` FIPS gate: enables the module's algorithm set, keeps the RNG/DRBG enabled, and provides a DRBG entropy seed (`/dev/urandom` on sim, BCM2711 RNG200 hardware TRNG on CM4).
- `src/loader.c` runs the power-on self-test at boot (`fipsEntry`, since bare-metal has no `.init_array`) and refuses to boot unless the module is operational; on a self-test failure it prints the runtime in-core hash to seal into `verifyCore[]`.
- Configs: `config/examples/sim-fips.config`, plus an off-by-default `FIPS=1` block in `config/examples/cm4.config` / `config/examples/cm4_emmc.config`.

## Documentation

- `docs/Targets.md`: CM4 section covering the ARM64 header + `0x200000` load, mini-UART console, build, signing, `config.txt`, flashing (microSD / eMMC via rpiboot), eMMC A/B disk boot, and boot output.
- `docs/FIPS.md`: building with FIPS, DRBG entropy, in-core hash sealing (including the one-pass reseal of `fips_test.o` only), and bare-metal notes.

## Testing

- Builds clean with both `aarch64-none-elf-` and `aarch64-linux-gnu-`.
- CI (`test-configs.yml`): `cm4_test`, `cm4_sdcard_test`, `cm4_fips_test` (build-only FIPS on aarch64), and `sim_fips_test` (host FIPS build); host SDHCI-glue unit test (`tools/unit-tests/unit-cm4-sdhci.c`).
- Signing + verify/update state machine validated on the native `sim` target with `SIGN=ECC384 HASH=SHA384`.
- **FIPS 140-3 on the simulator: fully validated** - `sim-fips.config` builds, the module reaches operational state (in-core integrity + POST/CASTs pass), and wolfBoot performs FIPS-backed integrity + signature verification and A/B updates.
- **eMMC A/B disk boot validated end-to-end on real CM4 hardware** (BCM2711, eMMC, CM4-IO-BASE-B): SDHCI/eMMC card init -> GPT parse -> A/B version select -> SHA-384 integrity -> ECDSA-P384 signature verify -> ELF load -> boot of a signed payload. A/B behavior confirmed on hardware: higher-version select, same-version failover (corrupt A -> boot B), and anti-rollback (higher fails -> refuse lower).
- **FIPS 140-3 on CM4 hardware: fully validated** - after sealing the in-core hash, wolfBoot reports `FIPS 140-3 module operational` and the module gates the eMMC disk boot with SHA-384 integrity and ECDSA-P384 signature verification (entropy from the BCM2711 RNG200 TRNG).

```
wolfBoot CM4 (BCM2711 Cortex-A72) hal_init, EL2
FIPS 140-3 module operational
...
Versions, A:1 B:2
Attempting boot from P:B
Checking image integrity...done
Verifying image signature...done
Firmware Valid.
...
>>> CM4 DISK APP OK: wolfBoot read + verified + booted from eMMC <<<
```

## Notes

- FIPS requires a licensed/validated (or FIPS-ready) wolfSSL tree; see `docs/FIPS.md`. A production, CMVP-validated deployment additionally requires the licensed validated bundle at the validated revision.


jackctj117 (dismissed):

options.mk
```diff
@@ -1835,3 +1835,48 @@ endif
...
+# ---------------------------------------------------------------------------
+# wolfCrypt FIPS 140-3 module (FIPS=1)
+#
+# Point WOLFBOOT_LIB_WOLFSSL at an unpacked FIPS / FIPS-ready wolfSSL tree.
+# The in-core integrity hash on GCC/ELF is enforced by LINK ORDER:
+# wolfcrypt_first.o must be first and wolfcrypt_last.o last, with the FIPS
+# boundary (crypto + fips.o + fips_test.o) between them. We therefore rebuild
+# WOLFCRYPT_OBJS from scratch in that order, replacing the piecemeal per-SIGN
+# selection above (the SIGN/HASH CFLAGS remain in effect). See docs/FIPS.md.
+ifeq ($(FIPS),1)
```
danielinux, line 1899: Suggest also to add the following to the `ifeq ($(FIPS),1)` block to intercept impossible build cases:

```
  ifneq ($(NO_ARM_ASM),1)
    $(error FIPS=1 requires NO_ARM_ASM=1 (the FIPS module uses portable-C crypto))
  endif
  ifeq ($(filter $(SIGN),ECC256 ECC384 ECC521 RSAPSS2048 RSAPSS3072 RSAPSS4096),)
    $(error FIPS=1 requires a FIPS-approved SIGN (see docs/FIPS.md); got $(SIGN))
  endif
```

config/examples/sim-fips.config
```diff
@@ -0,0 +1,22 @@
+# wolfBoot simulator build using the wolfCrypt FIPS 140-3 module.
```
danielinux, line 1: This config does not have any corresponding CI job

danielinux (changes requested):
A couple of minor things. Also the PR description seems outdated (says CM4 FIPS validation is work in progress but it seems complete here)

hal/cm4.c
```diff
@@ -0,0 +1,886 @@
...
+        wolfBoot_printf("cm4: firmware DTB size %u out of range\n",
+            (unsigned)sz);
+        return NULL;
+    }
+    /* SECURITY: the firmware DTB lives on the unsigned FAT partition (unless
+     * the RPi EEPROM secure-boot is enabled), so it is NOT covered by wolfBoot's
+     * signature. Only /chosen/bootargs is overwritten below; /memory,
+     * /reserved-memory, per-device reg windows and initrd remain firmware /
+     * attacker controlled. This path is opt-in via CM4_FIRMWARE_DTB. See the
+     * trust-boundary note in docs/Targets.md. */
```
danielinux: This comment ends "See the trust-boundary note in docs/Targets.md", but the new CM4 section of `docs/Targets.md` contains no such note — zero hits for *unsigned*, *unverified*, *not covered*, *attacker* or *trust boundary* in the added text. It documents the DTB relocation and the bootargs injection as plain features.

Compounding it, the comment above frames this as "opt-in via CM4_FIRMWARE_DTB", but `CM4_FIRMWARE_DTB` is enabled by default in both shipped Linux configs (`cm4_emmc_linux.config:68`, `cm4_emmc_rauc.config:56`). A user following the documented Linux recipe gets the unverified-DTB path without ever seeing the caveat.

Either write the note or stop pointing at it.

hal/cm4.c
```diff
@@ -0,0 +1,886 @@
...
+    /* fdt_check_header() does not validate totalsize; bound it so a corrupt DTB
+     * header cannot make the copy/fixup clobber memory past the DTS landing
+     * zone (leaving room for the fixup headroom too). */
+    if (sz < sizeof(struct fdt_header) ||
+            sz > CM4_FDT_MAX_SIZE - WOLFBOOT_FDT_FIXUP_HEADROOM) {
+        wolfBoot_printf("cm4: firmware DTB size %u out of range\n",
+            (unsigned)sz);
+        return NULL;
+    }
+    /* SECURITY: the firmware DTB lives on the unsigned FAT partition (unless
```
danielinux, line 379: This is the same trust boundary as Fenrir #7998, and master now carries the mechanism to close it: PR #852 merged on 2026-08-13, so `HDR_DEVICE_TREE_DIGEST` and `sign --dts` are available specifically to bind a raw DTB to the signed image.

`/memory`, `/reserved-memory`, per-device `reg` windows and `initrd` all reach the kernel unverified here. I suspect the real answer is that CM4 *cannot* use the digest — VideoCore patches RAM size, mini-UART clock and serial number into the DTB at runtime, so its hash is not predictable at signing time — but that reasoning should be stated in the PR and the docs rather than left implicit. If that is the reason, say so and point at the RPi EEPROM secure-boot as the only thing that actually closes it.

hal/cm4.c
```diff
@@ -0,0 +1,886 @@
...
+         * "slot B good" system to "boot A"). */
+        wolfBoot_printf("cm4: uboot-env read failed (%d); default slot, no writeback\n",
+            r);
+        memset(cm4_uboot_env, 0, sizeof(cm4_uboot_env));
+    }
+    if (uboot_env_select_slot(cm4_uboot_env, UBOOT_ENV_SIZE, &slot) != 0)
+        return 0; /* malformed BOOT_ORDER / redundant env -> static cmdline */
+    if (!slot.selected) {
+        /* All slots were exhausted; counters were re-armed - pick slot A now. */
+        wolfBoot_printf("cm4: all RAUC slots exhausted; re-armed, retrying\n");
```
danielinux, line 295: The header contract and its only caller disagree. `include/ubootenv.h:88` says that when all counters are exhausted the caller "should reboot". This instead calls `uboot_env_select_slot()` a second time and boots slot A.

A board with both slots broken therefore boots A forever, cycling 3 → 2 → 1 → 0 → re-arm → 3, and the try counter never escalates to anything. That may well be the right call for a bootloader with nowhere else to go, but then `ubootenv.h` should describe that behaviour instead of a reboot that never happens.

hal/cm4.c
```diff
@@ -0,0 +1,886 @@
...
+    /* Persist the decremented counter only when the env read a CRC-valid image.
+     * A bad-CRC full read leaves slot.reinitialized set: booting the default is
+     * fine, but writing the wiped buffer back would destroy the last-good RAUC
+     * state, so skip the writeback (see src/ubootenv.c). A short/failed write
+     * leaves the counter un-decremented, so fall back to the static cmdline. */
+    if (env_ok && !slot.reinitialized) {
+        r = disk_part_write(BOOT_DISK, CM4_UBOOT_ENV_PART, 0, UBOOT_ENV_SIZE,
+            cm4_uboot_env);
+        if (r != (int)UBOOT_ENV_SIZE) {
+            wolfBoot_printf("cm4: uboot-env write failed (%d); static cmdline\n",
```
danielinux: A failed env write silently disables failover: the function returns 0, the boot falls back to the static command line, the counter is never decremented, and the same slot is chosen again next boot.

On a worn or read-only env partition, A/B redundancy stops working and the only evidence is this one printf — which is the exact failure mode the feature exists to prevent. Worth making this louder, or at minimum documenting it.

src/sdhci.c
```diff
@@ -1618,6 +1655,24 @@ static int sdhci_transfer(int dir, uint32_t cmd_index, uint32_t block_addr,
...
+     * the card actually finished the program cycle (CMD13 -> READY_FOR_DATA via
+     * sdhci_wait_busy) before reporting success. A silently timed-out write must
+     * NOT be treated as landed - the RAUC try-counter writeback relies on a
+     * failed write propagating (see hal/cm4.c). Safe for the CM4 EMMC2 where TC
+     * legitimately never precedes CMD12: after CMD12 the card reports ready. */
+    if (write_settle_timeout) {
+        wolfBoot_printf("sdhci_transfer: multi-block write settle timeout "
+            "(SRS12=0x%08X); confirming completion via CMD13\n", reg);
+        if (status == 0) {
+            status = sdhci_wait_busy(0);
```
danielinux: Duplicate on the non-error path: line 1650 already ran `status = sdhci_wait_busy(0);` after CMD12 when `status == 0`, so a bounded-out multi-block write with no SRS12 error issues a second CMD13 here.

Harmless in itself, but the comment above ("require positive evidence the card actually finished the program cycle") reads as if this is the only confirmation, which will mislead the next reader.

src/sdhci.c
```diff
@@ -1568,10 +1585,30 @@ static int sdhci_transfer(int dir, uint32_t cmd_index, uint32_t block_addr,
...
+#endif
             while (((reg = SDHCI_REG(SDHCI_SRS12)) &
-                (SDHCI_SRS12_TC | SDHCI_SRS12_EINT)) == 0);
+                    (SDHCI_SRS12_TC | SDHCI_SRS12_EINT)) == 0) {
+#ifdef SDHCI_WRITE_SETTLE_SPINS
+                /* Bounded (opt-in): record the timeout rather than discarding
+                 * it silently, so the post-transfer path re-confirms the card
+                 * before declaring success. */
+                if (is_multi_block && spins-- == 0) {
+                    write_settle_timeout = 1;
```
danielinux, line 1614: The bound only covers multi-block. With `SDHCI_WRITE_SETTLE_SPINS` defined, a wedged *single*-block write still spins forever, because the break is gated on `is_multi_block`.

That is deliberate per the comment about CMD24 setting TC after the data phase, but the macro name and the surrounding "bounded instead of spinning forever" framing say otherwise. One clarifying clause would settle it.

src/ubootenv.c
```diff
@@ -0,0 +1,371 @@
...
+
+static void wr_le32(uint8_t *p, uint32_t v)
+{
+    p[0] = (uint8_t)(v);
+    p[1] = (uint8_t)(v >> 8);
+    p[2] = (uint8_t)(v >> 16);
+    p[3] = (uint8_t)(v >> 24);
+}
+
+/* Parse a small non-negative decimal; -1 on a non-numeric value. */
```
danielinux: `env_atol` accepts trailing garbage despite this comment promising -1 on a non-numeric value — only the first character is checked, so `"1x"` parses as 1. Verified: it selects the slot and rewrites the counter to `0`.

Related, a counter longer than 15 digits is silently truncated by the `val_max = 16` clamp in the caller and then written back shorter (`9999999999999999` → `999999999999998`).

Both are cosmetic for RAUC's single-digit counters; the issue is the comment claiming behaviour the code does not have.

src/ubootenv.c
```diff
@@ -0,0 +1,371 @@
...
+        return -1;
+    v = 0;
+    while (*s >= '0' && *s <= '9') {
+        v = (v * 10) + (long)(*s - '0');
+        s++;
+    }
+    return v;
+}
+
+/* Format a non-negative decimal into out (caller supplies >= 16 bytes). */
```
danielinux: The stated contract is one byte short of what this can write. For a value with 16 or more digits the `tmp[16]` loop fills all 16 slots and `out[16] = '\0'` then writes a 17th byte.

Nothing reaches it today only because the single caller's buffer is fed through `uboot_env_get(..., 16)`, which clamps to 15 digits — a coupling across two functions that nothing documents or enforces. Bounding `tmp` to 20 and documenting `out >= 21`, or rejecting over-long values in `env_atol`, makes it self-consistent.

danielinux (changes requested):

danielinux (approved):

PR #830, opened 2026-07-22
sbom: generate SBOMs from every wolfBoot build system

sameehj:
### Description

wolfBoot ships as source. Users build it in many ways. Each user needs an
SBOM for the build that the user makes. This PR makes wolfBoot produce a
CycloneDX 1.6 and SPDX 2.3 SBOM from every build system, and it makes the
SBOM reproducible.

One engine does the work. Each build system gives the engine two inputs:
the compiled source list and the build configuration. The engine captures
the configuration with the host compiler, so the SBOM is the same for GCC,
Clang, LLVM, IAR, armcl, CCRX, and XC32.

### What is new

- A shared driver: `tools/scripts/wolfboot-sbom.sh`.
- Front ends for each build system:
  - Make and arch.mk and vendor SDKs: `make sbom`.
  - CMake and the Pico SDK: `cmake --build --target sbom`.
  - IAR Embedded Workbench: `ide-sbom/iar_sbom.py`.
  - Any IDE with a compilation database: `ide-sbom/compdb_sbom.py`.
  - TI CCS, MPLAB X, Renesas, Xilinx: `ide-sbom/route_through_sbom.sh`.
  - Per-HAL component: `make sbom-hal`.
  - Zephyr module: `ide-sbom/zephyr_sbom.py`.
- A validator: `ide-sbom/validate_sbom.py`.
- A CI canary: `.github/workflows/test-sbom.yml`.
- Documentation: `docs/SBOM.md` and `docs/SBOM-WOLFGLASS.md`.

### Reproducibility fix

The driver captures build macros with the host compiler. Some macros hold
an absolute host path. For example, `arch.mk` passes
`-DPICO_SDK_PATH=$(PICO_SDK_PATH)`. Before this change, the path entered
the SBOM. The SBOM was then different on each machine, and it leaked the
local file system.

The driver now redacts each absolute path in the captured macros. It keeps
the macro name, so the configuration record stays complete. The CI canary
asserts that no path leaks and that the macro name stays. Use `--no-scrub`
for debug only.

### Toolchain support

The SBOM content does not depend on the cross-compiler. The config uses
the host compiler. The composition is a source list. So Clang, LLVM,
Renesas CCRX, TI armcl, and IAR all give the same SBOM. A new toolchain
needs no work. A new IDE that emits a compilation database needs no work.

### Relation to wolfGlass

The tools are product-neutral. They are self-contained in wolfBoot on
purpose. They are the reference implementation of the shared SBOM layer
for every wolfSSL product (wolfSSL, wolfHSM, and others). `docs/
SBOM-WOLFGLASS.md` gives the file-by-file migration map and the list of
changes that wolfGlass must cover. No wolfBoot logic changes when the
tools move.

### Prerequisite

The tools need `gen-sbom` from wolfSSL (`lib/wolfssl/scripts/gen-sbom`).
If the pinned submodule does not carry it, pass a path with `GEN_SBOM=...`
or the `--gen-sbom` option. The CI canary fetches it from wolfSSL master
as a fallback.

### Testing

- `make sbom TARGET=sim` produces a valid SBOM. The validator passes.
- `make sbom-hal TARGET=sim` produces a valid per-HAL SBOM.
- `cmake --build <dir> --target sbom` produces a valid SBOM.
- The IAR, compdb, and Zephyr extractors produce valid SBOMs.
- The scrub check confirms that no absolute path enters the SBOM and that
  the macro name stays.

### Notes

- wolfBoot embeds wolfCrypt sources. The SBOM lists these as wolfBoot
  sources, not as a separate component.
- A vendor SDK build lists only the source files on disk.
- The driver is a POSIX shell script. On Windows, run it in WSL, MSYS, or
  Git Bash. As an alternative, use `compdb_sbom.py`.


sameehj:
## Update: re-vendor wolfGlass tip `9bdf5b7`

Syncs `tools/sbom/` to the pushed wolfGlass master tip and hardens the Make path:

- `sbom.mk`: honest `WOLFSSL_DIR`/`version.h` `--dep-version` fallback (with `$$$$` for `$(call)`/`$(eval)` double-expansion); documents that `--cflags` keeps only `-D` tokens
- `validate_sbom.py`: `--min-properties` / `--require-dep-version`
- CI Make-path validates with `--min-properties 20` so an empty capture cannot pass a gcc≡clang diff

`.wolfglass-rev` now matches wolfGlass `9bdf5b7`.

danielinux:
@sameehj please rebase on latest master

lib/wolfssl
```diff
@@ -1 +1 @@
-Subproject commit ac01707f552c611fbd135cc723b2682b3e7f80f2
+Subproject commit 887f242ee8570f7d8403e002c5b2b88929b86544
```
danielinux: This sets `lib/wolfssl` to `887f242e` (2026-05-25). `master` is currently at `5418d6cf` (2026-07-16), and the GitHub compare API reports `master` 1122 commits ahead of `887f242e` with 0 behind, so merging as-is rolls the submodule back by about seven weeks.

It also looks unnecessary now that `gen-sbom` is vendored under `tools/sbom/`. I checked the pinned tree and `scripts/gen-sbom` is not present in it, so nothing in this PR reads the generator from the submodule.

Suggest dropping the submodule change from the PR.

Makefile
```diff
@@ -799,71 +799,84 @@ pico-sdk-info: FORCE
...
 #   CRA_PYTHON             Python interpreter (default: python3)
 
 HOSTCC?=cc
 WOLFBOOT_VERSION:=$(shell sed -n \
     's/.*LIBWOLFBOOT_VERSION_STRING[[:space:]]*"\([^"]*\)".*/\1/p' \
     include/wolfboot/version.h)
-GEN_SBOM?=$(WOLFBOOT_LIB_WOLFSSL)/scripts/gen-sbom
+SBOM_ROOT:=$(WOLFBOOT_ROOT)
+SBOM_NAME:=wolfboot
+SBOM_SRCS=$(wildcard $(patsubst %.o,%.c,$(OBJS))) $(wildcard $(patsubst %.o,%.S,$(OBJS)))
```
danielinux: `$(wildcard)` removes sources that are not on disk before the driver ever sees them, so a missing submodule or SDK silently shrinks the SBOM with no diagnostic anywhere.

Measured with `config/examples/sim-tpm.config`, same TARGET, same config, twice:

| | sources | wolfTPM files | diagnostic | `validate_sbom.py` |
|---|---|---|---|---|
| without `lib/wolfTPM` checked out | 21 | 0 | none | `All SBOMs valid.` |
| with `lib/wolfTPM` checked out | 29 | 8 | none | `All SBOMs valid.` |

Both runs write a file named `wolfboot-2.9.0.cdx.json` carrying the same `serialNumber`, so the two documents are not distinguishable by identity either (see the `gen-sbom` comment on that).

The driver already has a warning for exactly this case at `tools/sbom/sbom-driver.py:212`, but `--skip-missing` never fires on the Make path because `$(wildcard)` has already filtered the list. That is the path every embedded user takes.

An object that maps to no source on disk is the signal. Something like:

```make
SBOM_SRCS=$(wildcard $(patsubst %.o,%.c,$(OBJS))) $(wildcard $(patsubst %.o,%.S,$(OBJS)))
# Every object must map to a source that exists. If one does not, a submodule or
# vendor SDK is absent and the SBOM would under-report the image it describes.
SBOM_SRCS_MISSING=$(filter-out $(basename $(SBOM_SRCS)),$(basename $(OBJS)))
```

then fail the recipe when `SBOM_SRCS_MISSING` is non-empty unless the caller sets `SBOM_ALLOW_MISSING=1`, and in that case record the dropped count as a `wolfssl:sbom:sources-missing` property so the omission is visible in the document rather than only on a terminal nobody kept.

tools/sbom/sbom-driver.py
```diff
@@ -0,0 +1,436 @@
...
+def load_srcs(srcs_file, skip_missing):
+    """Read the source list; optionally drop paths that do not exist."""
+    kept, dropped = [], 0
+    with open(srcs_file, encoding="utf-8", errors="replace") as f:
+        for line in f:
+            s = line.strip()
+            if not s or s.startswith("#"):
+                continue
+            if skip_missing and not os.path.isfile(s):
+                print(f"WARNING: skipping missing source: {s}", file=sys.stderr)
```
danielinux, line 212: This warning is unreachable from `make sbom`, which is the primary path for every embedded target. `Makefile:826` filters the source list through `$(wildcard)` first, so `load_srcs()` only ever sees files that exist and `--skip-missing` never drops anything. Detail in the Makefile comment.

tools/sbom/gen-sbom
```diff
@@ -0,0 +1,1803 @@
...
+                file=sys.stderr)
+        lib_hash = srcs_merkle_hash(srcs)
+        hash_kind = 'source-merkle-omnibor'
+        hash_source = 'srcs'
+        srcs_basenames = sorted({os.path.basename(p) for p in srcs})
+
+    dt, timestamp = build_timestamp()
+    year = dt.year
+    serial = derived_uuid(args.name, args.version, 'serial')
+    doc_ns_uuid = derived_uuid(args.name, args.version, 'document')
```
danielinux: `serialNumber` and the SPDX `documentNamespace` are derived from name and version only, so every wolfBoot configuration at a given version produces the same document identity.

`sim` vs `stm32h7`, both 2.9.0, `SOURCE_DATE_EPOCH` pinned:

```
sim serial   urn:uuid:6dc51759-ade2-5967-b392-2f23ed0eb4df
h7  serial   urn:uuid:6dc51759-ade2-5967-b392-2f23ed0eb4df
sim SPDX ns  urn:uuid:54060fbb-3fa0-5085-96ef-cc44675172e2
h7  SPDX ns  urn:uuid:54060fbb-3fa0-5085-96ef-cc44675172e2
identical namespace: True
identical content:   False
```

The two documents have different source sets (18 vs 28 files) and different component hashes. CycloneDX requires `serialNumber` to be unique per BOM, and SPDX 2.3 section 6.5 requires the same of `documentNamespace`. Dependency-Track keys projects on `serialNumber`, so ingesting a second target overwrites the first rather than adding it. wolfBoot has on the order of 100 target configurations, so this is the normal case rather than a corner.

The composition hash is already computed above and is the right discriminator:

```suggestion
    serial = derived_uuid(args.name, args.version, 'serial', lib_hash)
    doc_ns_uuid = derived_uuid(args.name, args.version, 'document', lib_hash)
```

The default output filenames collide for the same reason: both targets write `wolfboot-2.9.0.cdx.json`. `SBOM_CDX_OUT` / `SBOM_SPDX_OUT` in the Makefile want a config discriminator too, for example `wolfboot-$(TARGET)-$(SIGN)-$(WOLFBOOT_VERSION).cdx.json`.

tools/sbom/gen-sbom
```diff
@@ -0,0 +1,1803 @@
...
+#
+# Every entry carries both machine-resolvable identifiers, because the two
+# scanner families do not agree on one.  PURL serves the ecosystem scanners
+# (OSV, GHSA, Trivy, Dependency-Track); CPE serves NVD, which is what a CRA /
+# IEC 62443 vulnerability-monitoring process keys on.  A dependency with only
+# a PURL is invisible to a CPE-driven scan, so wolfSSL advisories never reach
+# the integrator of a product that embeds wolfSSL.  Each `cpe` value must be
+# the vendor:product pair NVD actually registers for that dependency; never
+# synthesize one.
+DEP_META = {
```
danielinux, line 357: `DEP_META` covers wolfssl, wolfcrypt, libz and openssl, and the driver exposes only `--dep-wolfssl`, `--dep-wolfcrypt` and `--dep-openssl`. wolfBoot has six submodules: wolfssl, wolfTPM, wolfPKCS11, wolfHSM, wolfPSA and wolfHAL. The five that are not wolfssl can never be emitted as components.

Built `config/examples/sim-tpm.config` with `lib/wolfTPM` checked out:

* the captured config records `WOLFBOOT_TPM=1`, `WOLFTPM_SWTPM=1`, `WOLFTPM2_MAX_BUFFER=1500`, so the document knows perfectly well that a TPM stack is compiled in
* `tpm2.c`, `tpm2_wrap.c`, `tpm2_tis.c`, `tpm2_packet.c`, `tpm2_param_enc.c`, `tpm2_swtpm.c`, `tpm2_util.c` and `tpm2_crypto.c` are folded into wolfBoot's own source set
* `components[]` is `wolfssl -> wolfcrypt` only. No wolfTPM component, no version, no CPE, no PURL.

A wolfTPM advisory cannot match that document. The same holds for wolfHSM and wolfPKCS11 builds. This is the specific thing CRA Annex I Part II(1) asks for ("covering at the very least the top-level dependencies of the product"), so it is worth closing rather than deferring.

The detection signal is already present. Suggested shape: make `DEP_META` a table keyed by name with the same fields it has now, add a generic `--dep NAME` alongside the existing `--dep-version NAME=VER`, and have `sbom.mk` declare the component whenever the corresponding macro (`WOLFBOOT_TPM`, `WOLFHSM_CLIENT`, `WOLFBOOT_PKCS11` and so on) appears in the captured config.

Two related exclusions are worth stating explicitly in the `docs/SBOM.md` Limitations section rather than leaving implicit: the compiler runtime (libgcc, newlib) is linked into every image and appears nowhere in the document, and vendor SDK sources are absent whenever the SDK lives outside the source tree.

tools/sbom/gen-sbom
```diff
@@ -0,0 +1,1803 @@
...
+        {'name': 'wolfssl:sbom:hash-source', 'value': hash_source})
+    if hash_source == 'none':
+        properties.append(
+            {'name': 'wolfssl:sbom:no-artifact-hash-note',
+             'value': _NO_HASH_NOTE})
+    if srcs_basenames:
+        properties.append({
+            'name': 'wolfssl:sbom:source-set',
+            'value': ','.join(srcs_basenames),
+        })
```
danielinux, line 1197: This property is the only file-level record in the document, and it is a comma joined list of basenames: no path, no per file hash, no per file license. SPDX comes out with `files: 0` and `filesAnalyzed: false`.

Two consequences. You cannot answer "which wolfssl file revisions are in this image", which is the question an incident response actually asks. And the basename collapse is ambiguous wherever a name repeats: `sha256.c` in a wolfBoot SBOM could be `lib/wolfssl/wolfcrypt/src/sha256.c` or a HAL file. The Merkle root cannot be decomposed to check one file.

The inputs already exist. `srcs_merkle_hash` computes a gitoid per source, and `generate_cdx` already has a `file_entries` path (line 1174) that the `--lib` route uses. Emitting SPDX `files[]` with repo-relative `fileName` plus SHA-256 and `CONTAINS` relationships, and reusing `file_entries` for the source route, closes this cheaply.

Related: `srcs_merkle_hash` calls `sys.exit` on any duplicate basename, so a target that pairs a vendor `flash.c` with another `flash.c` fails SBOM generation outright rather than degrading. Sorting on the repo-relative path instead of the basename would remove both the ambiguity and the hard failure.

docs/SBOM.md
```diff
@@ -0,0 +1,367 @@
...
+
+```
+wolfssl:sbom:wolfssl-subset=wolfcrypt-only
+wolfssl:sbom:wolfssl-subset-basis=captured
+```
+
+The `wolfssl` component stays in the document regardless. Dropping it would
+read as more precise and would take the scan from every wolfSSL advisory to
+none. Narrow the TLS-only CVEs with a VEX statement instead, which is the
+mechanism designed to say "present but not exploitable here".
```
danielinux, line 167: This paragraph and the matching comment in `gen-sbom` (`resolve_crypto_only`) both direct the reader to narrow TLS-only wolfSSL CVEs with a VEX statement, but nothing in this PR produces one and there is no target that does.

`include/user_settings.h` sets `WOLFCRYPT_ONLY` for every configuration except a wolfHSM server with certificate chain verification, so essentially every wolfBoot user's scanner will report wolfSSL TLS advisories against an image that contains no TLS. Recommending the remedy without shipping it leaves that noise in place for everyone.

This looks like the highest value follow-up and a small one: a `make vex` emitting a CycloneDX VEX with `not_affected` / `code_not_present` for the TLS-only CVEs, keyed off the `WOLFCRYPT_ONLY` the document already captures. Worth filing as a follow-up issue even if it stays out of this PR.

tools/sbom/.wolfglass-rev
```diff
@@ -0,0 +1 @@
+1bfcf4f1a293ba09f0ff6d67904dca09ee8eb6d7
```
danielinux: `tools/sbom/gen-sbom` is a 1803 line copy pinned here by a bare SHA, with nothing checking that the copy still matches the pin. It will drift from wolfGlass and from wolfSSL's own `scripts/gen-sbom` without anything noticing, and a locally patched copy would be invisible.

Worth a step in `test-sbom.yml` that fetches this revision and diffs it against `tools/sbom/`, so a stale or modified vendored copy fails CI rather than quietly producing a different document from the one wolfSSL ships.

danielinux (commented):
Reviewed against head `aafafcb4`. Built SBOMs locally for `sim`, `stm32h7` and `sim-tpm` with `SOURCE_DATE_EPOCH=1700000000` and compared the outputs.

The architecture is the right one. Generating per build from `OBJS` rather than shipping a canned release SBOM is correct for wolfBoot, since every integrator's image is a different composition and a static document would be wrong for all of them. Three things here are clear improvements over the previous recipe:

* The config capture is now accurate. The old `make sbom` ran `$(HOSTCC) -dM -E` on the bare `-D` list with no `-include`, so `include/user_settings.h` was never evaluated and `HAVE_ECC` (derived from `WOLFBOOT_SIGN_ECC256`) was absent. The SBOM described a secure bootloader with no signature algorithm. `SBOM_SETTINGS_H` fixes that: a `sim` build now records 230 config macros.
* The absolute path scrub keeps `-DPICO_SDK_PATH=<abs path>` out of the document while preserving the macro name, and CI asserts both halves on a real rp2350 build rather than only on a synthetic command line.
* The `registered` / `pending` split in `PRODUCT_CPE` is the right discipline. I checked the NVD API: `cpe:2.3:a:wolfssl:wolfboot:2.9.0:*:*:*:*:*:*:*` is in the dictionary (created 2026-08-10, 33 entries), so the emitted CPE resolves.

Comments inline. Of those, the `lib/wolfssl` change and the SBOM identity collision are the two I would want resolved before this merges; the `$(wildcard)` drop and the missing sibling-library components are the two that decide whether the output actually satisfies "top-level dependencies".

lib/wolfssl
sameehj: Dropped. The branch is rebased on master and the submodule now pins 5418d6cf, identical to master, so the change is gone from the diff entirely. You were right that nothing here reads the generator from the submodule — GEN_SBOM now resolves to the vendored tools/sbom/gen-sbom.

Makefile
sameehj: Implemented as sbom-check-sources in ec4f8334. Your sim-tpm case now fails and names all eight tpm2*.c files instead of quietly emitting 21 sources. SBOM_ALLOW_MISSING=1 downgrades it to a warning.

Two things your snippet surfaced that are worth recording. $(wildcard) could not be kept even for the list itself: GNU Make 3.81 caches directory contents, so any source generated during the run stays invisible for the rest of it. The list is now built with $(shell), which re-reads the filesystem. And src/keystore.c is generated from the signing key, so on a fresh checkout the guard read its absence as an absent submodule and failed every Make-path CI job. It is now a prerequisite, so make sbom generates it exactly as a build does, and the trust anchor is no longer the one file missing from a bootloader's SBOM.

The wolfssl:sbom:sources-missing property is not done. It needs a gen-sbom flag that does not exist, and Mark asked that the vendored copy not be patched locally, so I am filing it as a wolfGlass issue and will pick it up on the re-vendor.

tools/sbom/sbom-driver.py
sameehj: The Make path no longer reaches the driver with a silently shortened list; sbom-check-sources stops the build first and names the objects. The driver-side --skip-missing warning stays as the second net for the non-Make routes. The line itself is vendored, so I have not touched it.

tools/sbom/gen-sbom
sameehj: The filename half is done in 52c1349c. Output is now wolfboot-<target>-<sign>-<hash>-<version>.cdx.json, so sim and stm32h7 no longer overwrite each other, and the CMake path was changed to match.

The serialNumber and documentNamespace derivation is in vendored gen-sbom. Mark has that in flight as wolfGlass #15, using the composition hash as the discriminator exactly as you suggest, so I have deliberately not patched it locally. It arrives here on the re-vendor. I agree this should land before merge — Dependency-Track keying on serialNumber is the real failure, and the filename change does not address it.

tools/sbom/gen-sbom
sameehj: Agreed, and agreed it should be closed rather than deferred. The generic --dep NAME plus a name-keyed DEP_META is a gen-sbom change, so it is going upstream as a wolfGlass issue rather than a local patch.

The two exclusions you asked to be explicit are in docs/SBOM.md as of a050aeef: the compiler runtime (libgcc, newlib) and vendor SDK sources outside the tree. I also listed the sibling submodules there so the gap is stated rather than implicit until the upstream fix lands.

tools/sbom/gen-sbom
sameehj: Agreed on all three points, including that srcs_merkle_hash should sort on the repo-relative path so a duplicate basename degrades instead of calling sys.exit. All of it is in vendored gen-sbom, so it goes upstream as a wolfGlass issue. Documented as a limitation meanwhile.

docs/SBOM.md
sameehj: Agreed, and the noise is as broad as you describe given WOLFCRYPT_ONLY. Filing it as a follow-up rather than growing this PR.

One constraint worth flagging: gen-sbom has no VEX support at all today, only a docstring mention. wolfGlass keeps VEX in gen-advisory, which wolfBoot does not vendor. So make vex needs a decision on whether to extend gen-sbom or vendor gen-advisory too, which is why the issue goes to wolfGlass first.

tools/sbom/.wolfglass-rev
sameehj: Both halves fixed. The pin was 1bfcf4f, which resolves in no repository; it is now d34a906, whose content matches the vendored tree. A wolfglass_drift job in test-sbom.yml checks out wolfGlass at the pinned revision and runs wolfglass-sync --check, so a stale pin, a drifted copy, or a local patch all fail CI. It passes in 7 seconds. It needs no token, since wolfGlass is public, so it runs on fork pull requests too.

danielinux:
waiting for https://github.com/wolfSSL/wolfGlass/pull/15 to merge

sameehj:
> waiting for [wolfSSL/wolfGlass#15](https://github.com/wolfSSL/wolfGlass/pull/15) to merge

Merged, and synced. Thank you @danielinux

danielinux:
please rebase this again on latest master (trivial .gitignore conflict apparently)

danielinux (approved):

PR #836, opened 2026-07-29
fix unconditional SHA512 opt-in for wolfHSM builds

bigbrett:
Removes unconditional SHA348 and SHA512 opt-in for all wolfHSM server builds and relocates to TZ-specific user_settings.h block


aidangarske (approved):
Thanks!

PR #846, opened 2026-08-06
Add Agilex 5 FCS and wolfBoot integration

aidangarske:
# Description
  
  - Adds the Altera Agilex 5 013B wolfBoot port for the GSRD SD-card boot flow
  - Preserves the existing SPL and TF-A responsibilities for DDR, clocks, resets, PSCI, and EL3
  - Loads wolfBoot as the signed BL33 payload at 0x80200000
  - Adds Agilex 5 SDHCI and platform initialization support
  - Integrates signed Linux FIT images with wolfBoot verification
  - Adds a four-partition WIC layout with initialized A and B image slots
  - Adds Yocto/meta-wolfSSL integration, CI coverage, and customer bring-up documentation
  - Validated the generated WIC through the customer Kas build and on Agilex 5 hardware

partner pr: https://github.com/wolfSSL/meta-wolfssl/pull/177

### DIFF
--- a/hal/agilex5.c
+++ b/hal/agilex5.c
@@ -315,7 +315,7 @@ void sdhci_platform_dma_prepare(void *buf, uint32_t size, int is_write)
         if (is_write != 0)
             __asm__ volatile("dc cvac, %0" : : "r"(addr) : "memory");
         else
-            __asm__ volatile("dc civac, %0" : : "r"(addr) : "memory");
+            __asm__ volatile("dc ivac, %0" : : "r"(addr) : "memory");
     }
     __asm__ volatile("dsb sy" : : : "memory");
 }
@@ -329,7 +329,7 @@ void sdhci_platform_dma_complete(void *buf, uint32_t size, int is_write)
 
     if (is_write == 0) {
         for (addr = start; addr < end; addr += CACHE_LINE_SIZE)
-            __asm__ volatile("dc civac, %0" : : "r"(addr) : "memory");
+            __asm__ volatile("dc ivac, %0" : : "r"(addr) : "memory");
         __asm__ volatile("dsb sy" : : : "memory");
     }
 }
--- a/hal/agilex5.c
+++ b/hal/agilex5.c
@@ -1,0 +1,339 @@
+/* agilex5.c
+ *
+ * Copyright (C) 2026 wolfSSL Inc.
+ *
+ * This file is part of wolfBoot.
+ *
+ * wolfBoot is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * wolfBoot is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1335, USA
+ */
+
+#ifdef TARGET_agilex5
+
+#include <stdint.h>
+
+#include "fdt.h"
+#include "hal.h"
+#include "hal/agilex5.h"
+#include "image.h"
+#include "printf.h"
+
+#ifndef ARCH_AARCH64
+#error "wolfBoot Agilex 5 HAL requires ARCH=AARCH64"
+#endif
+
+#ifndef LINUX_BOOTARGS
+#ifndef LINUX_BOOTARGS_ROOT
+#define LINUX_BOOTARGS_ROOT "/dev/mmcblk0p4"
+#endif
+#define LINUX_BOOTARGS \
+    "earlycon=uart8250,mmio32,0x10c02000 console=ttyS0,115200 " \
+    "root=" LINUX_BOOTARGS_ROOT " rootwait"
+#endif
+
+#ifdef DEBUG_UART
+#define UART_RBR 0U
+#define UART_THR 0U
+#define UART_LSR 5U
+#define UART_LSR_THRE (1U << 5)
+#define UART_LSR_TEMT (1U << 6)
+#define UART_TIMEOUT 1000000U
+#define UART_REG(_n) \
+    (*(volatile uint32_t *)(AGILEX5_UART0_BASE + ((uintptr_t)(_n) << 2)))
+
+void uart_init(void)
+{
+    /* SPL configured the DesignWare APB UART, its clock and pinmux. */
+}
+
+void uart_write(const char *buf, uint32_t len)
+{
+    uint32_t i;
+
+    for (i = 0; i < len; i++) {
+        uint32_t timeout = UART_TIMEOUT;
+        if (buf[i] == '\n') {
+            while (((UART_REG(UART_LSR) & UART_LSR_THRE) == 0U) &&
+                    (--timeout != 0U)) {}
+            UART_REG(UART_THR) = '\r';
+            timeout = UART_TIMEOUT;
+        }
+        while (((UART_REG(UART_LSR) & UART_LSR_THRE) == 0U) &&
+                (--timeout != 0U)) {}
+        UART_REG(UART_THR) = (uint32_t)(uint8_t)buf[i];
+    }
+}
+#else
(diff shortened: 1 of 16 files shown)


danielinux:
@aidangarske please rebase.

src/sdhci.c
```diff
@@ -594,10 +605,20 @@ static int sdhci_send_cmd_internal(uint32_t cmd_type,
...
             cmd_index, cmd_arg, resp_type, SDHCI_REG(SDHCI_SRS12));
         status = -1; /* error */
+        if (sdhci_reset_cmd_line() != 0)
+            wolfBoot_printf("sdhci_send_cmd: command line reset timeout\n");
     }
 
     SDHCI_REG_SET(SDHCI_SRS12, SDHCI_SRS12_CC); /* clear command complete */
-    while ((SDHCI_REG(SDHCI_SRS09) & SDHCI_SRS09_CICMD) != 0);
+    timeout = 0x000FFFFF;
+    while (((SDHCI_REG(SDHCI_SRS09) & SDHCI_SRS09_CICMD) != 0U) &&
```
danielinux, line 666: This trailing inhibit wait is now bounded, but the one at the *top* of `sdhci_send_cmd_internal` is untouched:

```c
/* src/sdhci.c:582 */
while ((SDHCI_REG(SDHCI_SRS09) & SDHCI_SRS09_CICMD) != 0);
```

That is the loop this PR makes most likely to hang. `arch.mk:178` sets `SDHCI_SKIP_HOST_RESET=1` for agilex5, which deliberately inherits whatever state SPL/TF-A left the Cadence host in. If that stage leaves CMD inhibit asserted — exactly the condition skipping the soft reset exposes you to — the first command spins forever, before any of the new timeouts can fire, with no output at all.

Bounding the entry wait with the same idiom and the same `sdhci_reset_cmd_line()` recovery is what makes `SDHCI_SKIP_HOST_RESET` safe to ship.

src/sdhci.c
```diff
@@ -610,6 +631,10 @@ static int sdhci_send_cmd_internal(uint32_t cmd_type,
         }
     }
 
+#if SDHCI_WAIT_AFTER_CMD_US > 0U
```
danielinux: This fires on **every** command, not just the init commands `arch.mk:179` describes ("the controller's required post-command settle time"). `sdhci_send_cmd_internal` is the common path for CMD17/18/24/25 and CMD12 as well.

With `SDHCI_WAIT_AFTER_CMD_US=1000` and `DISK_BLOCK_SIZE=0x80000`, each 512 KB chunk pays ~1 ms for CMD18, ~1 ms for CMD12, and ~1 ms per `sdhci_wait_busy()` CMD13 — several hundred ms to over a second of pure `udelay` on a Linux FIT, for a settle requirement that only applies to command setup.

The sharper problem is the footgun for the next target: combined with `SDHCI_FORCE_SINGLE_BLOCK_READ` (zynq/versal/mpfs250 use it) this becomes 1 ms per 512-byte block, i.e. roughly seven minutes for a 200 MB read. Either scope the delay to `sdhci_init()` / the card-init sequence, or document that the two options are mutually exclusive.

src/sdhci.c
```diff
@@ -555,6 +555,17 @@ static uint32_t sdhci_get_response_type(uint8_t resp_type)
 
 #define DEVICE_BUSY 1
 
+static int sdhci_reset_cmd_line(void)
```
danielinux, line 598: This changes command-error recovery on every SDHCI target, not just agilex5 — `sdhci_reset_cmd_line()` now runs on the EINT path for zynq, versal, mpfs250, zynq7000 and tegra234 too.

It is the right thing per SDHCI spec 3.10.1, and it arguably fixes the pre-existing CMD0 retry loop which retried without ever resetting the line. But it is an unguarded change to five validated paths backed by hardware testing on one board, while the other two behaviour changes in this PR (`SDHCI_SKIP_HOST_RESET`, `SDHCI_WAIT_AFTER_CMD_US`) are both opt-in.

Either gate it the same way, or state in the PR description that it is a deliberate cross-target change so it isn't discovered later.

src/sdhci.c
```diff
@@ -555,6 +555,17 @@ static uint32_t sdhci_get_response_type(uint8_t resp_type)
...
+{
+    uint32_t timeout = 0x000FFFFF;
+
+    sdhci_reg_or(SDHCI_SRS11, SDHCI_SRS11_RESET_CMD);
+    while (((SDHCI_REG(SDHCI_SRS11) & SDHCI_SRS11_RESET_CMD) != 0U) &&
```
danielinux, line 603: There is already a reset helper 130 lines below this one, and it is the unbounded one:

```c
/* src/sdhci.c:691 */
static inline void sdhci_reset_lines(void)
{
    sdhci_reg_or(SDHCI_SRS11, SDHCI_SRS11_RESET_DAT_CMD);
    while (SDHCI_REG(SDHCI_SRS11) & SDHCI_SRS11_RESET_DAT_CMD);
}
```

Same register, same wait, no bound — and it is on the data-transfer path (`sdhci_transfer` calls it at :1365 and again at the end). Since you are already in this code adding a bounded version, giving `sdhci_reset_lines()` the same treatment, or expressing one in terms of the other, closes a hang the rest of the PR is otherwise eliminating.

include/sdhci.h
```diff
@@ -481,6 +483,16 @@ void sdhci_platform_init(void);
  * to the card clock and return that, keeping the internal divider ~1. */
 uint32_t sdhci_platform_set_clock(uint32_t clock_khz, uint32_t base_clk_khz);
 
+/* Preserve a host controller initialized by an earlier boot stage. */
+#ifndef SDHCI_SKIP_HOST_RESET
```
danielinux: Both new knobs break if defined without a value. `#if SDHCI_SKIP_HOST_RESET == 0` here, and `#if SDHCI_WAIT_AFTER_CMD_US > 0U` at :492, are preprocessor errors when the macro expands to nothing — so `-DSDHCI_SKIP_HOST_RESET` fails where `-DSDHCI_SKIP_HOST_RESET=1` works.

The neighbouring SDHCI options are all `#ifdef`-style (`SDHCI_FORCE_CARD_DETECT`, `SDHCI_SDMA_DISABLED`, `SDHCI_BLOCK_VIA_PDMA`), so the valueless form is what someone will reach for. `#if defined(X) && (X != 0)` costs nothing and removes the trap.

hal/agilex5.c
```diff
@@ -0,0 +1,352 @@
...
+
+    for (i = 0; i < len; i++) {
+        uint32_t timeout = UART_TIMEOUT;
+        if (buf[i] == '\n') {
+            while (((UART_REG(UART_LSR) & UART_LSR_THRE) == 0U) &&
+                    (--timeout != 0U)) {}
+            UART_REG(UART_THR) = '\r';
+            timeout = UART_TIMEOUT;
+        }
+        while (((UART_REG(UART_LSR) & UART_LSR_THRE) == 0U) &&
```
danielinux, line 74: The timeout is measured but not acted on: both THRE waits fall through to `UART_REG(UART_THR) = ...` whether or not `timeout` reached zero, so a wedged UART produces corrupted output rather than dropped output.

Not a boot hazard — bounding the loop is the important part and you have that — but skipping the write when the timeout expires is one line and makes the bound mean something.

danielinux (changes requested):

src/sdhci.c
aidangarske: Valdiated on hw. SDHCI_SKIP_HOST_RESET strands the Cadence host in the UHS-I 1.8V/SDR104 state SPL left so CMD8 never answers and it panics. Removed the knob and it now does the normal host reset like the build that was actually validated (the define lived in hal/agilex5.h, which sdhci.c never includes so it was dead until it got moved to arch.mk untested). Entry wait is still bounded. Verified booting Linux + ALTERA-FCS on the 013B.

src/sdhci.c
aidangarske: Removed the knob. It plus SKIP_HOST_RESET was the actual bug; full host reset instead, no per-command settle needed

src/sdhci.c
aidangarske: Gone with the skip-reset path, so no cross-target behavior change.

src/sdhci.c
aidangarske: bounded same way

include/sdhci.h
aidangarske: Both removed

hal/agilex5.c
aidangarske: Fixed, skips the THR write when the timeout expires.

PR #851, opened 2026-08-11
Fenrir fixes 2026-08 -11 + build regressions fixes

danielinux:
cfaf1456 libwolfboot: propagate the head-block write failure
fca6bf0f Address PR review on the encrypted write and sector copy paths
31a9adc6 test: bump footprint limits
e24fff0a libwolfboot: declare ForceZero() in the test-app build
db8e768b aarch64: link the ARM ChaCha port when ChaCha is selected
4726dce4 update_disk: use wc_ForceZero() in the DISK_ENCRYPT helpers
f9fe1386 F-6130: clear disk_encrypt_key/nonce on the FIT DTS load failure path
f7551256 F-6757: fix partial-word hal_flash_write on nrf52, nrf5340 and stm32l0
40021b8c F-7069: clear the EH authValue from the stack in wolfBoot_tpm2_get_timestamp
a799a98c F-7382: clip QSPI page program transfers at the device page boundary
ab7b79cd F-7383: use one consistent sector size in mcxw hal_flash_erase
bc743ad3 F-7987: abort the swap when a sector copy fails
f446a4ad F-8007: wipe TPM advanced-IO staging buffers in TPM2_IoCb()
93edc290 F-7985: enter legacy uImage at ih_ep when it differs from ih_load
5dfdec31 F-7992: bound staged ciphertext in ext_flash_encrypt_write()
f9957da5 F-8003: separate decoded key objects for hybrid signers
748fa8a1 F-8006: return errors from delta base-hash validation in sign tool
e200579d F-7969: validate boot-side digest before delta base hash compare
7608e333 F-7988: clamp unaligned head size in ext_flash_encrypt_write()
e6655e69 F-7989: clamp unaligned head size in ext_flash_decrypt_read()


src/libwolfboot.c
```diff
@@ -2616,27 +2626,59 @@ int RAMFUNCTION ext_flash_encrypt_write(uintptr_t address, const uint8_t *data,
...
+        if (step > len)
+            step = len;
         if (ext_flash_read(row_address, block, ENCRYPT_BLOCK_SIZE)
                 != ENCRYPT_BLOCK_SIZE) {
             return -1;
         }
         XMEMCPY(block + row_offset, data, step);
         crypto_encrypt(enc_block, block, ENCRYPT_BLOCK_SIZE);
-        ext_flash_write(row_address, enc_block, ENCRYPT_BLOCK_SIZE);
+        ret = ext_flash_write(row_address, enc_block, ENCRYPT_BLOCK_SIZE);
```
dgarske, line 2650: **🔴 [High] ext_flash_encrypt_write() discards the head-block write failure, defeating the F-7987 swap-abort it is paired with**

The PR now captures the head block's write result (`ret = ext_flash_write(row_address, enc_block, ENCRYPT_BLOCK_SIZE);`) but only returns it when the whole request fit inside that block (`if (step == len) return ret;`). When the request extends past the head block, `ret` is unconditionally overwritten by `ret = 0;` ten lines later, so a failure programming the first (unaligned) block is silently swallowed and the function can still return 0. This matters much more after this PR than before it: F-7987 rewires `wolfBoot_copy_sector()` to abort the swap on `wb_flash_write(...) < 0`, and `wb_flash_write()` on an external encrypted partition *is* `ext_flash_encrypt_write()`. A failed head-block program therefore still reports success, `wolfBoot_copy_sector()` returns `pos`, and `wolfBoot_update()` advances the persistent sector flag over a partially written sector - exactly the corrupted-backup scenario F-7987 sets out to prevent. The same call is error-checked in the new chunk loop (`if (ret < 0) return ret;`), so this is an internal inconsistency, not a deliberate policy.

**Fix:** Add `if (ret < 0) return ret;` immediately after the head-block `ext_flash_write()`, matching the check already added to the chunk loop below. Extend `test_encrypt_write_keeps_trailing_partial_block` (or add a sibling) that forces the head-block write to fail and asserts a negative return.

src/update_ram.c
```diff
@@ -650,6 +661,14 @@ void RAMFUNCTION wolfBoot_start(void)
     }
 #endif /* MMU */
 
+#ifdef WOLFBOOT_UBOOT_LEGACY
+    /* Enter the uImage at ih_ep. Skipped if a later stage (ELF/FIT) re-derived
```
dgarske, line 678: **🟠 [Medium] The load_address == uboot_load identity guard is defeated by elf_load_image_mmu() publishing *pentry before it validates**

The new override comment states it is "Skipped if a later stage (ELF/FIT) re-derived the load address, since that stage supplies its own entry point." That inference relies on pointer identity, which is not a reliable proxy. `src/elf.c:111` does `*pentry = GET_H64(entry);` *before* the program-header bounds check and the segment loop, so the `-3` (ph table out of bounds), `-4` (segment offset/size out of bounds) and `-5` (segment destination out of range) returns all leave `load_address` overwritten with the ELF's declared entry. `update_ram.c` then prints "Invalid elf, falling back to raw binary" and continues - but `load_address != uboot_load`, so the new ih_ep override is silently skipped and wolfBoot enters at the ELF-declared address rather than at either ih_load or ih_ep. The converse also exists: if a later stage legitimately re-derives an address numerically equal to ih_load (an ELF whose entry == ih_load, or a FIT kernel sub-image loaded at ih_load), the identity test passes and the override wrongly redirects to ih_ep. Verified: both `do_boot()` call sites and the ZynqMP BL31 handoff are after the override, so those are consistent.

**Fix:** Track "a later stage took over the entry point" explicitly with a flag set inside the ELF/FIT success branches, instead of comparing `load_address` to `uboot_load`. Separately, `elf_load_image_mmu()` should not publish `*pentry` until it has passed all of its validation.

src/update_disk.c
```diff
@@ -235,13 +236,13 @@ static int decrypt_header(const uint8_t *src, uint8_t *dst)
 
 static void disk_crypto_clear(void)
```
dgarske, line 237: **🟠 [Medium] disk_crypto_clear() leaves the expanded key schedule and nonce live, so the new scrub does not actually destroy the key**

The PR adds a `disk_decrypted_header_clear(dec_hdr); disk_crypto_clear();` pair before the FIT DTS-load panic (F-6130), whose stated purpose is that key material must not stay resident when wolfBoot_panic() spins forever. But `disk_crypto_clear()` only zeroes this file's own copies (`disk_encrypt_key`, `disk_encrypt_nonce`). The cipher state that actually performs the decryption lives in globals in `src/libwolfboot.c`: `Aes aes_dec, aes_enc;` (line 2155) or `ChaCha chacha;` (line 2105), plus `static uint8_t encrypt_iv_nonce[ENCRYPT_NONCE_SIZE]` (line 92). `aes_init()`/`chacha_init()` expand the key into those structs via `wc_AesSetKeyDirect()`/`wc_Chacha_SetKey()`, and `crypto_decrypt()` uses them directly. None are touched. After the new scrub (and after every other scrub site in this file) the raw key is still trivially recoverable from `aes_dec.key[]` / `chacha.X[]` and the nonce from `encrypt_iv_nonce`. This is a pre-existing limitation of the helper, but the PR is specifically extending its reach, so it is worth closing in the same change.

**Fix:** Add a `crypto_deinit()` helper in `libwolfboot.c` that `wc_ForceZero()`s `aes_dec`/`aes_enc` (plus `wc_AesFree()`) or `chacha`, and `encrypt_iv_nonce`, then call it from `disk_crypto_clear()`. Extend `unit-update-disk-fit.c`'s panic-hook snapshot to also inspect the cipher context, so the test proves the key is really gone.

tools/keytools/sign.c
```diff
@@ -3807,33 +3856,9 @@ int main(int argc, char** argv)
...
-    else if (CMD.sign == SIGN_LMS) {
-        wc_LmsKey_Free(&key.lms);
-    }
-    else if (CMD.sign == SIGN_XMSS) {
-        wc_XmssKey_Free(&key.xmss);
-    }
-    else if (CMD.sign == SIGN_ML_DSA) {
-        wc_MlDsaKey_Free(&key.ml_dsa);
+    free_key(CMD.sign, 0);
+    if (CMD.hybrid) {
```
dgarske, line 3871: **🟠 [Medium] main() still exit(1)s on secondary-key load failure, skipping the very key scrubbing F-8006 restores**

F-8006's rationale is that `exit(1)` deep in the call chain "skips ... main()'s zero_and_free(kbuf, key_buffer_sz) and algorithm-specific key free, so the raw and decoded private signing key are never scrubbed." That reasoning applies verbatim to the hybrid path a few lines above the changed cleanup: when `load_key(..., 1)` fails, `exit(1)` runs while the *primary* raw key buffer `kbuf` is live and the primary decoded key object `key` is fully initialized. Neither `zero_and_free(kbuf, key_buffer_sz)` nor the new `free_key(CMD.sign, 0)` is reached. A missing or unreadable secondary key file is an ordinary user error, so this is easy to hit. Note `unit-sign-hybrid-keyload.c`'s `tcase_add_exit_test(..., 1)` currently pins the exit(1) behaviour, so a fix needs that test updated too.

**Fix:** Replace the `exit(1)` with a jump to the existing tail cleanup (or set `ret` and skip to it), so the primary raw and decoded key are scrubbed on this path too. Update `test_sign_main_fails_when_secondary_key_missing` to match the new exit status.

src/libwolfboot.c
```diff
@@ -2616,27 +2626,59 @@ int RAMFUNCTION ext_flash_encrypt_write(uintptr_t address, const uint8_t *data,
...
 
-    /* encrypt remainder */
+    /* encrypt remainder, staging at most one cache worth at a time */
+    ret = 0;
     step = sz & ~(ENCRYPT_BLOCK_SIZE - 1);
-    for (i = 0; i < step / ENCRYPT_BLOCK_SIZE; i++) {
-        XMEMCPY(block, data + (ENCRYPT_BLOCK_SIZE * i), ENCRYPT_BLOCK_SIZE);
-        crypto_encrypt(ENCRYPT_CACHE + (ENCRYPT_BLOCK_SIZE * i), block,
-            ENCRYPT_BLOCK_SIZE);
+    while (step > 0) {
```
dgarske, line 2664: **🔵 [Low] ext_flash_encrypt_write() chunk size is not rounded down to an ENCRYPT_BLOCK_SIZE multiple**

The new staging loop caps `chunk` at `NVM_CACHE_SIZE` but never rounds it down to a whole number of encryption blocks. The `for` loop then encrypts only `floor(chunk / ENCRYPT_BLOCK_SIZE)` blocks while `ext_flash_write(address, ENCRYPT_CACHE, chunk)` writes all `chunk` bytes - so if `NVM_CACHE_SIZE % ENCRYPT_BLOCK_SIZE != 0` the tail of every chunk is stale/uninitialized cache content, `address` stops being block-aligned, and the keystream desynchronises against `ext_flash_decrypt_read()`. Worse, if `NVM_CACHE_SIZE < ENCRYPT_BLOCK_SIZE` the loop encrypts nothing at all and writes pure garbage. Today `NVM_CACHE_SIZE` defaults to `WOLFBOOT_SECTOR_SIZE`, whose smallest value across `config/examples/` is 512, so every shipped config is a clean multiple of both 16 (AES) and 64 (ChaCha) and this is latent. It is a silent trap for any `-DNVM_CACHE_SIZE` override or a user-supplied `WOLFBOOT_ENCRYPT_CACHE` (whose size is never checked against `NVM_CACHE_SIZE`).

**Fix:** Round the cap down to a block multiple, and add a compile-time assertion (the file already uses the `typedef char ...[cond ? 1 : -1]` idiom at line 2803) that `NVM_CACHE_SIZE >= ENCRYPT_BLOCK_SIZE` and `NVM_CACHE_SIZE % ENCRYPT_BLOCK_SIZE == 0`.

hal/mcxw.c
```diff
@@ -223,15 +227,21 @@ static void erase_flash_sector(uint32_t *dst) {
     /* Wait for completion */
     while (!(FMU0->FSTAT & 0x00000080)) {}
 }
+#endif /* !WOLFBOOT_UNIT_TEST_FLASH_ERASE */
 
 int RAMFUNCTION hal_flash_erase(uint32_t address, int len)
```
dgarske, line 232: **🔵 [Low] mcxw hal_flash_erase() does not extend len by the round-down amount, leaving the last sector unerased**

The fix correctly unifies the stride, but `len` is still the caller's original value after `address` has been rounded down to the sector boundary. A request whose start is unaligned and whose end falls into a later sector therefore under-erases: with `sector_size = 0x1000`, `hal_flash_erase(FLASH_BASE + 0xF00, 0x200)` rounds the address to `FLASH_BASE`, runs one iteration (`len` goes 0x200 -> -0xE00) and erases only sector 0, leaving `[0x1000, 0x1100)` in sector 1 unerased even though it was inside the requested range. `hal/mcxn.c` (cited as the model) has the same shape, and the new `test_erase_unaligned_start_rounds_down` picks `(0x800, 0x1800)` which happens to be covered, so the gap is codified rather than caught. Practical reachability is low - the standard `wb_flash_erase()` callers pass sector-aligned addresses - but the guard is cheap.

**Fix:** Add the alignment adjustment to `len` when rounding the address down, and add a `unit-flash-erase-mcxw` case such as `hal_flash_erase(FLASH_BASE + 0xF00, 0x200)` with `sector_size = 0x1000` asserting two erase commands. Consider the same fix in hal/mcxn.c.

src/update_ram.c
```diff
@@ -506,14 +510,21 @@ void RAMFUNCTION wolfBoot_start(void)
                     os_image.fw_size);
             }
 #endif
+            /* bootm relocates to ih_load but enters at ih_ep: kernels built
```
dgarske, line 515: **🔵 [Low] ih_ep is still discarded when ih_load == 0, and the TODO that documented that gap was removed**

The PR deletes `(void)ih_ep; /* TODO: pass through to do_boot when ih_ep != ih_load */` and replaces it with the ih_ep capture - but the capture lives inside the `if (ih_load != 0)` branch. In the `else` branch (ih_load == 0, the Linux/PPC convention) the code still only advances `load_address` past the 64-byte uImage header and never looks at `ih_ep`. A uImage built with `mkimage -a 0 -e \<ep>` is valid and U-Boot bootm would enter at ih_ep; wolfBoot will still enter at `load_address + 64`. The capture point within the `ih_load != 0` branch is otherwise correct - `load_address = (uint32_t*)(uintptr_t)ih_load` happens above it, so `uboot_load == ih_load` as intended, and the `(uintptr_t)` cast matches the existing ih_load cast so it is fine on 64-bit.

**Fix:** Either handle ih_load == 0 as well (capture `uboot_load` after the +UBOOT_IMG_HDR_SZ advance and set `uboot_entry` when ih_ep differs), or add an explicit comment in the else branch stating that ih_ep is intentionally ignored there, so the information carried by the removed TODO is not lost.

tools/keytools/sign.c
```diff
@@ -3807,33 +3856,9 @@ int main(int argc, char** argv)
...
-    }
-    else if (CMD.sign == SIGN_LMS) {
-        wc_LmsKey_Free(&key.lms);
-    }
-    else if (CMD.sign == SIGN_XMSS) {
-        wc_XmssKey_Free(&key.xmss);
-    }
-    else if (CMD.sign == SIGN_ML_DSA) {
-        wc_MlDsaKey_Free(&key.ml_dsa);
+    free_key(CMD.sign, 0);
```
dgarske, line 3870: **🔵 [Low] free_key() can run on a never-initialized key2 object**

`main()` now calls `free_key(CMD.secondary_sign, 1)` whenever `CMD.hybrid` is set. There are paths where the corresponding member of `key2` was never initialized: with `--manual-sign` or `--sha-only`, `load_key_rsa()` takes the branch at line 599 and never calls `wc_InitRsaKey()`; `load_key()`'s ED25519/ED448 raw-public-key branches (lines 736-740, 805-808) never call `wc_ed25519_init()`/`wc_ed448_init()`; and in the DER-public-key branches `load_key()` already frees "no matter what" (lines 754-755, 822-823), so `free_key()` frees a second time. In practice this is safe - `key`/`key2` are zero-initialized file-scope statics and the wolfCrypt frees are NULL-checked and idempotent - and the primary object had exactly the same exposure before the PR, so the change is symmetric rather than worse. Flagged because the new helper makes it cheap to state the invariant instead of relying on it implicitly.

**Fix:** Either add a `loaded` flag to `struct signing_key`, set where the object is actually initialized and cleared in load_key()'s "free no matter what" paths, or document the zero-initialized/idempotent assumption in the free_key() comment.

.gitignore
```diff
@@ -204,13 +204,15 @@ tools/unit-tests/unit-otp-keystore
 tools/unit-tests/unit-otp-keystore-gen-zeroize
 tools/unit-tests/unit-tpm-api-names
 tools/unit-tests/unit-tpm-nsc-cert
+tools/unit-tests/unit-tpm-advio-zeroize
```
dgarske, line 207: **🔵 [Low] Two new unit-test binaries are missing from .gitignore**

The PR adds five new unit-test targets to `tools/unit-tests/Makefile` (`unit-tpm-advio-zeroize`, `unit-tpm-mfgid-eh-zeroize`, `unit-flash-erase-mcxw`, `unit-flash-write-nrf52`, `unit-update-disk-fit`) but only three of the resulting binaries to `.gitignore`. `tools/unit-tests/unit-tpm-mfgid-eh-zeroize` and `tools/unit-tests/unit-flash-write-nrf52` are missing, so they show up as untracked files after `make run` in tools/unit-tests.

**Fix:** Add the two missing entries next to their siblings in .gitignore.

hal/nrf52.c
```diff
@@ -83,15 +83,19 @@ int RAMFUNCTION hal_flash_write(uint32_t address, const uint8_t *data, int len)
         } else {
```
dgarske, line 87: **🟠 [Medium] nrf52/nrf5340/stm32l0: the sibling 32-bit fast path still indexes off the original address and is left broken by the…**

The PR fixes the byte-wise branch to derive the containing word from `address + i`, but the 32-bit fast branch immediately above it still uses `dst = (uint32_t *)address; ... dst[i >> 2] = src[i >> 2];`. That addresses physical byte `address + 4*(i>>2)`, which equals the intended `address + i` only when `i % 4 == 0` - i.e. only when `address % 4 == 0`. When the destination is misaligned and the source shares that misalignment (e.g. `address % 4 == 1` and `data % 4 == 1`), the fixed byte loop advances `i` to 3, at which point both fast-path conditions are satisfied and the code writes `data[0..3]` to `address..address+3` instead of `data[3..6]` to `address+3..address+6` - plus it issues an unaligned 32-bit flash access, which faults outright on the Cortex-M0+ of stm32l0. This is the same defect class F-6757 fixes in the byte path, in the same function, in all three files the PR touches. The new `unit-flash-write-nrf52.c` deliberately picks a source misaligned by 2 against a destination misaligned by 1 so that "the fast 32-bit path is never taken", so the regression suite does not cover it.

**Fix:** Fix the fast path the same way as the byte path (index the pointers by `i` directly rather than by `i >> 2` off the base) in hal/nrf52.c, hal/nrf5340.c and hal/stm32l0.c, and add a test case to `unit-flash-write-nrf52.c` with `address % 4 == data % 4 != 0` and `len >= 8`, which is the combination that currently reaches it.

---
> **Note:** Referenced line (`hal/nrf52.c:75-82`) is outside the diff hunk. Comment anchored to nearest changed region.

dgarske (commented):
## Skoll Code Review

**Scan type:** `review`**Overall recommendation:** `REQUEST_CHANGES`
**Findings:** 19 total — 12 posted, 7 skipped
**10 finding(s) posted as inline comments** (see file-level comments below)
**2 finding(s) not tied to a diff line** (full detail below)

### Posted findings
- [High] **ext_flash_encrypt_write() discards the head-block write failure, defeating the F-7987 swap-abort it is paired with** — `src/libwolfboot.c:2638-2648`
- [Medium] **The load_address == uboot_load identity guard is defeated by elf_load_image_mmu() publishing *pentry before it validates** — `src/update_ram.c:665-670`
- [Medium] **disk_crypto_clear() leaves the expanded key schedule and nonce live, so the new scrub does not actually destroy the key** — `src/update_disk.c:237-241`
- [Medium] **main() still exit(1)s on secondary-key load failure, skipping the very key scrubbing F-8006 restores** — `tools/keytools/sign.c:3825-3828`
- [Medium] **nrf52/nrf5340/stm32l0: the sibling 32-bit fast path still indexes off the original address and is left broken by the F-6757 fix** — `hal/nrf52.c:75-82`
- [Low] **ext_flash_encrypt_write() chunk size is not rounded down to an ENCRYPT_BLOCK_SIZE multiple** — `src/libwolfboot.c:2650-2665`
- [Low] **mcxw hal_flash_erase() does not extend len by the round-down amount, leaving the last sector unerased** — `hal/mcxw.c:232-247`
- [Low] **ih_ep is still discarded when ih_load == 0, and the TODO that documented that gap was removed** — `src/update_ram.c:513-527`
- [Low] **free_key() can run on a never-initialized key2 object** — `tools/keytools/sign.c:3859-3862`
- [Low] **Two new unit-test binaries are missing from .gitignore** — `.gitignore:207-227`

#### F-8006's double-fclose fix is incomplete: three more fclose(f) sites in make_header_ex still leave f dangling
**File:** `tools/keytools/sign.c:1779,1855,1929`
**Function:** `make_header_ex`
**Severity:** Medium

The `f = NULL;` added at line 1515 is correct and necessary - without it the four new `goto failure` sites in the delta base-hash block would double-fclose at the cleanup label. But the identical pattern remains at three other sites in the same function: line 1779 (SHA256 image-hash loop), line 1855 (SHA384) and line 1929 (SHA3-384) all call `fclose(f)` without clearing `f`. Between those points and the next reassignment of `f` (line 1951 or 1988) several `goto failure` sites are reachable: line 1968 (signature malloc failure), line 1982 (primary sign_digest failure), lines 2008/2015 (secondary sign_digest failure). The cleanup block at line 2347 then does `if (f) fclose(f);` on the stale pointer. Reproduced on the PR build under valgrind: `./sign --rsapss2048 --sha3 image.bin rsa2048.der 7` (sign_digest returns -1 for RSA-PSS + SHA3, then `goto failure`) gives `Invalid read of size 4 ... fclose ... Block was alloc'd at ... fopen ... make_header_ex`. This is pre-existing on master, but it is exactly the defect class F-8006 set out to remove, a few lines away in the same function.

**Recommendation:** Add `f = NULL;` after the `fclose(f)` calls at lines 1779, 1855 and 1929. That makes `if (f) fclose(f);` in the cleanup block correct for every `goto failure` path in the function, matching what the rest of the function already does.

_Referenced code: `tools/keytools/sign.c:1779,1855,1929` (8 lines)_

---

#### PART_SANITY_CHECK() panics three lines above the DISK_ENCRYPT scrub, leaving the last terminal exit uncovered
**File:** `src/update_disk.c:687-692`
**Function:** `wolfBoot_start`
**Severity:** Low

Auditing the terminal exits reached by the F-6130 change: every `wolfBoot_panic()` after the key is populated is now covered (lines 300, 307, 318, 332, 371, 415, 513, 563, 598, 613, 638 (new) and 675), and the FSP "doesn't fit in low memory" break falls into the `if (failures)` scrub. The one remaining hole is `PART_SANITY_CHECK(&os_image)` at line 687, which expands to `if (hdr_ok != 1 || sha_ok != 1 || signature_ok != 1) wolfBoot_panic();` (include/image.h:1692) or to the ARMORED `bne .` spin (include/image.h:278). It sits immediately above the success-path scrub at lines 690-692, so a sanity-check failure hangs forever with `disk_encrypt_key`, `disk_encrypt_nonce` and `dec_hdr` still live in RAM - the exact exposure this PR closes on the DTS path. `wolfBoot_hook_boot(&os_image)` at line 684 has the same property if a project hook does not return. (`dec_hdr` is correctly in scope and populated at the new cleanup site.)

**Recommendation:** Move the `#ifdef DISK_ENCRYPT` scrub block above `wolfBoot_hook_boot()` and `PART_SANITY_CHECK()`. The key material is no longer needed once the payload is decrypted, so scrubbing earlier costs nothing and closes the last terminal exit.

_Referenced code: `src/update_disk.c:687-693` (7 lines)_

---

### Skipped findings
- [Low] ~~New trailing partial-block RMW programs up to ENCRYPT_BLOCK_SIZE-1 bytes past the requested range~~
- [Low] ~~F-7987's abort-on-failure is a no-op for internal-flash HALs, including three the same PR touches~~
- [Low] ~~The delta loop's non-encrypted wb_flash_write() return is still ignored while everything around it now checks~~
- [Low] ~~unit-flash-write-nrf52 build rule omits -Wno-int-to-pointer-cast, unlike its sibling rules~~
- [Info] ~~Delta base-hash validation failure now exits with status 255 instead of 1~~
- [Info] ~~TPM2_IoCb wipes the full staging buffers on every TIS transaction instead of just the transferred bytes~~
- [Info] ~~Blanket +40 byte size-limit bump across every algorithm with no explanation~~

dgarske (approved):

PR #857, opened 2026-08-12
Build coding.o unconditionally with asn.o

holtrop-wolfssl:
This is needed for an upcoming wolfssl change which adds functionality to coding.c that asn.c depends on (wolfboot keytools CI test was failing in wolfssl without this in place)


JacobBarthelmeh (approved):

JacobBarthelmeh:
This change seems reasonable, will leave it to you @danielinux for final review of impact to wolfBoot and merge.

PR #868, opened 2026-08-21
Fenrir fixes 2026 08 21

danielinux:
04531cbf F-9751: Add positive E2E encrypted-update test, fix what it exposes
5caac27e F-9750: decrypt the stored block before the encrypted RMW patch
d64bf671 F-9755: assert wolfBoot_success erases the firmware encryption key
00bc8b58 F-6762: compare the TLV field budget in a 32-bit domain
e9ee8379 F-7396: scrub the decrypted-header cache after field extraction
495b80b9 F-7966: propagate the CTR counter carry without branching on the nonce
f3098cbf F-9766: scrub the nonce copies in the IV-derivation helpers
58390cbf F-7071: scrub the staging buffer on the swap resume early-return
8d39a5f9 F-9721: make ext_flash_encrypt_write doc match its signature
f08c62a0 F-9756: validate PT_LOAD segments before the scatter flash hash walk
d554bbb9 F-7066: bound the FIT DTS relocation copy to the staging size
fe1dcb08 F-9757: length-bound the FIT name/compression string properties
19d8a9e3 F-9758: validate FDT layout in fdt_check_header, use it in fdt_get_string
08903c25 F-7990: bound the compatible string walk to declared lengths
8aa5221f F-7056: fix node-found guards and esdhc log label in hal_dts_fixup
c5ae3680 F-7974: declare hdr_cpy_done as int in both translation units
d722c0f2 F-7971: free the ECC key on the wolfHSM setup error paths
ce5f20c7 F-7970: free the LMS key on the parameter/import error paths
431b0156 F-9759: document that QE/FMan microcode is not authenticated
0a6a5fe3 F-7381: fail SAMA5D3 NAND write/erase instead of reporting success
f5401b0e F-7380: make SAMA5D3 ext_flash_read byte-accurate for partial pages


src/libwolfboot.c
```diff
@@ -2313,12 +2331,18 @@ void aes_set_iv(uint8_t *nonce, uint32_t iv_ctr)
         iv_buf[i] = wb_reverse_word32(iv_buf[i]);
     }
 #endif
-    iv_buf[3] += iv_ctr;
-    if (iv_buf[3] < iv_ctr) { /* overflow */
+    /* Add the block counter with an unconditional, branch-free carry:
+     * a conditional carry loop's trip count would depend on the nonce
+     * content and leak it through timing. */
+    {
```
dgarske: Let's avoid the empty braces please.

src/libwolfboot.c
```diff
@@ -2472,12 +2497,16 @@ void pkcs11_crypto_set_iv(uint8_t *nonce, uint32_t iv_ctr)
             cb_words[i] = wb_reverse_word32(cb_words[i]);
         }
 #endif
-        cb_words[3] += iv_ctr;
-        if (cb_words[3] < iv_ctr) { /* overflow */
+        /* Unconditional, branch-free carry (see aes_set_iv) */
+        {
```
dgarske: Here too. Thanks

src/libwolfboot.c
```diff
@@ -1402,7 +1405,7 @@ uint16_t wolfBoot_find_header(uint8_t *haystack, uint16_t type, uint8_t **ptr)
 
 #ifdef EXT_FLASH
 uint8_t hdr_cpy[IMAGE_HEADER_SIZE] XALIGNED(4);
-uint32_t hdr_cpy_done = 0;
+int hdr_cpy_done = 0;
```
dgarske, line 1408: Curious what drove changing this type considering hdr_cpy_done is only ever set to 0 and 1.

hal/sama5d3.c
```diff
@@ -601,28 +601,29 @@ static int nand_check_bad_block(uint32_t block)
...
-    }
+    if (len <= 0)
+        return 0;
 
-    while (len_to_read > 0) {
-        uint32_t sz = len_to_read;
-        uint32_t pages_to_read;
-        if (sz > remaining)
-            sz = remaining;
+    while (remaining > 0) {
```
dgarske, line 619: The F-7380 rewrite correctly fixes three real bugs (column offset, sub-word tail copy, full-page overrun) but breaks bad-block relocation. I confirmed it against the diff:

dgarske (changes requested):

src/libwolfboot.c
danielinux: that's right, it's not because of the value, it's the fix for F-7974. this is used as extern in src/image.c around line 1049. This is what fenrir flagged as incompatible objects types, which is curiously not flagged by -W

hal/sama5d3.c
danielinux: Added tests to simulate the bad blocks on this target, and found the regression in the skipped blocks index. Re-instated the "cursor" index instead of recalculating based on block position 
[d7ddf62](https://github.com/wolfSSL/wolfBoot/pull/868/commits/d7ddf620febbdf37a2c208ce101b2d119e0c8070)

dgarske (approved):

PR #877, opened 2026-08-27
Add SD card disk boot support for NXP T1040 (eSDHC)

dgarske:
This PR adds SD card disk boot to the NXP QorIQ T1040 (PPC e5500) using the on-chip Freescale eSDHC controller, making it the first big-endian target to use the `src/update_disk.c` boot path: wolfBoot reads the signed image from GPT/MBR A/B partitions on the card, picks the higher version, verifies it and boots it from RAM.

### DIFF
--- a/hal/nxp_esdhc.c
+++ b/hal/nxp_esdhc.c
@@ -1,0 +1,762 @@
+/* nxp_esdhc.c
+ *
+ * Copyright (C) 2026 wolfSSL Inc.
+ *
+ * This file is part of wolfBoot.
+ *
+ * wolfBoot is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * wolfBoot is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1335, USA
+ */
+
+/* Freescale/NXP eSDHC block driver for QorIQ (T1040, T1024, T2080).
+ *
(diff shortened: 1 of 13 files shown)


hal/nxp_esdhc.c
```diff
@@ -0,0 +1,723 @@
...
+    int ret;
+
+    if ((count == 0U) || (buf == NULL)) {
+        return -1;
+    }
+
+    /* Standard-capacity cards are addressed in bytes, high-capacity in
+     * blocks. Getting this backwards reads from a wildly wrong offset. */
+    arg = (g_esdhc_hc != 0) ? (uint32_t)lba
+                            : (uint32_t)(lba * SD_BLOCK_SIZE);
```
danielinux: This cast will likely be flagged for truncating LBA to uint32_t.

In the sdhci driver we guard this by returning if > 0xFFFFFFFFUL, suggesting to add the same check here

hal/nxp_esdhc.c
```diff
@@ -0,0 +1,723 @@
...
+    }
+    if (ret != 0) {
+        ESDHC_DBG("esdhc: read cmd failed at lba %u\r\n", (uint32_t)lba);
+        return ret;
+    }
+
+    for (i = 0; i < count; i++) {
+        ret = esdhc_read_block(buf + ((size_t)i * SD_BLOCK_SIZE));
+        if (ret != 0) {
+            return ret;
```
danielinux, line 622: Double check whether this return leaves the flags in the wrong state with PRSSTAT[CDIHB|DLA] set. Later commands may be affected.

hal/nxp_esdhc.c
```diff
@@ -0,0 +1,723 @@
...
+    return ((uint64_t)hi << 32) | (uint64_t)lo;
+}
+
+static uint64_t esdhc_timer_us(void)
+{
+    uint32_t tb_hz = TIMEBASE_HZ;
+
+    if (tb_hz == 0U) {
+        return 0;
+    }
```
danielinux: Returning 0 here means that the `esdhc_udelay()` below will spin forever instead of producing an error.

I'd suggest propagating an error instead in case tb_hz is 0, and correctly panic in disk_init() rather than spinning forever.

Also: this function implies a register MMIO and division at each round, perhaps some options can be compressed by caching the timebase constant value in a local variable, e.g. not call TIMEBASE_HZ() macro all the time if that remains constant

danielinux (changes requested):

danielinux (approved):

PR #880, opened 2026-09-01
Fenrir fixes 2026 09 01

danielinux:
634d6790 F-12061: unit test: keep the XIP fast-path source in the flash model
5e809031 F-12104: kontron tgl: correct the SPIBAR FREG/FPR register offsets
65cab0a0 F-12104: unit test: model the TGL SPI MMIO as 32-bit registers
1c57f8c0 F-12114: pkcs11: wipe the PIN even when no session was established
b548341e F-12114: pkcs11: wipe the login PIN on all pre-handoff paths
6dabdad0 F-12114: pkcs11: wipe the login PIN in crypto deinit
0da41145 F-12066: pci: restore original COMMAND and clear windows on bridge error
7736d975 F-12060: multiboot2: terminate the boot info tag list with an end tag
c55795b5 F-12065: update_ram: reject short ext flash reads on the RAM load
4c447f4b F-12104: Kontron TGL: apply the SPI BIOS-region lock through the BAR
20c90319 F-12064: P1021: check bad-block markers per erase block
6792b5da F-12063: STM32WB: require a full double word in the fast write path
40d448b7 F-12062: STM32L4: require a full double word in the fast write path
5c22c271 F-12061: RP2350: RMW partial pages in hal_flash_write

### DIFF
--- a/hal/kontron_vx3060_s2.c
+++ b/hal/kontron_vx3060_s2.c
@@ -35,14 +35,18 @@
 #define SPI_PCI_FUN 5
 #define SPI_BAR_OFF 0x10
 /* Tiger Lake SPI controller register offsets, memory-mapped at the
- * BAR0 base. FREG0 holds the BIOS flash region base/limit; FPR0 is
- * the protected range register with the same base/limit layout. */
-#define SPI_FREG0 0x50
+ * BAR0 base. FREG1 holds the BIOS flash region base/limit (region 0
+ * is the flash descriptor); FPR0 is the protected range register
+ * with the same base/limit layout. Offsets per the Intel PCH SPI
+ * register map (see drivers/spi/spi-intel.c in the Linux kernel):
+ * FDATA0-15 at 0x10-0x4C, FRACC at 0x50, FREG0-7 at 0x54-0x74,
+ * FPR0-4 at 0x84-0x9C. */
+#define SPI_FREG1 0x58
 #define SPI_FREG_BASE_MASK (0x7fffU << 0)
 #define SPI_FREG_LIMIT_MASK (0x7fffU << 16)
 #define SPI_FREG_LIMIT_SHIFT (16)
 #define SPI_FREG_ADDR_SHIFT (12)
-#define SPI_FPR0 (0x48)
+#define SPI_FPR0 (0x84)
 #define SPI_FPR_WPE (1U << 31)
 #define SPI_FPR_RPE (1U << 15)
 #define SPI_BIOS_HSFSTS_CTL (0x4)
@@ -67,10 +71,10 @@ int tgl_lock_bios_region()
 
     /* The Flash Protected Range register has the same base/limit
      * layout as the Flash Region register: take the BIOS region
-     * (FREG0) and enable read and write protection on it. The SPI
-     * registers live in the BAR's memory-mapped space, not in PCI
-     * configuration space. */
-    reg = mmio_read32(spi_bar + SPI_FREG0);
+     * (FREG1, flash region 1) and enable read and write protection
+     * on it. The SPI registers live in the BAR's memory-mapped
+     * space, not in PCI configuration space. */
+    reg = mmio_read32(spi_bar + SPI_FREG1);
 #if defined(DEBUG)
     bios_reg_base = (reg & SPI_FREG_BASE_MASK) << SPI_FREG_ADDR_SHIFT;
     bios_reg_lim = ((reg & SPI_FREG_LIMIT_MASK) >> SPI_FREG_LIMIT_SHIFT)
@@ -104,7 +108,7 @@ int hal_flash_protect(haladdr_t address, int len)
     (void)len;
 
     /* The TGL BIOS region covers the bootloader partition, so the
-     * hook's address/len are the same range FREG0 describes. */
+     * hook's address/len are the same range FREG1 describes. */
     return tgl_lock_bios_region();
 }
 
--- a/hal/kontron_vx3060_s2.c
+++ b/hal/kontron_vx3060_s2.c
@@ -20,6 +20,7 @@
  */
 
 #include <wolfboot/wolfboot.h>
+#include <hal.h>
 #include <stdint.h>
 #include <uart_drv.h>
 #include <printf.h>
@@ -33,7 +34,10 @@
 #define SPI_PCI_DEV 31
 #define SPI_PCI_FUN 5
 #define SPI_BAR_OFF 0x10
-#define SPI_FREG1 0x58
+/* Tiger Lake SPI controller register offsets, memory-mapped at the
+ * BAR0 base. FREG0 holds the BIOS flash region base/limit; FPR0 is
+ * the protected range register with the same base/limit layout. */
+#define SPI_FREG0 0x50
 #define SPI_FREG_BASE_MASK (0x7fffU << 0)
 #define SPI_FREG_LIMIT_MASK (0x7fffU << 16)
 #define SPI_FREG_LIMIT_SHIFT (16)
@@ -48,6 +52,7 @@ int tgl_lock_bios_region()
 {
     uint32_t spi_bar, spi_cmd;
     uint32_t reg;
+    int ret = 0;
 
 #if defined(DEBUG)
     uint32_t bios_reg_base, bios_reg_lim;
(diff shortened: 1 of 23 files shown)


src/multiboot.c
```diff
@@ -299,6 +300,20 @@ int mb2_build_boot_info_header(uint8_t *mb2_boot_info,
...
+    if (max_size < sizeof(struct mb2_tag)) {
+        MB2_DEBUG_PRINTF("Not enough size to build mb2 end tag\r\n");
+        return -1;
+    }
+    max_size -= sizeof(struct mb2_tag);
+    end_tag = (struct mb2_tag *)idx;
+    end_tag->type = 0;
+    end_tag->flags = 0;
+    end_tag->size = sizeof(*end_tag);
+    idx += sizeof(*end_tag);
```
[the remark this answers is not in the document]
danielinux: Not taking this one: the index is 8-byte aligned by construction. Every tag the builder writes has a size that is a multiple of 8 (memory map: 16-byte header + 24-byte entries, basic memory info: 16 bytes, end tag: 8 bytes), and the builder advances idx only by the sizes it just wrote, starting right after the 8-byte boot info header. So idx is 8-aligned at the end-tag write, and total_size comes out a multiple of 8 as the spec requires. An explicit align-up here would be dead code; the consumer walker in this file (mb2_find_tag_by_type) already aligns up to 8 defensively.

hal/kontron_vx3060_s2.c
```diff
@@ -33,7 +34,10 @@
 #define SPI_PCI_DEV 31
 #define SPI_PCI_FUN 5
 #define SPI_BAR_OFF 0x10
-#define SPI_FREG1 0x58
+/* Tiger Lake SPI controller register offsets, memory-mapped at the
+ * BAR0 base. FREG0 holds the BIOS flash region base/limit; FPR0 is
+ * the protected range register with the same base/limit layout. */
+#define SPI_FREG0 0x50
```
danielinux: Confirmed - fixed in 72e9a207. The Linux kernel Intel PCH SPI driver (drivers/spi/spi-intel.c) defines FREG(n) = 0x54 + 4n (FREG0 = 0x54, FREG1 = 0x58) and FRACC = 0x50, and its partition code treats region 0 as the mandatory flash descriptor and iterates FREG(1..) for the BIOS partition - so the BIOS range is FREG1 at 0x58, which the original code already read. d259977c regressed that to FREG0 (FRACC); restored to FREG1.

hal/kontron_vx3060_s2.c
```diff
@@ -60,29 +65,47 @@ int tgl_lock_bios_region()
...
     wolfBoot_printf("Bios reg base: 0x%x lim: 0x%x\r\n", bios_reg_base,
                     bios_reg_lim);
 #endif
-    /* Flash Protected Range register has very similar layout of the Flash
-     * Region Register, so we can reuse it and just enable read and write
-     * protection
-     */
     reg |= (SPI_FPR_RPE) | (SPI_FPR_WPE);
-    pci_config_write32(0, SPI_PCI_DEV, SPI_PCI_FUN, SPI_FPR0, reg);
+    mmio_write32(spi_bar + SPI_FPR0, reg);
```
danielinux, line 86: Confirmed - fixed in 72e9a207. The kernel defines FDATA(n) = 0x10 + 4n, so 0x48 is FDATA14 (scratch, reads back what is written - exactly the failure mode described), and the protection-range registers sit at 0x84-0x9C (BXT/CNL PR base). SPI_FPR0 is now 0x84. One caveat: JSL is not in the kernel driver platform table (BYT/LPT = 0x74, BXT/CNL = 0x84), so 0x84 follows the same-generation BXT/CNL layout.

tools/unit-tests/unit-rp2350-flash-write.c
```diff
@@ -0,0 +1,277 @@
...
+
+/* Fast path with the source in XIP flash: the data must be staged
+ * to RAM before programming (XIP is disabled while a page is
+ * programmed). */
+START_TEST(test_aligned_page_multiple_write_xip)
+{
+    int i;
+
+    for (i = 0; i < 512; i++)
+        g_flash[3 * FLASH_PAGE_SIZE + i] = (uint8_t)(0x80 ^ (i & 0x3F));
```
danielinux: Confirmed - fixed in 13b70de2. The source now sits at pages 2-3 (offset 2 * FLASH_PAGE_SIZE) with the destination at pages 0-1: non-overlapping and fully inside the 1024-byte model. 6/6 green.

hal/kontron_vx3060_s2.c
danielinux, line 86: Confirmed - same fix as the two sibling comments, 72e9a207: FREG1 at 0x58 as the BIOS range source, SPI_FPR0 at 0x84, and the mirrored constants in unit-kontron-tgl-spi.c updated (FREG0 decoy at 0x54, FPR0 at 0x84, asserting FPR0 carries the FREG1 range). The test runs the real extracted tgl_lock_bios_region(), so it fails if the HAL offsets drift again.

hal/nxp_p1021.c
```diff
@@ -1765,9 +1769,11 @@ int ext_flash_read(uintptr_t address, uint8_t *data, int len)
             /* check for bad page. if either of the first two pages are bad then
              * skip to next block */
             if (i++ < 2 && flash_buf[bad_marker] != 0xFF) {
-                /* skip block - advance address by block and restart position */
+                /* skip block: the bad block's bytes are not delivered
+                 * and the read continues at the next block. pos and
+                 * data already agree (data = original + pos), so only
+                 * the source address moves. */
                 address = (address + block_size) & ~(block_size - 1);
```
danielinux, line 1787: Confirmed - this is real, and it was introduced by my F-12064 fix (e18da317).

The trace: with the marker on the block's second page, page 0 has already been copied to the output and counted in `pos` when the skip fires. The old code rewound `pos` to the block boundary but not the data pointer (the overflow I fixed); my fix removed the `pos` rewind to keep `data = original + pos`, which stopped the overflow but left the bad block's delivered page in the stream - `ext_flash_read` returned `len` with the bad block's content and everything after it shifted.

Fixed in fa0c329a: the output position is recorded at the start of each erase block, and on a bad block both `pos` and the data pointer are rewound to it before the source address advances past the block. That preserves the no-overflow invariant and drops the delivered bad pages.

The regression test `test_bad_marker_second_page_dropped` previously asserted the buggy behavior (bad block's page 0 in the output) - it now asserts the block is fully skipped, fails against the pre-fix code (memcmp of the first 512 bytes), and passes post-fix. Full unit suite green, p1021 powerpc target build green (CI-identical flow).

tools/unit-tests/unit-kontron-tgl-spi.c
```diff
@@ -0,0 +1,229 @@
...
+static void mmio_write32(uintptr_t address, uint32_t value)
+{
+    uint32_t *slot = (uint32_t *)&g_mmio[address - MMIO_BASE];
+
+    *slot = value;
+}
+
+static uint32_t mmio_read32(uintptr_t address)
+{
+    uint32_t val = *(uint32_t *)&g_mmio[address - MMIO_BASE];
```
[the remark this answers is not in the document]
danielinux: Confirmed - fixed in 65cab0a0. g_mmio is now a uint32_t array (64 words = 256 bytes of SPI BAR MMIO) and every access is an aligned 32-bit read/write indexed by (offset / 4) through the mmio_read32/mmio_write32 helpers - no unaligned pointer casts. g_pci_cfg is modeled the same way.

src/libwolfboot.c
```diff
@@ -2585,11 +2585,27 @@ int pkcs11_crypto_decrypt(uint8_t *out, uint8_t *in, size_t size)
...
+}
+
 void pkcs11_crypto_deinit(void)
 {
     if (encrypt_initialized) {
         pkcs11_function_list->C_CloseSession(pkcs11_session);
         encrypt_initialized = 0;
+        pkcs11_pin_wipe();
     }
 }
```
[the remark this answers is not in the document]
danielinux: Confirmed - fixed in 1c57f8c0. pkcs11_crypto_deinit() now calls pkcs11_pin_wipe() unconditionally; only the C_CloseSession/encrypt_initialized handling stays conditional. The pin buffer is pre-populated from the compile-time credential before any session exists, so the wipe must not depend on encrypt_initialized. unit-pkcs11-pin-zeroize carries a dedicated test (test_pkcs11_deinit_no_session): deinit with no session makes zero token calls, is safe to repeat, and still wipes every byte of the pin.

dgarske (approved):

PR #896, opened 2026-09-15
Update wolfTPM submodule to v4.2.0

aidangarske:
- Update wolfTPM submodule pin
- wolfTPM 4.2.0's larger fwTPM context left <12 KB of secure stack on STM32H5, which overran the session slots (StartAuthSession → 0x905); I fixed by trimming the unused NV index slots (context ~93 KB → ~60 KB) and asserting a 32 KB stack reservation at link time.


danielinux (dismissed):

dgarske (dismissed):

dgarske (approved):

options.mk
```diff
@@ -1269,6 +1269,10 @@ ifeq ($(WOLFCRYPT_TZ_FWTPM),1)
   CFLAGS+=-DWC_RSA_PSS
   CFLAGS+=-DWOLFSSL_PSS_SALT_LEN_DISCOVER
   CFLAGS+=-DFWTPM_MAX_COMMAND_SIZE=4096
+  # NV is disabled here; the default 16x2KB NV index slots are dead weight that
```
dgarske, line 1272: @aidangarske could we improve wolfTPM itself to not use so much stack in the FWTPM_NO_NV case by setting or disabling this region? I will accept this PR as is but let's review stack use in wolfTPM with this option.

aidangarske: Ya ill look at wolftpm stack use for this

PR #902, opened 2026-09-17
x86 FSP: Tiger Lake boot fixes, 64-bit Linux payload, OS measurement, and a UPD decoder

dgarske:
## Summary

This PR introduces generic x86 Firmware Support Package (FSP) improvements developed while bringing up wolfBoot on an Intel Tiger Lake UP3 board. 

* **NotifyPhase reset handling:** Honours reset requests (warm, cold, global) returned from `NotifyPhase` instead of treating them as fatal errors. Adds support for Tiger Lake's global reset (`CF9_GLB_RST` in PMC ETR3) with a fail-secure halt if ETR3 is locked.
* **64-bit Linux boot protocol:** Enables the 64-bit bzImage loader under `WOLFBOOT_64BIT` to allow 64-bit x86 FSP targets to boot signed bzImages directly.
* **Optional initrd container:** Adds support for packing an optional initrd into the signed image using a lightweight container header, enabling modular distribution kernels to mount root.
* **Disk OS image measurement:** Adds `WOLFBOOT_MEASURED_PCR_OS` (opt-in) to measure verified OS images into a configurable TPM PCR, including digest algorithm re-hashing and fail-secure error handling.
* **Debug UART selection:** Adds `X86_UART_NUMBER` (default 0) to choose which PCH LPSS SerialIo UART controller the FSP uses for debug output, set consistently across FSP-M and FSP-S so the console does not change between phases. Pairs with the existing `X86_UART_BASE`, which places the controller's MMIO base.
* **FSP fixes & cleanup:** 
  * Initialises `hobList` before `FspMemInit` to prevent uninitialised pointer access on error or reset paths.
  * Fixes a link failure (`wolfBoot_print_hexstr`) when `MEASURED_BOOT=1` is enabled without TPM seal/keystore options.
* **Tooling & Docs:** 
  * Adds `tools/x86_fsp/decode_fsp_upd.py` to parse FSP UPD blocks using header offset comments.
  * Updates `docs/Targets.md` covering the Linux payload container, PCR measurement, and decoder usage.

## Test Status

* **Hardware validation (Tiger Lake UP3 SBC):** Confirmed global reset handling, signed 64-bit bzImage boot (with/without initrd) to userspace, and verified TPM PCR measurements match predicted values.
* **Build verification:** Verified against `kontron_vx3060_s2` target using `MEASURED_BOOT` + `LINUX_PAYLOAD` + SHA-384 options.
* **Regression testing:** Passed QEMU x86 FSP filesystem test (`tools/scripts/x86_fsp/qemu/test_qemu.sh -f`).


src/x86/fsp.c
```diff
@@ -99,6 +100,39 @@ void print_fsp_image_revision(struct fsp_info_header *h)
     wolfBoot_printf("%x.%x.%x build %x\r\n", maj, min, rev, build);
 }
 
+/* Act on a reset request returned by an FSP NotifyPhase call. WARM/COLD do the
+ * matching reset; RESET_REQUIRED_3 is the Tiger Lake global reset (host + CSME)
+ * needed after the FSP-S ChipsetInit sync. A non-success, non-reset status is
+ * fatal. Returns only on EFI_SUCCESS. */
+static void notify_phase_handle_reset(uint32_t status)
```
rizlik: consider generalize it as `fsp_handle_reset` and use it for `mem_init` and `silicon_init` as well.

src/x86/fsp.c
```diff
@@ -99,6 +100,39 @@ void print_fsp_image_revision(struct fsp_info_header *h)
...
+        wolfBoot_printf("ETR3 is locked, cannot arm a global reset\n");
+        panic();
+#else
+        reset(0);
+#endif
+    }
+    if (status != EFI_SUCCESS) {
+        wolfBoot_printf("notify phase failed %d\n", status);
+        panic();
+    }
```
rizlik: NIT: 
change naming so that it's clear that handle errors, or move the error handling outside of the reset handler, eg:
```
    param.Phase = EnumInitPhaseReadyToBoot;
    status = x86_run_fsp_32bit(notify_phase, &param);
    notify_phase_handle_reset(status);
    if (status != EFI_SUCCESS) {
         ...    
    }
```

src/x86/linux_loader.c
```diff
@@ -185,6 +201,30 @@ void load_linux(uint8_t *linux_image, void *params, const char *cmd_line)
     memcpy((uint8_t *)KERNEL_LOAD_ADDRESS, linux_image + param_size,
```
rizlik: Probably better in a follow-up PR but I think we can add should check that the kernel and the initrd do not overlap with wolfBoot code and with the payload.

src/update_disk.c
```diff
@@ -814,6 +835,40 @@ void RAMFUNCTION wolfBoot_start(void)
...
+        wc_Sha256Final(&os_sha, os_pcr) != 0) {
+        wolfBoot_printf("Failed to re-hash the OS digest\r\n");
+        wolfBoot_panic();
+    }
+    measure_ret = wolfBoot_tpm2_extend(WOLFBOOT_MEASURED_PCR_OS, os_pcr,
+                                       __LINE__);
+#else
+    measure_ret = wolfBoot_tpm2_extend(WOLFBOOT_MEASURED_PCR_OS,
+                                       os_image.sha_hash, __LINE__);
+#endif
```
rizlik: did you considered to always re-hash? I think this simplify the design

src/update_disk.c
```diff
@@ -814,6 +835,40 @@ void RAMFUNCTION wolfBoot_start(void)
...
+                        WOLFBOOT_SHA_DIGEST_SIZE) != 0 ||
+        wc_Sha256Final(&os_sha, os_pcr) != 0) {
+        wolfBoot_printf("Failed to re-hash the OS digest\r\n");
+        wolfBoot_panic();
+    }
+    measure_ret = wolfBoot_tpm2_extend(WOLFBOOT_MEASURED_PCR_OS, os_pcr,
+                                       __LINE__);
+#else
+    measure_ret = wolfBoot_tpm2_extend(WOLFBOOT_MEASURED_PCR_OS,
+                                       os_image.sha_hash, __LINE__);
```
rizlik: `sha_hash` includes the wolfboot manifest header. wouldn't be better to measure the actual payload instead?

src/x86/tgl_fsp.c
rizlik: shouldn't this be `upd->SerialIoUartMode[X86_UART_NUMBER]`, (also line 970 probably needs to be adjusted).

src/x86/tgl_fsp.c
```diff
@@ -956,7 +968,7 @@ int fsp_machine_update_s_parameters(uint8_t *default_s_params)
 
     upd->EnableMultiPhaseSiliconInit = 0;
     upd->SerialIoUartMode[1] = upd->SerialIoUartMode[2] = 0x1;
-    upd->SerialIoDebugUartNumber = 0x0;
+    upd->SerialIoDebugUartNumber = X86_UART_NUMBER;
```
rizlik: In FSP-T parameters, the `PcdSerialIoUartNumber` should also be set to `X86_UART_NUMBER`

